PR #14, opened 2022-04-05
Enhancement/docker

### LINKED ISSUE #9: Basic docker-compose boilerplate

gbudau:
Add Docker

close #9


Damrod:
Esto ya no es boilerplate, no?

transcendence-app/Dockerfile
```diff
@@ -0,0 +1,24 @@
+ARG NODE_VERSION=16.14.2
+ARG ALPINE_VERSION=3.15
+
+FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS development
+ENV NODE_ENV=development
```
vdedios, line 5: Just to get how this works. I understand this kind of two targets for whether you deploy the app as dev or prod?

webapp/Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG NODE_VERSION=16.14.2
+ARG ALPINE_VERSION=3.15
+ARG NGINX_VERSION=1.21.6
+
+FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS development
+ENV NODE_ENV=development
+RUN apk add --no-cache dumb-init
```
vdedios, line 7: Just to make image lighter I'd have fewer layers, try to group directives. It is less legible but when deploying it helps to save space

transcendence-app/Dockerfile
vdedios: I've seen this is the multi stage thing you mentioned, I'll read about it, thanks!

nginx/default.conf
```diff
@@ -0,0 +1,23 @@
+server {
+  listen 80;
```
vdedios, line 2: What is the idea of having a proxy?

vdedios (changes requested):
Very nice work!! Please rebase main because there are many unrelated commits. Thanks 😄

transcendence-app/Dockerfile
gbudau: It's a pretty cool feature. You can have one Dockerfile and build various images(development, production, staging, CI, etc...). Also, it's possible to copy files from other stages or create images targeting a specific stage with the `--target` option.

fizcamposancos:
I have a problem with docker-compose up. When I try to run it I got the following message:
"ERROR: Version mismatch: file ./docker-compose.yml specifies version 3.0 but extension file ./docker-compose.override.yml uses version 3.4"
I can change versions locally but I don't know if you get this message too.

nginx/default.conf
gbudau: Pros:
We can use the same port for multiple services, and the proxy does the internal routing. We only have to expose one port to the host.

Cons:
It makes the development more complex.

Maybe it's a bit overkill to use one in this project, but I think we can learn more if we do.

These are pros and cons only for development because there are [many](https://medium.com/intrinsic-blog/why-should-i-use-a-reverse-proxy-if-node-js-is-production-ready-5a079408b2ca) reasons why it's better to use a reverse proxy for production.

gbudau:
> Very nice work!! Please rebase main because there are many unrelated commits. Thanks :smile:

Thank you! I've done a rebase. Please let me know what you think.

gbudau:
> I have a problem with docker-compose up. When I try to run it I got the following message: "ERROR: Version mismatch: file ./docker-compose.yml specifies version 3.0 but extension file ./docker-compose.override.yml uses version 3.4" I can change versions locally but I don't know if you get this message too.

I didn't get this error, but it makes sense because the docker-compose.yml had version 3 and the override and prod version 3.4 because they use the [target](https://docs.docker.com/compose/compose-file/compose-file-v3/#target) directive.

Thank you, I've updated the version to 3.4.

docker-compose.override.yml
```diff
@@ -0,0 +1,31 @@
...
+  nginx:
+    ports:
+      - "3000:80"
+
+  webapp:
+    build:
+      context: ./webapp
+      target: development
+    volumes:
+      - ./webapp/src:/webapp/src:ro
```
Damrod, line 12: why mount it as a volume instead of copying the folder inside the container?
AFAIK, mounting it as a volume is kind of a shortcut when you need persistency, which links to statefulness. Do we need statefulness here?

nginx/default.conf
Damrod: I have used reverse proxies before, +1 from me.

webapp/Dockerfile
Damrod: I think very often (at least last time I was using docker) the layers will break, and a full removal of the image will be required to rebuild anyway. So hard to leverage on caching.
Theoretically is better, but I'm not sure this works practically.

transcendence-app/src/main.ts
```diff
@@ -1,8 +1,10 @@
 import { NestFactory } from '@nestjs/core';
+import { NestExpressApplication } from '@nestjs/platform-express';
```
Damrod, line 2: What does this do? Why do we need to use NestFactory.create<> here instead of the way it was?

docker-compose.override.yml
gbudau: Thank you for asking this.

Docker has many types of volumes (volume, bind mounts, tmpfs mounts, etc.).

This one here is a [bind mount](https://docs.docker.com/storage/bind-mounts/), and we use it to mount the src directory from the host into the container.

Mounting the directory allows us to modify the code from the host while npm reloads the code on change inside the container without having to rebuild the image.

transcendence-app/src/main.ts
gbudau: [NestJS](https://docs.nestjs.com/first-steps#platform) is an abstraction over other Node HTTP frameworks. By default, it uses Express.

On line 7, we use 
```
app.enable('trust proxy');
```

Which is an [Express](https://expressjs.com/en/guide/behind-proxies.html) feature, and to use it, we have to import this module and specify the NestExpressApplication interface.

webapp/Dockerfile
gbudau: I will do some tests to see exactly how much space we save by grouping the directives versus how many seconds we do by caching the dependencies each time we rebuild the images.

In my opinion, time > disk space, but I agree that it's hard to justify without more information.

docker-compose.override.yml
Damrod: I'm understood that earlier, but I'm challenging the concept:

Why do we need to reload the code on change?
It is this only meant as a development shortcut?

Generally you have a program running without necessarily a state, and then the database behind it to preserve the state. Once your program "ships into production", it's meant as an off the shelf product: you don't change the code anymore. You only change the data.

gbudau: Yes, this is a development shortcut :+1:. docker-compose.override.yml has the configuration for the development environment.

fizcamposancos (approved):
After reading all the comments, @gbudau answers, and tested the code, tried some things. I think I have everything viewed and more or less clear, so far...
Very good job, and that's only the begining

Damrod (approved):
Good! Thank you!

webapp/src/App.tsx
```diff
@@ -10,6 +10,7 @@ function App() {
         <p>
           Edit <code>src/App.tsx</code> and save to reload.
         </p>
+        <p>{`Hello, ${process.env.REACT_APP_NAME}!`}</p>
```
Damrod, line 13: So, as far as I understand, this PR is still not connecting frontend to backend, is it? REACT_APP_NAME is not anything we pull for the backend API, just an environment variable that we set in the webapp container. There are not calls to the backend API.
Or did I miss something?

gbudau: Correct, the webapp frontend is at the `/` location, and the transcendence-app backend is at the `/api` location, and it has one route `/api/v1/hello`.

`REACT_APP_NAME` and `TRANSCENDENCE_APP_NAME` are placeholders used to test that Docker passes the environment variables to the containers.

webapp/Dockerfile
gbudau: I've done some tests with these changes:
<details>
<summary>Click to see changes</summary>

```diff
diff --git a/transcendence-app/Dockerfile b/transcendence-app/Dockerfile
index bd686cc..13b3963 100644
--- a/transcendence-app/Dockerfile
+++ b/transcendence-app/Dockerfile
@@ -3,22 +3,18 @@ ARG ALPINE_VERSION=3.15

 FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS development
 ENV NODE_ENV=development
-RUN apk add --no-cache dumb-init
 WORKDIR  /transcendence-app
-COPY ["package.json", "package-lock.json", "./"]
-RUN npm clean-install
 COPY . .
-RUN npm run build
+RUN apk add --no-cache dumb-init && npm clean-install && npm run build
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["npm", "run", "start:dev"]

 FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS production
 ENV NODE_ENV=production
-RUN apk add --no-cache dumb-init
 WORKDIR  /transcendence-app
 COPY ["package.json", "package-lock.json", "./"]
-RUN npm clean-install --production
 COPY --from=development --chown=node:node /transcendence-app/dist ./dist
+RUN apk add --no-cache dumb-init && npm clean-install --production
 USER node
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["node", "dist/main"]

```

```diff
diff --git a/webapp/Dockerfile b/webapp/Dockerfile
index e8f4b31..3f20d83 100644
--- a/webapp/Dockerfile
+++ b/webapp/Dockerfile
@@ -4,11 +4,9 @@ ARG NGINX_VERSION=1.21.6

 FROM node:${NODE_VERSION}-alpine${ALPINE_VERSION} AS development
 ENV NODE_ENV=development
-RUN apk add --no-cache dumb-init
 WORKDIR /webapp
-COPY ["package.json", "package-lock.json", "./"]
-RUN npm clean-install
 COPY . .
+RUN apk add --no-cache dumb-init && npm clean-install
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["npm", "start"]

@@ -17,10 +15,8 @@ ENV NODE_ENV=production
 ARG REACT_APP_NAME
 ENV REACT_APP_NAME=${REACT_APP_NAME}
 WORKDIR /webapp
-COPY ["package.json", "package-lock.json", "./"]
-RUN npm clean-install --production
 COPY . .
-RUN npm run build
+RUN npm clean-install --production && npm run build

 FROM nginx:${NGINX_VERSION}-alpine AS production
 COPY --from=builder --chown=nginx:nginx /webapp/build /usr/share/nginx/html

```

</details>

Optimizing disk space by grouping directives saves 2MB of disk space.
<details>
<summary>Click to see results</summary>

Image size development - with disk size optimization
```
REPOSITORY                           TAG                  IMAGE ID       CREATED              SIZE
ft_transcendence_webapp              latest               3526b74e2744   32 seconds ago       369MB
ft_transcendence_transcendence-app   latest               9146b88eb919   About a minute ago   409MB

```

Image size development - with caching optimization
```
REPOSITORY                           TAG                  IMAGE ID       CREATED              SIZE
ft_transcendence_webapp              latest               39583615cf11   53 seconds ago       370MB
ft_transcendence_transcendence-app   latest               4988fa1fcb8e   About a minute ago   410MB
```

Image size production - with disk size optimization
```
REPOSITORY                           TAG                  IMAGE ID       CREATED              SIZE
ft_transcendence_webapp              latest               b91eab7312c4   14 seconds ago       24MB
ft_transcendence_transcendence-app   latest               66c7eba73357   About a minute ago   126MB
```

Image size production - with caching optimization
```
REPOSITORY                           TAG                  IMAGE ID       CREATED              SIZE
ft_transcendence_webapp              latest               408d3bf353eb   4 days ago           24MB
ft_transcendence_transcendence-app   latest               d4af61464114   2 days ago           126MB
```

</details>

On my machine, for development, `npm clean-install` inside Docker takes around 25 seconds for transcendence-app and 35 seconds for webapp, without caching. With caching, it takes 0 seconds.
Optimizing build time by using layer caching saves 60 seconds of build time.
<details>

```
% docker-compose  up -d --build
...
Building transcendence-app
...
 => [development 5/7] RUN npm clean-install        26.4s 
...

...
Building webapp
...
 => [development 5/6] RUN npm clean-install        36.6s
...

```

</details>

vdedios: For me is reasonable to keep layers then since we are not effectively saving anything

vdedios (approved):
LGTM!

gbudau:
Thank you! :heart:

Issue #10, opened 2022-04-01
Basic GH-Actions


vdedios:
Configure action to test, compile, lint etc PRs

vdedios:
:warning:  Blocked by #8

PR #15, opened 2022-04-18
#10 add basic GH actions for checking frontend and backend

vdedios:
- [x] add `automerge` label and configure generic branch protection for all branches -> (this is the void `integration.yml` file you'll find)

### DIFF
--- a/.github/workflows/transcendence-app_check.yml
+++ b/.github/workflows/transcendence-app_check.yml
@@ -12,18 +12,20 @@ jobs:
     strategy:
       matrix:
         os: [macos-10.15, ubuntu-18.04]
-        steps:
-          - name: Check out Github repository
-            uses: actions/checkout@v2
-          - name: Set up Node
-            uses: actions/setup-node@v3
-            with:
-              node-version: '16'
-          - name: Compile
-            run: npm run build --prefix transcendence-app
-          - name: Check style
-            run: npm run style --prefix transcendence-app
-          - name: Lint
-            run: npm run lint --prefix transcendence-app
-          - name: Test
-            run: npm run test --prefix transcendence-app
+    steps:
+      - name: Check out Github repository
+        uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+      - name: Install dependencies
+        run: npm install --prefix transcendence-app
+      - name: Compile
+        run: npm run build --prefix transcendence-app
+      - name: Check style
+        run: npm run style --prefix transcendence-app
+      - name: Lint
+        run: npm run lint --prefix transcendence-app
+      - name: Test
+        run: npm run test --prefix transcendence-app
--- a/.github/workflows/transcendence-app_check.yml
+++ b/.github/workflows/transcendence-app_check.yml
@@ -1,0 +1,29 @@
+name: transcendence-app check
+
+on:
+  push:
+    branches: [$default-branch]
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-10.15, ubuntu-18.04]
+        steps:
+          - name: Check out Github repository
+            uses: actions/checkout@v2
+          - name: Set up Node
+            uses: actions/setup-node@v3
+            with:
+              node-version: '16'
+          - name: Compile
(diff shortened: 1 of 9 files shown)


vdedios:
As you'll see, auto-generated `package-lock.json` files have huge diffs. This is due to being scaffolded generated with >=npm 7 (not sure which version) and me having npm 6.* -> https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion

Please check which npm version you have (`npm --version`) so all of us are aligned

transcendence-app/src/app.service.ts
```diff
@@ -3,6 +3,6 @@ import { Injectable } from '@nestjs/common';
 @Injectable()
 export class AppService {
   getHello(): string {
-    return `Hello ${process.env.TRANSCENDENCE_APP_NAME}!`;
+    return `Hello World!`;
```
vdedios, line 6: Reverted this change because it will be removed soon and was the easiest way to fix broken test without having to pass any ENVs (not needed for now)

transcendence-app/package.json
```diff
@@ -49,7 +50,8 @@
     "ts-loader": "^9.2.3",
     "ts-node": "^10.0.0",
     "tsconfig-paths": "^3.10.1",
-    "typescript": "^4.3.5"
+    "typescript": "^4.3.5",
+    "webpack": "^5.72.0"
```
vdedios, line 54: Not sure why but had to add webpack as a dev-dependency to be able to compile

.github/workflows/transcendence-app_check.yml
```diff
@@ -0,0 +1,29 @@
+name: transcendence-app check
+
+on:
+  push:
+    branches: [$default-branch]
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
```
gbudau, line 7: Should we also add this part to webapp_check.yml?

.github/workflows/webapp_check.yml
```diff
@@ -0,0 +1,21 @@
...
+          - name: Check out Github repository
+            uses: actions/checkout@v2
+          - name: Set up Node
+            uses: actions/setup-node@v3
+            with:
+              node-version: '16'
+          - name: Compile
+            run: npm run build --prefix webapp
+          - name: Check style
+            run: npm run style --prefix transcendence-app
```
gbudau, line 17: I think this should be `npm run style --prefix webapp`

gbudau (commented):
Thanks for this feature!

Output of `node --version && npm --version`:
```
v16.14.2
8.5.0
```

fizcamposancos:
> node --version
I had an older version but now:
Output of node --version && npm --version:

v16.14.2
8.5.0

fizcamposancos:
Sorry @gbudau, I clicked somewhere and re-requested your review. I don't know how to undo that.

.github/workflows/transcendence-app_check.yml
vdedios: yup, I copy-pasted only part of it :sweat_smile:

.github/workflows/webapp_check.yml
vdedios: same as above, thanks!

.github/workflows/transcendence-app_check.yml
```diff
@@ -0,0 +1,29 @@
...
+            with:
+              node-version: '16'
+          - name: Compile
+            run: npm run build --prefix transcendence-app
+          - name: Check style
+            run: npm run style --prefix transcendence-app
+          - name: Lint
+            run: npm run lint --prefix transcendence-app
+          - name: Test
+            run: npm run test --prefix transcendence-app
```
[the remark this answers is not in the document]
vdedios: completely right, thanks!

vdedios:
It was my problem then, I had installed an old version of npm. I have updated it, pushed new `package-lock.json` generated with it and now we have a much smaller diff. Thanks both of you!

vdedios:
I have finally dismissed adding a new runner for auto-merging and decided to enable [GitHub's feature for this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request), it works really well! I have also enabled only `merge commit`  strategy so we do not get confused

gbudau (approved):
Thank you!

LGTM :rocket:

fizcamposancos (approved):
I think I will need further explanation about this but I can accept it to unblock merging.

PR #16, opened 2022-04-21
#3 Add contributing documentation 

### LINKED ISSUE #3: Add contributing documentation
Basic documentation about how we expect PRs, code reviews and merging into master to be done

vdedios:
closes #3

⚠️ Disclaimer! what you'll see is purely based on my opinion and experience.

The idea is to discuss every point through a review so we all agree how we want to work as a team

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,52 @@
 # ft_transcendence
-🏓 Multiplayer pong tournaments
+🏓 Play pong online with your friends in real time
+
+# 🤝 Contributing
+
+If you want to contribute to this project you may follow a set of rules.
+
+### 1. Create a new issue or task:
+
+- With every contribution there must be a related issue/task, so if you
+don't have any, create a new one on the project board with some
+detailed info about what do you want to accomplish.
+
+### 2. Create a new branch:
+   
+- ⚠️ We **do not** push to `master`, never. So you'll need to create a branch
+to work on it.
+- Try to start always from `master` and do not forget to bring
+all new changes, preferably with `git pull --rebase` so you don't corrupt
+commit history.
+
+### 3. Naming:
+
+- What about branch naming? A good standard way to name could be your
+GH userName + brief description of the task. For example:
+`marvin/add-contributing-docs`
+- We also try to have a **normalized way to name commits**. Because is complicated
+to always have great commit messages (despite you should) a good enough 
+naming strategy is adding your issue number followd by the message. This way other devs can
+keep track of some context of the changes just visiting the related issue
+because GitHub will link to it. Try something like:
+`#11 / add main contributing points`
+
+### 4. Pull Requests:
+    
+- When you are ready, you can create a new PR pointing to master. We have defined
+two requisites to be able to merge your branch into master:
+    - At least **two reviewers approved your changes**
+    - You also have **passed all CI checks** (tests, linting etc.)
+- As with the commits, the ideal is to name it with the issue number followed by a general
+description of what is about: `#11 / add contributing documentation`
+- Once you pass all these checks, **you are responsible to integrate your changes**.
+We have actually two ways: manually (always with a `merge commit` strategy) or setting
+[auto-merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request)
+
+### 5. Conflicts:
+
+- How do I solve conflicts? Most likely, someone merged some commits into `master`
+that affect the same files and lines as yours. One of the **cleanest** ways to solve
+this is just to update (pull) master in your local and, then, rebase master into
+your branch: `git rebase master`. Git will make you solve conflicts one by one,
+you just have to follow the instructions!
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ If you want to contribute to this project you may follow a set of rules.
 - With every contribution there must be a related issue/task, so if you
 don't have any, create a new one on the project board with some
 detailed info about what do you want to accomplish.
+- Add an appropiate label which describes issue type: `feature`, `bug`,
+`task` etc.
 
 ### 2. Create a new branch:
    
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ detailed info about what do you want to accomplish.
 
 ### 2. Create a new branch:
    
-- ⚠️ We **do not** push to `master`, never. So you'll need to create a branch
+- ⚠️ We **do not** push to `main`, never. So you'll need to create a branch
 to work on it.
-- Try to start always from `master` and do not forget to bring
+- Try to start always from `main` and do not forget to bring
 all new changes, preferably with `git pull --rebase` so you don't corrupt
 commit history.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,2 +1,52 @@
...
+- What about branch naming? A good standard way to name could be your
+GH userName + brief description of the task. For example:
+`marvin/add-contributing-docs`
+- We also try to have a **normalized way to name commits**. Because is complicated
+to always have great commit messages (despite you should) a good enough 
+naming strategy is adding your issue number followd by the message. This way other devs can
+keep track of some context of the changes just visiting the related issue
+because GitHub will link to it. Try something like:
+`#11 / add main contributing points`
+
```
fizcamposancos, line 33: When talking about comits, sometimes I just push a commit to the working branch to save my progress, saying WIP or something like that. Should I say '#11 / WIP: comments to this PR' or something like that?

README.md
```diff
@@ -1,2 +1,52 @@
...
+# 🤝 Contributing
+
+If you want to contribute to this project you may follow a set of rules.
+
+### 1. Create a new issue or task:
+
+- With every contribution there must be a related issue/task, so if you
+don't have any, create a new one on the project board with some
+detailed info about what do you want to accomplish.
+
```
fizcamposancos, line 13: Any naming convention for this? #number, FIX, new feature, ...?

Damrod:
By the way, did I read somewhere, or did I dream, that you intend for merged branches to automatically close issues? If it's like this, I have to say I disagree. An issue will very often be a piece of work bigger than what can be sensibly merged at once from one branch. It makes sense to have issues have one or many branches linked to them.

README.md
vdedios: In that case, what I usually do is amend the commit and then force push. I believe having a commit named "WIP" doesn't add much value... Something like:
1. Commit with the name you'd use if task was finished.
2. Next time, `git commit --amend --cleanup=whitespace` (`--cleanup=...` option is due to issues starting with `#` are treated as comments by git so we want to avoid this) Then an editor will open, remove everything but the commit message before saving)
3. And then, `git push --force`

⚠️ Again, this is my usual way of working, if someone has another opinion please manifest yourself!!

README.md
vdedios: Nice idea! I'd say just adding a brief description about what is about and, then, tag appropiately the issue with the type (feature, task, bug etc.)

Do all you agree? If so I'll update the doc

README.md
```diff
@@ -1,2 +1,52 @@
...
+to work on it.
+- Try to start always from `master` and do not forget to bring
+all new changes, preferably with `git pull --rebase` so you don't corrupt
+commit history.
+
+### 3. Naming:
+
+- What about branch naming? A good standard way to name could be your
+GH userName + brief description of the task. For example:
+`marvin/add-contributing-docs`
```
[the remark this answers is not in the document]
gbudau: +1 for adding the issue number

README.md
```diff
@@ -1,2 +1,52 @@
...
+
+### 1. Create a new issue or task:
+
+- With every contribution there must be a related issue/task, so if you
+don't have any, create a new one on the project board with some
+detailed info about what do you want to accomplish.
+
+### 2. Create a new branch:
+   
+- ⚠️ We **do not** push to `master`, never. So you'll need to create a branch
```
gbudau, line 16: suggestion: What do we think about replacing `master` with `main`? To reflect the branch naming we have.

README.md
```diff
@@ -1,2 +1,52 @@
...
+to work on it.
+- Try to start always from `master` and do not forget to bring
+all new changes, preferably with `git pull --rebase` so you don't corrupt
```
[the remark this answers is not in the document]
gbudau: question: This should be `git rebase main`, from the branch we create?  (second point, instead of `git pull --rebase`)

README.md
gbudau: praise: I like this, I think it makes sense to add the issue number to the commit message, and it's a great way to relate commits to issues.

README.md
```diff
@@ -1,2 +1,52 @@
...
+    
+- When you are ready, you can create a new PR pointing to master. We have defined
+two requisites to be able to merge your branch into master:
+    - At least **two reviewers approved your changes**
+    - You also have **passed all CI checks** (tests, linting etc.)
+- As with the commits, the ideal is to name it with the issue number followed by a general
+description of what is about: `#11 / add contributing documentation`
+- Once you pass all these checks, **you are responsible to integrate your changes**.
+We have actually two ways: manually (always with a `merge commit` strategy) or setting
+[auto-merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request)
```
gbudau, line 44: praise: Nice feature, and I like that we can opt-in.

gbudau (commented):
Thank you for writing this!

README.md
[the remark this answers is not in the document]
vdedios: If you `pull` with `--rebase` you'll use a `rebase` strategy rather than `merge` to add remote changes into local. This is useful, for example, if you started working in master forgetting to create a new branch. Your commits will placed at the top of the tip like if you already started working from what is in the remote (which has latest changes)

The overall idea is everything that gets integrated into master should be the starting point of our task

README.md
vdedios: As talked yesterday with @Damrod, maybe we don't feel very comfy with force pushing and we want to have some freedom in our branch commits. He proposed a nice idea which is:
1. total freedom in how we make our commits but, when integrating, we should always use a `squash` strategy with the format -> `issue number + meaningful message for the task`
2. The PR should be associated with a reasonable size task so we don't get a huge commit

We all agree with this? If so, I'll update

gbudau (approved):
Thank you!

LGTM :+1:

Damrod:
> * The ideal would be a single issue per PR (and a single PR for branch as you say) Multiple PRs for one issue would not be desirable. In case we face big tasks I'd say we should split it into smaller chunks/subtasks (and yes, it may be difficult sometimes, but it facilitates review work a lot and encourage us to have a more modular code)

I often find myself without much to add on a PR description besides what is there in the ticket description, because the issue is a small one and there is a one to one correspondence between the issue description of the problem and the motivation for the changes merged from the particular PR linked to it. In this case, the issue description is "duplicated content", and is only useful because we have the "issue" -> "PR" directed dependency graph that makes work trackable and searchable. But I don't think it's always the case. I think there are bigger pieces of work where it is desiderable for an issue to have more PRs linked to it. 
How do you group together PRs if not in issues? Would you only use issues that are specifically tagged epic? Or some other system I'm unaware of?

> * The idea of the automerge is just enabling the option to merge automatically a PR once all checks passed (CI checks + at least two reviews) so you don't have to manually do it. This option is not meant to be mandatory but adding convinience, just experimenting new features. If you don't like it no worries, is disabled by default

Okay, automerge is fine as long as we can chose wether we want it or not.

vdedios:
> I often find myself without much to add on a PR description besides what is there in the ticket description, because the issue is a small one and there is a one to one correspondence between the issue description of the problem and the motivation for the changes merged from the particular PR linked to it. In this case, the issue description is "duplicated content", and is only useful because we have the "issue" -> "PR" directed dependency graph that makes work trackable and searchable. But I don't think it's always the case. I think there are bigger pieces of work where it is desiderable for an issue to have more PRs linked to it.
How do you group together PRs if not in issues? Would you only use issues that are specifically tagged epic? Or some other system I'm unaware of?

1. Agree with avoid duplicating information in PR description if there is already a defined issue behind it.
2. If there are big issues which should be attempted in several PRs, having, at least, a small comment in the PR with some information about which part of that issue is being achieved should be ok.
    Anyway, if the issue is VERY big I believe we should review as a team if it could be splitted. If there is no way we can do that, what you mentioned is good to me :ok_hand:

Damrod (approved):
LGTM

PR #37, opened 2022-05-01
#19/authentication flow with 42oauth

### LINKED ISSUE #19: Create a complete oauth flow with 42 API
- [ ] #23 
- [ ] #24

### LINKED ISSUE #23: Create a route to authenticate users through oauth

### LINKED ISSUE #24: Parse and validate response from oauth

gbudau:
A few steps to test this PR:
Add a new application on the intranet (Settings/API/Register a new app)
Set the redirect URIs to
```
http://127.0.0.1:3000/api/v1/oauth42
http://127.0.0.1/api/v1/oauth42
```

Copy the `.env.sample` file to `.env`
Change the `FORTYTWO_APP_ID` to the app `UID`
Change the `FORTYTWO_APP_SECRET` to the app `SECRET`

Run with `docker-compose up -d --build`, and access http://127.0.0.1:3000/api/v1/oauth42

```mermaid
sequenceDiagram
    participant U as User
    participant T as transcendence-app
    participant A as 42 API
    T->>A: Request token
    U->>A: User login & consent
    A->>T: Authorization code
    T->>A: Exchange code for token
    A->>T: Token response
    T->>A: Use token to call 42 API
```

closes #19
closes #23
closes #24


transcendence-app/src/users/users.service.ts
```diff
@@ -0,0 +1,36 @@
+import { Injectable } from '@nestjs/common';
+import { CreateUserDto } from './dto/create-user.dto';
+
+// This should be a real class/interface representing a user entity
+export type UserEntity = {
```
vdedios, line 5: Why do you need to redefine this type if you have already a class defined with this same info (createUserDto)

transcendence-app/src/auth/fortytwo/fortytwo.strategy.ts
```diff
@@ -0,0 +1,70 @@
...
+
+@Injectable()
+export class FortyTwoStrategy extends PassportStrategy(Strategy, '42') {
+  constructor(
+    private usersService: UsersService,
+    private httpService: HttpService,
+    protected configService: ConfigService<FortyTwoAuthConfig>,
+  ) {
+    super({
+      authorizationURL: 'https://api.intra.42.fr/oauth/authorize',
```
vdedios, line 21: I believe we should not hardcode urls or data prone to change, this should be in an env variable

transcendence-app/src/auth/fortytwo/fortytwo.strategy.ts
```diff
@@ -0,0 +1,70 @@
...
+      tokenURL: 'https://api.intra.42.fr/oauth/token',
```
vdedios, line 22: Same

transcendence-app/src/auth/auth.module.ts
```diff
@@ -0,0 +1,7 @@
+import { Module } from '@nestjs/common';
+import { FortyTwoModule } from './fortytwo/fortytwo.module';
+
+@Module({
+  imports: [FortyTwoModule],
+})
+export class AuthModule {}
```
vdedios, line 7: Why do we need this module? Why not having just 42 auth module for now?

transcendence-app/src/users/users.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  username: string;
+  email: string;
+  image_url: string;
+};
+
+@Injectable()
+export class UsersService {
+  private readonly users = [];
+
+  findOne(username: string) {
```
vdedios, line 17: This could be private method?

transcendence-app/src/users/users.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    const user = this.users.find((user) => user.username === username);
+    return user;
+  }
+
+  create(userDto: CreateUserDto) {
```
vdedios, line 22: This too?

vdedios (changes requested):
Despite having several comments, good work 😊

I'd also like to take this opportunity to make an observation. We already have at this point four defined modules: main, user, auth, fortyTwo. We still have 90% of the project to finish, how many modules can we end up having? Are we running the risk of spawning a Frankenstein?

transcendence-app/src/users/dto/create-user.dto.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  IsEmail,
+  IsInt,
+  IsNotEmpty,
+  IsPositive,
+  IsString,
+  IsUrl,
+} from 'class-validator';
+
+export class CreateUserDto {
```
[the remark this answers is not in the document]
fizcamposancos: From my point of view, this is a Dto used to create users and make an initial validation, so the name CreateUserDto sounds good to me.

transcendence-app/src/auth/fortytwo/fortytwo-config.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface FortyTwoAuthConfig {
```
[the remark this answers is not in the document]
fizcamposancos: I agree with @vdedios, and there is a auth module which just import 42 module. May only 42oAuth module is enough

transcendence-app/src/users/users.service.ts
fizcamposancos: I don't really know the standard way of working with these classes and Dtos. We have the Dto class to create and validate our user, but maybe we shoud store it in a UserEntity class. Maybe it is a duplicating a little but the purpose of those classes are different. But I would have to go deeper into the documentation or examples to give a better answer.

fizcamposancos:
A general question reagarding how Nestjs works. You've @gbudau  added some libraries like Joi, Axios, Passport, rxjs, ... And I guess they require some type of package installation (ie. `npm install --save @nestjs/passport passport passport-local`). But I haven't installed anything and everything works (in my ubuntu computer).
However, in my Ubuntu VM in my other computer I got tonnes of errors when I tried to execute it locally, without docker. Like this:
 `error TS2307: Cannot find module '@nestjs/passport' or its corresponding type declarations.
2 import { AuthGuard } from '@nestjs/passport';`
Do you know why?

transcendence-app/src/auth/fortytwo/fortytwo-config.interface.ts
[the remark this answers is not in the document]
Damrod: You cannot start symbol names with digits in many programming languages because of the ambiguity on wether symbols are symbols or literal numbers. I would prefer to follow it here, even if not strictly necessary (is it not necessary?).
TLDR, why have we been naming everything as ft_ rather than 42_ so far otherwise?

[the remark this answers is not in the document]
vdedios: `oauth42` :smile:

transcendence-app/src/users/users.service.ts
Damrod: why is this a type and not a class? Why is it commented as class/interface?

transcendence-app/src/app.module.ts
```diff
@@ -1,10 +1,20 @@
...
 
 @Module({
-  imports: [],
-  controllers: [AppController],
-  providers: [AppService],
+  imports: [
+    ConfigModule.forRoot({
+      isGlobal: true,
+      cache: true,
+      validationSchema: Joi.object({
```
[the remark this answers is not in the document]
Damrod: I have read [here](https://dev.to/yawaramin/comment/dfhg) that it's okay to mix and match. Typescript provides compile time validation, joi provides runtime validation.

But mostly I think we should let ourselves a bit of freedom to experiment and add random modules just for the sake of discovering them, even if they pollutes our code a bit.

If it enables you, I think it's okay to leave it there for now. If we don't like it, we can remove it later on.

That's my opinion.

fizcamposancos (commented):
I think I would need further explanation or more time to understand how the 42.strategy works. But anyway, very good work @gbudau . Well done!!!

transcendence-app/src/app.module.ts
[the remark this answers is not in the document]
gbudau: Done, I've replaced this package with the `class-validator` since we are also using it in other places.

As @Damrod says, it's nice to have compile-time and runtime validation.

transcendence-app/src/auth/fortytwo/fortytwo-config.interface.ts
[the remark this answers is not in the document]
gbudau: OAuth42 sound like a good name. Done :+1:

transcendence-app/src/auth/fortytwo/fortytwo.strategy.ts
gbudau: :+1:

transcendence-app/src/auth/auth.module.ts
gbudau: I guess we don't need it. I thought its a good idea to have an auth module, and inside that module, import other types of authentication modules.

```
./auth
  ./oauth42
  ./local (email/username and password)
  ./github
  ./etc.
```

transcendence-app/src/auth/fortytwo/fortytwo.strategy.ts
```diff
@@ -0,0 +1,70 @@
...
+    return validatedUser;
+  }
+
+  async validate(accessToken: string) {
+    const { data } = await lastValueFrom(
+      this.httpService.get('https://api.intra.42.fr/v2/me', {
+        headers: { Authorization: `Bearer ${accessToken}` },
+      }),
+    ).catch((err) => {
+      console.error(err.message);
```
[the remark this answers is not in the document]
gbudau: Done, I've replaced console logs with NestJS logger. I think for some cases, we could also log errors in production.

In this function, an error can happen if the 42 API is down, return the wrong information or if the URLs from the environment variables are incorrect.

gbudau:
> A general question reagarding how Nestjs works. You've @gbudau added some libraries like Joi, Axios, Passport, rxjs, ... And I guess they require some type of package installation (ie. `npm install --save @nestjs/passport passport passport-local`). But I haven't installed anything and everything works (in my ubuntu computer). However, in my Ubuntu VM in my other computer I got tonnes of errors when I tried to execute it locally, without docker. Like this: `error TS2307: Cannot find module '@nestjs/passport' or its corresponding type declarations. 2 import { AuthGuard } from '@nestjs/passport';` Do you know why?


Try removing the `node_modules` directory and installing them again.
```
rm -rf node_modules
npm clean-install
```
I think the error is from the TypeScript server of Visual Studio Code.

Docker installs packages inside the containers, and the project works without having to install anything.

But the TypeScript server looks for the modules inside the `node_modules` directories and does not find anything locally.

transcendence-app/src/users/users.service.ts
gbudau: This service is incomplete and more like a mock service. For now, the user entities are stored in an array.

A future goal is to store the users in a database.
We could inject a repository to access the database and manage user entities.

transcendence-app/src/users/dto/create-user.dto.ts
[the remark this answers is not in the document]
vdedios: Agree but two comments about this. The only place you are using this `userEntity` is, in fact, a param in the main controller methodof `oauth42`, not db related :\

Additionally, a `dto` is itself, as you pointed out, an object which carries data from process to process and just that. Having a `dto` named related to an action lacks of sense because `dto` is only that. If you want to create/update something and pass a dto just create a method for that. In your example, you could do something like:

```typescript
createUserDto(userName: string, password: string) {
    \\ some intermediate logic/validation you may need
    return UserDto(userName, password);
}
```

transcendence-app/src/users/users.service.ts
vdedios: Agree, but I think you are using dtos instead of entites for storing userData :see_no_evil:

transcendence-app/src/users/users.module.ts
```diff
@@ -0,0 +1,8 @@
+import { Module } from '@nestjs/common';
+import { UsersService } from './users.service';
+
+@Module({
+  providers: [UsersService],
+  exports: [UsersService],
+})
+export class UsersModule {}
```
vdedios, line 8: Check `user` module because you have inconsistent naming in your files, some things are named `user` while others `users`. I'd go with `user` (singular) since we can have one or many of them

transcendence-app/src/auth/fortytwo/fortytwo.strategy.ts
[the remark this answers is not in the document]
vdedios: Agree, having logs in production is very useful. And NestJS logger seems neat :ok_hand:

transcendence-app/src/users/users.service.ts
gbudau: Done

transcendence-app/src/users/users.service.ts
gbudau: :+1:

transcendence-app/src/users/users.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  externalProviderId: number;
+  username: string;
+  email: string;
+  image_url: string;
+};
+
+@Injectable()
+export class UsersService {
+  private readonly users = [];
+
```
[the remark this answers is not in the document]
gbudau: Great observation. Since this is a new project, it makes sense to enable safer checks.

I've added `"strict": true` to `tsconfig.json`

transcendence-app/src/users/users.module.ts
gbudau: Done

transcendence-app/src/users/users.service.ts
fizcamposancos: Are you sure this method, and the 'create' one can be private?
I tried to create a 'user' controller and tried to call this method from there, like we are going to do when we want to see one user info or to create a new user.
We can change it later when the controller is ready.

transcendence-app/src/users/dto/create-user.dto.ts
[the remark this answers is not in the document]
gbudau: Good point. I've removed the `UserEntity`, and we are using only the `UserDto`. We can add it back later if we need it.

transcendence-app/src/users/users.service.ts
gbudau: I've removed the user entity since there is no database yet. We can add it back later if we need it.

transcendence-app/src/users/users.service.ts
gbudau: I think in the context of this PR, they can be private. If we add some routes to the user controller which use these functions, we can make them public again.

vdedios: Agree with @fizcamposancos for now I'd stay with private and, if we need to use any of them, make public

gbudau:
> I'd also like to take this opportunity to make an observation. We already have at this point four defined modules: main, user, auth, fortyTwo. We still have 90% of the project to finish, how many modules can we end up having? Are we running the risk of spawning a Frankenstein?

I think it depends on the size of the application.
From the NestJS documentation:
>While very small applications may theoretically have just the root module, this is not the typical case. We want to emphasize that modules are strongly recommended as an effective way to organize your components. Thus, for most applications, the resulting architecture will employ multiple modules, each encapsulating a closely related set of capabilities.

vdedios (approved):
Great job! Ready to integrate 🎉

gbudau:
@fizcamposancos, between the passport library and how NestJS interacts with this library, the way it works is a bit of magic :smile:

Here is how I think it works(I might be wrong or miss some steps):

When we access a route that has `OAuth42Guard`

The guard redirects to the 42 authorization URL, passing some query parameters, like id, secret, callback URL, etc.

Then the user logs with his account and gets redirected to the callback URL (in our case, the callback URL is the same, but we can have a different URL), this time with a code in the query parameters.

This code is the access_token.

If there is a code parameter, the route does not redirect, and it will execute the `validate` function (from the `OAuth42Strategy` class)

In this function, we call the profile endpoint of the 42 API using the `access_token`, and we get the user information.

Then we validate this information and return the user.

The user (or whatever the `validate` function returns) gets added to the request object as a user parameter.

The last step is executing the `oauth42Login` function from the `OAuthController`, which returns the user from the request body.

fizcamposancos (approved):
Very good work and very good explanations. Well done!!!

gbudau:
Thank you for the suggestions and feedback! :bow:

PR #41, opened 2022-05-15
Update OAuth42Guard

### LINKED ISSUE #40: Catch exceptions thrown by PassportJS
Currently, if PassportJS throws an exception, NestJS catches it and returns a 500 Internal Server Error.

We could intercept these exceptions and return another HTTP status code.

gbudau:
Add an override for the `handleRequest` method.

Throw `BadGatewayExeption` if an exception occurs from `OAuth42Strategy` or if the validate function returns a falsy value instead of a user.

We can remove the try/catch from `OAuth42Strategy` because this method will catch any errors.

closes #40

### DIFF
--- a/transcendence-app/src/oauth42/oauth42.guard.ts
+++ b/transcendence-app/src/oauth42/oauth42.guard.ts
@@ -1,5 +1,12 @@
-import { Injectable } from '@nestjs/common';
+import { BadGatewayException, Injectable } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 
 @Injectable()
-export class OAuth42Guard extends AuthGuard('oauth42') {}
+export class OAuth42Guard extends AuthGuard('oauth42') {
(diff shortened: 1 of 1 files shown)


vdedios (approved):
Nice! You mention about removing try/catch but finally didn't? 🤔

Damrod:
> Nice! You mention about removing try/catch but finally didn't? thinking

+1

Damrod (approved):
LGTM

gbudau:
> Nice! You mention about removing try/catch but finally didn't? :thinking:

@vdedios, good catch, thank you. I didn't remove it in this PR because we will use the `Api42Service` in the future, and we can make that change then (to not change it twice)

Please see #43 for an example of how it could look.

fizcamposancos (approved):
Everything seems OK.

gbudau:
Thank you!

Issue #25, opened 2022-04-27
Create a mechanism to create new users and check if the user is already registered


gbudau:
For example, normal authentication flow, validate if userId exists with fetched info?

- [x] #26 
- [x] #27 
- [x] #28
- [x] #29 
- [x] #30
- [x] #31

gbudau:
I think we can close this task because all the subtasks are complete :rocket:

PR #42, opened 2022-05-16
fetch user with id

Damrod:


gbudau (commented):
Thanks for this feature!

transcendence-app/src/user/dto/user.dto.ts
```diff
@@ -1,6 +1,8 @@
 import { IsEmail, IsNotEmpty, IsString, IsUrl } from 'class-validator';
 
 export class UserDto {
+  id: number | undefined;
+
```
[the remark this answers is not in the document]
Damrod: I don't have any particular preference on which of these two do we like.
If we had users other than from 42, I'd say it makes more sense to use different Ids, otherwise, intranet Id works just as well.
I'll be trying the second one.

transcendence-app/src/user/user.controller.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { UserController } from './user.controller';
+import { HttpStatus, INestApplication } from '@nestjs/common';
+import * as request from 'supertest';
+import { UserModule } from './user.module';
+
+describe('User Controller', () => {
```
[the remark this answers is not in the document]
Damrod: We don't need the app.init() anymore? What about afterAll (() => app.close())?
Does the above test you wrote replace the one I wrote? Or is additional? Or is it just an example to illustrate what do you mean by mockUserService?

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.userService.findOneOrCreate(user);
+    return user;
+  }
+
+  @Get(':id')
+  async getUserById(@Param('id') param: string): Promise<UserDto | undefined> {
+    const result: UserDto | undefined = this.userService.retrieveUserWithId(
+      Number(param),
+    );
+    if (result === undefined) throw new NotFoundException();
```
Damrod, line 34: I had to force push because I rebased, and now your comment is gone, but I heeded it.

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+    });
+  }
+
+  @Get(':id')
+  async getUserById(@Param('id') param: string): Promise<UserDto | undefined> {
+    const result: UserDto | undefined = this.userService.retrieveUserWithId(
+      Number(param),
+    );
+    if (result === undefined)
+      throw new HttpException('Not found', HttpStatus.NOT_FOUND);
```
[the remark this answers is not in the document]
Damrod: Done (I had to rebase, so we lost the reference to this diff, but I made the change).

transcendence-app/src/user/dto/user.dto.ts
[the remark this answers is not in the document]
Damrod: But what about we do this in a following PR?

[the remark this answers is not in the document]
gbudau: Sure, as you see fit.

gbudau (approved):
Well done :rocket: LGTM!

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+      image_url: 'www.img_url.com/hello.jpg',
+      username: `user_${id}`,
+      email: 'afgv@github.com',
+    };
+    this.userService.findOneOrCreate(user);
+    return user;
+  }
+
+  @Get(':id')
+  async getUserById(@Param('id') param: string): Promise<UserDto | undefined> {
```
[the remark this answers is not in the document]
vdedios: Agree, and, for other scenarios we should never return undefined, maybe nullable

transcendence-app/src/user/dto/user.dto.ts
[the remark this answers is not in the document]
vdedios: I really align with the idea of getting the userId from the db, we have to control our data and a `userId` is a very sensitive one. If we use the one coming from 42API we'll be limiting ourselves in many ways:

what if we decide to implement another type of logging? what if userId intra42 specification changes?

Due to this, I'd genereate our own ids and, remove the `userId` from the dto. If we follow this approach it'll be useless

transcendence-app/src/user/api42.service.ts
```diff
@@ -42,6 +42,7 @@ export class Api42Service {
   async get42UserData(accessToken: string): Promise<UserDto> {
     const data = await this.fetch42UserData(accessToken);
     const user: UserDto = {
+      id: undefined,
```
vdedios, line 45: As said in a comment bellow, If we all agree in getting the id from the db, I'll just remove this

transcendence-app/src/user/user.controller.spec.ts
```diff
@@ -0,0 +1,93 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { UserController } from './user.controller';
+import { HttpStatus, INestApplication } from '@nestjs/common';
+import * as request from 'supertest';
+import { UserModule } from './user.module';
+import { UserService } from './user.service';
+import { UserDto } from './dto/user.dto';
+import { NotFoundException } from '@nestjs/common';
+
+describe('User Controller end to end test', () => {
```
vdedios, line 10: Nice you've added some tests :clap:

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+  Post,
+} from '@nestjs/common';
+import { UserService } from './user.service';
+import { UserDto } from './dto/user.dto';
+
+@Controller('user')
+export class UserController {
+  constructor(private userService: UserService) {}
+
+  @Post(':id')
```
vdedios, line 15: when did we decide to have an endpoint to create users? :smile:

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+  async addUser(@Param('id') param: string): Promise<UserDto> {
+    const id = Number(param);
+    const user: UserDto = {
+      id: id,
+      provider: 'ft_transcendence',
+      image_url: 'www.img_url.com/hello.jpg',
+      username: `user_${id}`,
+      email: 'afgv@github.com',
+    };
```
vdedios, line 24: What is the point of hard-coding all this data? if they are some mocked values, where do you expect they come from?

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.userService.findOneOrCreate(user);
+    return user;
+  }
+
+  @Get(':id')
+  async getUserById(@Param('id') param: string): Promise<UserDto | undefined> {
+    const result: UserDto | undefined = this.userService.retrieveUserWithId(
```
vdedios, line 31: I believe typescript inferes the type when you assign a value when declaring the variable, so maybe not needed to add the type of the variable here

vdedios (approved):
Overall, great work!! I am really glad to see we are getting this rolling :guitar:

transcendence-app/src/user/api42.service.ts
Damrod: Well, I would say that the whole user.service, with it's internal UserDTO[], is a mock that we are using as a temporary workaround standing in for the missing DB connection. This means that as you implement the internals of the DB connection you are expected to completely change and remove many parts of this.

transcendence-app/src/user/dto/user.dto.ts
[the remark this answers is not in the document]
Damrod: It's temporary. I intend to do this
```
Or create a new class UserEntity/UserModel, which will map to the User table inside the database and add all the additional fields (like id) to that class.
```
in a follow up PR.

transcendence-app/src/user/user.controller.ts
Damrod: same thing. You have assigned yourself to all of the PRs that implement the [addition of users](https://github.com/42AGV/ft_transcendence/issues/25). What do I do? Sit on my hands?
Just change it when you have a comprehensive solution.

transcendence-app/src/user/user.controller.ts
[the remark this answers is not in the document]
Damrod: fixed.

transcendence-app/src/user/user.controller.ts
Damrod: fixed.

transcendence-app/src/user/dto/user.dto.ts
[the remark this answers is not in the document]
Damrod: The follow up PR is [this one](https://docs.nestjs.com/techniques/database).

transcendence-app/src/user/user.controller.ts
vdedios: Didn't had the context of why you were adding this method and was just questioning it in general, as a new controller added to our system. Now that I have this info, let me explain myself properly :smiley:

An endpoint to add user data to our system expects the client to provide this information. A method like this one is about adding a new user that don't exist yet in our system, this data won't come from the DB. So no need to have anything mocked at this point despite not having any DB implementation ready yet. That is the reason I was asking _what was the point of hard-coding the data at this point_ and also asking _where did you expected this data to come from_.

That user data would have to come from the client and passed, for example, as a param to the controller method, something like:

```ts
  async addUser(@Param('user') param: UserDto): Promise<UserDto> {...}
```
Mocking that UserDto information at this point makes no sense because we are not mocking anything but hardcoding what should come from client.

We should mock this inside the tests (which is the desirable place to mock things as much as we can do it) Something like:

```ts
    await request(server).post({
          id: 123,
          provider: user.provider,
          username: user.provider,
          email: user.email,
          image_url: user.image_url,
        }).expect(HttpStatus.CREATED);
    await request(server).get('/user/1').expect(HttpStatus.OK);
```

That being said, since this method is just  for testing purposes and prone to change I'll just merge it. Just wanted to explain properly my point of view

fizcamposancos (approved):
Considering all the comments from the team, and knowing that this would change once we have our db up and running, that's right for me. Good job!!

transcendence-app/src/user/user.controller.ts
Damrod: I will do this in the MR I already have open with the user entity.

PR #45, opened 2022-05-19
Little version of MVP backend

fizcamposancos:
A little version of MVP back-end with some endpoints, db connected with TypeORM, sessions saved into the db, guard for routes based on the sessions (only in the findall() method to check it).

[http://127.0.0.1:3000/](http://127.0.0.1:3000/) has login button to authenticate using the OAuth system of 42

Then a cookie is created and will be available for 5 minutes to access all users at [http://127.0.0.1:3000/api/v1/user](http://127.0.0.1:3000/api/v1/user). All the other endpoint are not protected, just to try.

This PR is not aimed to be merged, just to "play" with it and mock some services while there definite version is not available, ie, direct SQL database, not through TypeORM.

 No tests added yet.

### DIFF
--- a/transcendence-app/src/app.module.ts
+++ b/transcendence-app/src/app.module.ts
@@ -2,6 +2,9 @@ import { Module } from '@nestjs/common';
 import { ConfigModule } from '@nestjs/config';
 import { OAuth42Module } from './oauth42/oauth42.module';
 import { validate } from './config/env.validation';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { UserEntity } from './typeorm/user.entity';
+import { SessionEntity } from './typeorm/session.entity';
 
 @Module({
   imports: [
@@ -11,6 +14,16 @@ import { validate } from './config/env.validation';
       validate,
     }),
     OAuth42Module,
+    TypeOrmModule.forRoot({
+      type: 'postgres',
+      host: 'db',
+      port: 5432,
+      username: 'transcendence',
+      password: 'example',
+      database: 'transcendence',
+      entities: [UserEntity, SessionEntity],
+      synchronize: true,
+    }),
   ],
 })
 export class AppModule {}
--- a/transcendence-app/src/typeorm/session.entity.ts
+++ b/transcendence-app/src/typeorm/session.entity.ts
@@ -1,0 +1,15 @@
+import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
+import { ISession } from 'connect-typeorm';
+
+@Entity({ name: 'sessions' })
+export class SessionEntity implements ISession {
+  @Index()
+  @Column('bigint')
+  expiredAt = Date.now();
+
+  @PrimaryColumn('varchar', { length: 255 })
+  id = '';
+
+  @Column('text')
+  json = '';
+}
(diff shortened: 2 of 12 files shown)


transcendence-app/src/app.module.ts
```diff
@@ -11,6 +14,16 @@ import { validate } from './config/env.validation';
       validate,
     }),
     OAuth42Module,
+    TypeOrmModule.forRoot({
+      type: 'postgres',
+      host: 'db',
+      port: 5432,
+      username: 'transcendence',
+      password: 'example',
```
gbudau, line 22: suggestion:  We could use the `ConfigService` here and add this info to a .env file or environment variables.

gbudau (approved):
Thank you and good job!

transcendence-app/src/app.module.ts
fizcamposancos: Yes, good point. If we finally decide to use TypeORM we can do that or even we can call a config file to clean the app.module with `TypeOrmModule.forRoot(config.typeorm.ts)` and in that file add the configuration with environment variables as you mentioned.

vdedios (approved):
Nice work! I like to see a functional mvp with typeorm in action to get a perspective of how it works. Thanks :)

Damrod:
If we finally decide to merge, we'll need to fix the merge conflicts. Could you please merge main into this branch and solve all of the merge conflicts, please?

transcendence-app/src/typeorm/session.entity.ts
```diff
@@ -0,0 +1,15 @@
+import { Column, Entity, Index, PrimaryColumn } from 'typeorm';
+import { ISession } from 'connect-typeorm';
+
+@Entity({ name: 'sessions' })
+export class SessionEntity implements ISession {
+  @Index()
+  @Column('bigint')
+  expiredAt = Date.now();
```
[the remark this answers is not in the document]
fizcamposancos: It was just to set a Date value, no reason behind it

fizcamposancos:
> 

Right now, database configuration is already done by Victor, sessions with Memcached and guards by George. The only difference remaning is TypeOrm. If we finally decide to implement it It would be easier to do that directly from the main branch and just close this PR.

transcendence-app/src/user/user.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+  findOne(@Param('username') username: string) {
+    return this.userService.findOne(username);
+  }
+
+  @Get()
+  @UseGuards(RouteGuard)
+  findAll() {
+    return this.userService.findAll();
+  }
+
```
Damrod, line 28: We decided to put 'provider' as a field in the UserDto. Why did you decide we don't want it anymore? I can think of some reasons, but I'd like to know your opinion.

transcendence-app/src/user/user.service.ts
```diff
@@ -1,27 +1,45 @@
...
+    private readonly userRepository: Repository<UserEntity>,
+  ) {}
+
   private readonly users: UserDto[] = [];
 
-  private findOne(username: string) {
-    const user = this.users.find((user) => user.username === username);
-    return user;
+  async findOne(username: string) {
+    return await this.userRepository.findOne({ username: username });
```
Damrod, line 17: We could shortcircuit to a null as gbudau did. Why is null better than undefined? I don't know. I'll ask there.

Damrod (approved):

fizcamposancos:
Just close it, for the time being

PR #46, opened 2022-05-22
#27 DB model and setup

### LINKED ISSUE #27: Create a model and db table to be able to persist userInfo
User ID must be unique, think about complex queries, filtering data, indexing, etc

vdedios:
closes #27


docker-compose.yml
```diff
@@ -21,8 +21,11 @@ services:
   db:
     image: postgres:14.2-alpine3.15
     restart: always
+    user: postgres
     volumes:
       - transcendence-db:/var/lib/postgresql/data
+      - ./db/fill-db.sh:/docker-entrypoint-initdb.d/fill-db.sh
```
vdedios, line 27: https://levelup.gitconnected.com/creating-and-filling-a-postgres-db-with-docker-compose-e1607f6f882f

db/fill-db.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Simple script to ensure db and table creation before filling up
```
vdedios, line 3: The idea of this script is, since official postgreSQL image entrypoint loads `.sh` and `.sql` files but with no particular order, ensure we setup db and tables before filling them up with mocked data or wathever we need

db/setup/schema.sql
```diff
@@ -0,0 +1,10 @@
+CREATE DATABASE ft_transcendence;
+
+\c ft_transcendence;
+
+CREATE TABLE IF NOT EXISTS users(
+    UUID        UUID DEFAULT gen_random_uuid()  PRIMARY KEY,
```
vdedios, line 6: I've gone with randomly generated UUID with a function already provided by postgreSQL -> https://www.cybertec-postgresql.com/en/uuid-serial-or-identity-columns-for-postgresql-auto-generated-primary-keys/

There are many more methods but this seemed the best to me

docker-compose.prod.yml
```diff
@@ -22,4 +22,4 @@ services:
     environment:
       - POSTGRES_USER=${POSTGRES_USER}
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
-      - POSTGRES_DB=${POSTGRES_DB}
+      - POSTGRES_DB=ft_transcendence
```
vdedios, line 25: since our db will be setup as `ft_transcendence` no need to read from env

db/setup/queries.sql
```diff
@@ -0,0 +1,41 @@
...
+
+INSERT INTO users (
+     username,
+     email,
+     img
+) VALUES (
+    'gbudau-admin',
+    'gbudau-admin@transcendence.com',
+    'https://upload.wikimedia.org/wikipedia/commons/8/8d/42_Logo.svg'
+);
```
vdedios, line 11: mocking some data

db/setup/schema.sql
```diff
@@ -0,0 +1,10 @@
...
+    USERNAME    VARCHAR(20)                     NOT NULL,
+    EMAIL       VARCHAR(50)                     NOT NULL,
```
gbudau, line 8: suggestion: We could make these fields unique.

gbudau (approved):
Thank you, good job!

db/setup/schema.sql
vdedios: Nice idea, I'll do it!

vdedios: done!

fizcamposancos (approved):
Nice job!!!

db/setup/schema.sql
Damrod: Is it possible (even if it's VERY unlikely) that we will get duplicated UUID?

Damrod (approved):
Looks good to me!

db/setup/schema.sql
vdedios: I think it is not very likely. But, with large set of data it may be possible -> https://stackoverflow.com/questions/47172263/many-duplicate-values-when-using-postgres-uuid-generate-v4

For our usecase I'd say it is good enough :)

PR #47, opened 2022-05-23
Feature/sessions

### LINKED ISSUE #26: Implement a cookie session mechanism to authorize users in our platform
Server sessions, JWT, etc.

### LINKED ISSUE #32: Create deauth flow (route)
Delete sessions/JWT and return OK response

### LINKED ISSUE #33: Create API endpoint to fetch data from current user
Some considerations:
- Route must be protected
- Route must be `/api/v1/users/me`

gbudau:
This PR implements server-side session management.

Rename the `oauth42` module to `auth`.
Add a Memcached service to store the sessions.
Add the `api/v1/auth/logout` route to log out a user.
Add the `api/v1/user/me` route to retrieve the info of the currently logged user.
Update the `api/v1/auth/oauth42` route to `api/v1/auth/login`.

Please change the redirect URLs of your intranet application to:
```
http://127.0.0.1:3000/api/v1/auth/login
http://127.0.0.1/api/v1/auth/login
```

And add the following environment variables to the `.env` file:
```
FORTYTWO_APP_CALLBACK_URL=http://127.0.0.1:3000/api/v1/auth/login
MEMCACHED_HOST=memcached
MEMCACHED_PORT=11211
SESSION_SECRET= UUID format, generate with uuidgen
MEMCACHED_SECRET= UUID format, generate with uuidgen
```

closes #26
closes #32 
closes #33


transcendence-app/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  ApiFoundResponse,
+  ApiTags,
+} from '@nestjs/swagger';
+import { Request } from 'express';
+import { OAuth42Guard } from './oauth42.guard';
+
+@ApiTags('auth')
+@Controller('auth')
+export class AuthController {
+  @Get('oauth42')
```
fizcamposancos, line 18: Suggestion: There is a 'logout' endpoint, what if we call 'oauth42' endpoint 'login'?

fizcamposancos, line 18: Suggestion: There is a 'logout' endpoint, what if we call 'oauth42' endpoint 'login'?

transcendence-app/src/user/user.controller.ts
```diff
@@ -30,10 +38,15 @@ export class UserController {
     return this.userService.findOneOrCreate(query);
   }
 
+  @Get('me')
+  getCurrentUser(@GetRequest() req: Request) {
+    return req.user;
+  }
+
```
fizcamposancos, line 45: Despite this issue was assigned to me , very nice job. 😄

fizcamposancos, line 45: Despite this issue was assigned to me , very nice job. 😄

transcendence-app/src/user/user.controller.ts
```diff
@@ -30,10 +38,15 @@ export class UserController {
...
   @Get(':id')
   async getUserById(@Param('id') param: string): Promise<UserEntity> {
     const result = this.userService.retrieveUserWithId(Number(param));
-    if (result === undefined) throw new NotFoundException();
+    if (result === null) throw new NotFoundException();
```
fizcamposancos, line 49: Nice solution, now I don't get unexpected 500 internal errors

fizcamposancos, line 49: Nice solution, now I don't get unexpected 500 internal errors

fizcamposancos (approved):
Very nice implementation of sessions and guards. I would have to learn more about memcached and middlewares fully undestand how them work.
Nice job

transcendence-app/src/user/user.controller.ts
gbudau: Thank you! I've added this endpoint to test if sessions are working correctly.

transcendence-app/src/app.module.ts
```diff
@@ -10,7 +13,29 @@ import { validate } from './config/env.validation';
...
+    const MemcachedStore = ConnectMemcached(session);
+
+    consumer
+      .apply(
+        session({
+          secret: this.configService.get('SESSION_SECRET') as string,
+          resave: false,
+          saveUninitialized: false,
+          store: new MemcachedStore({
+            hosts: ['memcached:11211'],
```
Damrod, line 32: Why is it not a good idea in this context to store this string in an environmental variable?

transcendence-app/src/config/env.validation.ts
```diff
@@ -31,6 +31,14 @@ class EnvironmentVariables {
 
   @IsUrl()
   FORTYTWO_APP_PROFILE_URL!: string;
+
+  @IsString()
+  @IsNotEmpty()
```
Damrod, line 36: Could we make this a bit more specific? I wouldn't demand it now, but 'a' something like this would work, and probably it shouldn't.
A bit more specific specifying something like a minimum length, of IsBase64 encoded, or is Hexadecimal, or something like this?

transcendence-app/src/user/user.controller.spec.ts
```diff
@@ -6,32 +6,53 @@ import { UserModule } from './user.module';
...
 
   it('should create a user', async () => {
     const server = await app.getHttpServer();
-    await request(server)
-      .post(
-        '/user/new?provider=42&username=afgv&email=hello@example.com&image_url=www.example.jpg',
-      )
+    const response = await request(server)
+      .post('/user')
+      .send(testUserDto)
```
Damrod, line 41: Is there a way to send this manually for testing purposes on some browser (e.g. firefox)? How do you manually set the body of the request to the Dto that you want to mock? How do you serialize that object? Why is it better than the query?

transcendence-app/src/user/user.service.ts
```diff
@@ -1,21 +1,22 @@
...
 import { UserEntity } from './user.entity';
+import { v4 as uuidv4 } from 'uuid';
 
 @Injectable()
 export class UserService {
   private readonly users: UserEntity[] = [];
 
   private findOne(username: string) {
-    return this.users.find((user) => user.Dto.username === username);
+    return this.users.find((user) => user.username === username) || null;
```
Damrod, line 11: Why is it better to return null than undefined?

Damrod (approved):
Chapeau, as usual :+1:

transcendence-app/src/config/env.validation.ts
gbudau: Good point. I've changed it to UUIDs. Or we could change to other formats. Please let me know what you think.

transcendence-app/src/user/user.controller.spec.ts
vdedios: For me, Postman is usually the way to go

transcendence-app/src/user/user.service.ts
vdedios: `null` is an intentional lack of content (controlled) `undefined` is a non-intentional lack of content (not-controlled)

transcendence-app/src/user/user.service.ts
```diff
@@ -1,21 +1,22 @@
...
   }
 
-  retrieveUserWithId(id: number) {
-    return this.users.find((user) => user.id === id);
+  retrieveUserWithId(id: string) {
+    return this.users.find((user) => user.id === id) || null;
```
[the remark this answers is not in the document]
vdedios: Yup, `UUID` is an alphanumerical string. TS have lots of types, maybe it also have an `UUID` type, I haven't checked it

transcendence-app/src/user/user.entity.ts
```diff
@@ -1,19 +1,7 @@
...
-    this.id = UserEntity.lastUserId++;
-  }
-
-  @IsPositive()
-  @IsInt()
-  id: number;
-
-  @ValidateNested()
-  Dto: UserDto;
+export interface UserEntity {
```
vdedios, line 1: Now that we have a DB model established, maybe we can update this to represent users table?

vdedios (approved):
Nice!!

The only objection I have is I wouldn't have changed oauth42 to auth because, despite only having this login maybe we want to implement another type so our friends and family can use our incredible app. Just kidding, but if we ever have this scenario we can update it later. LGTM!

transcendence-app/src/user/user.service.ts
Damrod: Obviously I understand the concept as in C, and it would definitely be my approach in C (because there is not undefined as a value/type) but I'm not sure it's the best solution here, (except for the explanation from gbudau about deserializeUser).

Alas, the method we are wrapping (.find() ) returns an undefined, because there is no defined member of this array that corresponds to the key we are trying to match. We are changing the semantics when we decide to return a null instead.

transcendence-app/src/user/user.service.ts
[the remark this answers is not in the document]
Damrod: UUID is data. The type depends on how do we want to store it. A string the matching of which is not case sensitive, with valid characters on the range of [0-9] and [a-f], can be understood, and stored, as a number, that was transmitted in hex form, and then streamed as appropriate (as a string for instance) if needed.

Just as a string with valid characters in the range [0-9] can always be stored as a number.

I don't know if this is the case in the uuids we are using.

transcendence-app/src/config/env.validation.ts
Damrod: lgtm

transcendence-app/src/user/user.entity.ts
gbudau: Thank you! I've made the change.

I've also renamed the `uuid` column to `id` and the `createdat` to `created_at` and added a default value. Please let me know what you think.

gbudau:
Thank you all!

Issue #34, opened 2022-04-29
Create API endpoint to fetch a list of other users


vdedios:
Some considerations:

- Route must be protected
- Route must be /api/v1/users
- Should research about ways to filter users. For example, what criteria we use to sort returned users, how many users do we want to return. Use something like pagination?

PR #48, opened 2022-05-25
#34 fportela/fetch users 

fizcamposancos:
#34 Create API endpoint to fetch a list of other
Two endpoints added:
[http://127.0.0.1:3000/api/v1/users](http://127.0.0.1:3000/api/v1/users) which return all users
[http://127.0.0.1:3000/api/v1/users/?limit=4&offset=0&sort=true](http://127.0.0.1:3000/api/v1/users/?limit=2&offset=0&sort=true) which return an array of usernames, maximun 20, where you must declare how many you want:
limit=number, how many you want to skip from the begining,
offset = number, and if you want the result to be sorted,
sort = boolean.
Once we have the db ready, and not an array, we can decide what we really want to retrieve.
No test added yet, sorry.


fizcamposancos:
Sorry, I've just realized that my start-point was gbudau's sessions branch. If you want me to fix it, please let me know

transcendence-app/src/user/user.controller.ts
```diff
@@ -33,6 +36,21 @@ export class UserController {
...
+  @Get('users')
+  getUsers() {
+    return this.userService.retrieveUsers();
+  }
+
+  @Get('usernames?')
+  getUsernames(
+    @Query('take', ParseIntPipe) take: number,
+    @Query('skip', ParseIntPipe) skip: number,
+    @Query('sort', ParseBoolPipe) sort: boolean,
```
vdedios, line 48: Could you make a brief explanation about what this params represent?

transcendence-app/src/user/user.controller.ts
```diff
@@ -33,6 +36,21 @@ export class UserController {
...
+  ) {
+    take = take > 20 ? 20 : take;
```
vdedios, line 50: This logic is a little bit weird and I'd say no to have magic numbers -> https://en.wikipedia.org/wiki/Magic_number_(programming)

transcendence-app/src/user/user.service.ts
```diff
@@ -14,6 +14,20 @@ export class UserService {
   retrieveUserWithId(id: string) {
     return this.users.find((user) => user.id === id) || null;
   }
+  retrieveUsers() {
+    return this.users;
+  }
+  retrieveUsernames(take: number, skip: number, sort: boolean): string[] {
+    const ret: string[] = [];
+    if (take < 0 || skip < 0 || skip >= this.users.length)
+      throw new BadRequestException('Provided values incorrect');
```
vdedios, line 23: Despite we are very used to not use curly braces in 42 in order to save function lines (:sweat: ) it is an industry standard to always surround functions with them because improves legibility and clarifies statements scope.
```suggestion
    if (take < 0 || skip < 0 || skip >= this.users.length) {
      throw new BadRequestException('Provided values incorrect');
    }
```

transcendence-app/src/user/user.service.ts
```diff
@@ -14,6 +14,20 @@ export class UserService {
...
+    take = skip + take > this.users.length ? this.users.length : skip + take;
+    for (let i = skip; i < take; i++) {
+      ret.push(this.users[i].username);
+    }
+    if (sort) return ret.sort();
```
vdedios, line 28: same about braces
```suggestion
    if (sort) {
     return ret.sort();
    }
```

transcendence-app/src/user/user.service.ts
```diff
@@ -14,6 +14,20 @@ export class UserService {
...
+    return ret;
```
vdedios, line 29:
```suggestion
     return sort ? ret.sort() : ret;
```

transcendence-app/src/user/user.controller.ts
Damrod: Normally you name them appropriately (which is hard), assign them to some constant/define/enum, and use the named constant/define/enum instead.
To me, here, it seems like a "MAX_USER_ENTRIES_PER_PAGE" or "MaxUserEntriesPerPage" (depending on our naming convention which I'm not sure about right now) kind of thing (I'm guessing)?
If you are doing a max, don't hand roll it with a ternary, use Math.max() and such.

transcendence-app/src/user/user.controller.ts
fizcamposancos: Of course.
take is a number which represent the amount of register you want to fetch.
skip is also a number representing the starting point of the retrieve
sort is a boolean to indicate if you want the results order arranged.
For example:
[http://127.0.0.1:3000/api/v1/user/usernames?take=3&skip=1&sort=true](http://127.0.0.1:3000/api/v1/user/usernames?take=3&skip=1&sort=false) would return 3 users, skipping the first entry, and as they are in the array (database), without being sorted.

transcendence-app/src/user/user.controller.ts
fizcamposancos: You are right. I correct it.

transcendence-app/src/user/user.service.ts
fizcamposancos: Perfect. I take note for future commits

transcendence-app/src/user/user.controller.ts
```diff
@@ -1,39 +1,63 @@
...
+
+  @Get('users')
+  getUsers() {
+    return this.userService.retrieveUsers();
+  }
+
+  @Get('usernames?')
+  getUsernames(
+    @Query('take', ParseIntPipe) take: number,
+    @Query('skip', ParseIntPipe) skip: number,
```
gbudau, line 48: suggestion: We could rename `take` and `skip` to `limit` and `offset`, because they are already part of [SQL](https://www.postgresql.org/docs/9.3/queries-limit.html) syntax.

transcendence-app/src/user/user.controller.ts
```diff
@@ -1,39 +1,63 @@
...
+    @Query('sort', ParseBoolPipe) sort: boolean,
+  ) {
+    take = take > MAX_USER_ENTRIES_PER_PAGE ? MAX_USER_ENTRIES_PER_PAGE : take;
```
gbudau, line 51: suggestion: We could remove this max limit and let the callers of the API decide the limit depending on their needs.

gbudau (commented):
Great job, and this is a good starting point.

transcendence-app/src/user/user.controller.ts
vdedios: I don't know if there is a max retrieval we could configure in DB but I'd say is a good point to limit queries sizes since this directly affects performance of our system

transcendence-app/src/user/user.service.ts
```diff
@@ -14,6 +14,20 @@ export class UserService {
...
+  }
+  retrieveUsernames(take: number, skip: number, sort: boolean): string[] {
+    const ret: string[] = [];
+    if (take < 0 || skip < 0 || skip >= this.users.length) {
+      throw new BadRequestException('Provided values incorrect');
+    }
+    take = skip + take > this.users.length ? this.users.length : skip + take;
+    for (let i = skip; i < take; i++) {
+      ret.push(this.users[i].username);
+    }
```
vdedios, line 28: Lets do it the good js way :yum:

```suggestion
    ret = this.users
      .filter((user, i) => i >= skip && i < take)
      .map(user => user.name);
```

vdedios (changes requested):
Requesting changes due to:
1. having duplicated code from the already @gbudau merged PR
2. strongly believe we should only have one single @GET(/users) endpoint with the possibility to filter them or not

transcendence-app/src/user/user.controller.ts
```diff
@@ -1,39 +1,63 @@
...
+    return req.user;
+  }
+
+  @Get('users')
+  getUsers() {
+    return this.userService.retrieveUsers();
+  }
+
+  @Get('usernames?')
+  getUsernames(
```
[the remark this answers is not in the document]
fizcamposancos: I agree with you. I'm going to change to a single endpoint with optional queryParams.
Regarding usernames or users, I was thinking in what we are going to show in the front-end, and I imagined a list of usernames, but I don't really know which is the normal procedure, to return all the user-data from the back-end and filter it in the front-end or just return the list of usernames from the back-end. That was my point. Maybe I should have asked it before.

transcendence-app/src/user/user.controller.ts
fizcamposancos: So, should we keep this max limit or remove it?

docker-compose.yml
```diff
@@ -31,5 +31,8 @@ services:
   memcached:
     image: memcached:1.6.15-alpine3.15
 
+  memcached:
+    image: memcached:1.6.15-alpine3.15
+
```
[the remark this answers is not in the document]
fizcamposancos: Fixed

transcendence-app/src/user/user.service.ts
fizcamposancos: Good js solution. I'll try something like that

fizcamposancos:
I will try to add some additional test to do more checks

vdedios (approved):
Nice, thanks for working on this! :rocket:

transcendence-app/src/user/user.controller.ts
vdedios: what do you think @gbudau? I haven't researched how to solve this but I think is not a good idea to let the client request as much data as it wants :cry: Maybe there's a cleaner way we can solve this

gbudau: Great point! It's a valid and a good reason to limit the query size for performance reasons :+1:

gbudau (approved):
Thank you for this feature! Looks good to me!

transcendence-app/src/user/dto/user.pagination.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+  @IsOptional()
+  @IsInt()
+  @Min(0)
+  @Max(MAX_USER_ENTRIES_PER_PAGE)
+  limit!: number;
+
+  @IsOptional()
+  @IsInt()
+  @Min(0)
+  @Max(MAX_USER_ENTRIES_PER_PAGE)
```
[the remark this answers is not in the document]
fizcamposancos: You are right, I remove the second @Max() decorator

transcendence-app/src/user/user.controller.ts
```diff
@@ -28,12 +30,16 @@ export class UserController {
     return this.userService.findOneOrCreate(user);
   }
 
-  @ApiBody({ type: UserEntity })
   @Get('me')
   getCurrentUser(@GetRequest() req: Request) {
     return req.user;
   }
 
+  @Get('users')
```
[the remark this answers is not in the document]
fizcamposancos: It seems a good idea to me, I change it.

transcendence-app/src/user/user.service.ts
```diff
@@ -14,6 +18,20 @@ export class UserService {
   retrieveUserWithId(id: string) {
     return this.users.find((user) => user.id === id) || null;
   }
+  retrieveUsers(queryDto: UsersPaginationQueryDto) {
+    if (queryDto.limit || queryDto.offset) {
+      let ret: string[] = [];
```
Damrod, line 23: > Variable initializer is redundant

You don't need to initialize with [] when the variable is already declared as an array.

Damrod (approved):
Sorry, I was just playing with this, and noticed what I'm mentioning below.

Damrod:
> #34 Create API endpoint to fetch a list of other Two endpoints added: http://127.0.0.1:3000/api/v1/user/users which return all users [http://127.0.0.1:3000/api/v1/user/usernames?take=4&skip=0&sort=true](http://127.0.0.1:3000/api/v1/user/usernames?take=2&skip=0&sort=true) which return an array of usernames, maximun 20, where you must declare how many you want, take=number, how many you want to skip from the begining, skip = number, and if you want the result to be sorted, sort = boolean. Once we have the db ready, and not an array, we can decide what we really want to retrieve. No test added yet, sorry.

Can you please update the endpoints you link in the description of the issue?

transcendence-app/src/user/user.service.ts
```diff
@@ -14,6 +18,20 @@ export class UserService {
...
+      queryDto.limit = queryDto.limit
+        ? queryDto.limit
+        : MAX_USER_ENTRIES_PER_PAGE;
+      queryDto.offset = queryDto.offset ? queryDto.offset : 0;
+      ret = this.users
+        .filter((user, i) => i >= queryDto.offset && i < queryDto.limit)
```
Damrod, line 29: A query like this:
`http://127.0.0.1:3000/api/v1/users?limit=20&offset=19&sort=true`

Shows only one username: the 19th one. How do we show any username at index past 19? Can you show me a query that does that? If I set limit higher than 20 it returns 400 Bad Request. From my understanding, a query like the above should still show 20 users, starting with the one in the position 19th. Am I wrong?

Damrod (changes requested):

transcendence-app/src/user/user.service.ts
vdedios: `i <= queryDto.limit` should fix that?

fizcamposancos: No, as I established a MAX() in the Dto, No limit above 20 will be considered. I'll fix it.

fizcamposancos: I've changed it to return always user. Without parameters it returns the maximum number of users established, and with queryparams, what has been selected. 
http://127.0.0.1:3000/api/v1/users?limit=20&offset=19&sort=true would return 20 users, from the 19th.

Damrod (approved):
Now THIS works! Thank you!! 
I'm sorry I was nagging.

PR #82, opened 2022-06-26
#56/Create-icon-UI-component

### LINKED ISSUE #56: ⚛️ Create icon UI component

gbudau:
This component is a wrapper around the SVG React Component which accepts a `title` prop (for accessibility) and all other props of an [SVG](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg) element.

We are exposing the following props: `type` (the type of the Icon), `size` (the height and width of the Icon), and `color`.

More info: https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs

closes #56


webapp/src/App.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import logo from './logo.svg';
 import './App.css';
+import { Icon, IconType } from './components/Icon/Icon';
```
vdedios, line 4: As discussed by chat, we should start organizing files and folders since the beggining so we don't face refactor issues in the future. We could have a folder named `pages` inside it, we could have the views related of each page. And inside each page folder, the needed `tsx`, `css` files and its own components/assets etc.

```txt
/src
- /pages
- App.tsx
- /landing
-- Landing.tsx
-- Landing.css
-- /components
--- (...)
...
```

Also, although for now is good, we shouldn't need any `App.css` and may have a `index.css `next to `index.tsx` so we have common styles there (for `:root`, `body` etc.)

webapp/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,51 @@
+import { ReactComponent as IconArrowBack } from './icons/arrow-back.svg';
```
vdedios, line 1: As before, I'd move this into a `common` folder since this components will be shared by multiple screens.

`webapp/src/common/components/Icon...`

webapp/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ReactComponent as IconArrowForward } from './icons/arrow-forward.svg';
+import { ReactComponent as IconChat } from './icons/chat.svg';
+import { ReactComponent as IconEdit } from './icons/edit.svg';
+import { ReactComponent as IconFile } from './icons/file.svg';
+import { ReactComponent as IconLogin } from './icons/login.svg';
+import { ReactComponent as IconLogout } from './icons/logout.svg';
+import { ReactComponent as IconPlay } from './icons/play.svg';
+import { ReactComponent as IconSearch } from './icons/search.svg';
+import { ReactComponent as IconUsers } from './icons/users.svg';
+
```
vdedios, line 11: What about the `size`? Each icon seem to have its own size. We should export them with a normalized size like in Figma

vdedios: Also, as said, I think we also have to think about a way to set `colors`

webapp/src/App.tsx
```diff
@@ -19,6 +20,7 @@ function App() {
         >
           Learn React
         </a>
+        <Icon type={IconType.LOGIN} style={{ fill: 'green' }} />
```
vdedios, line 23: I think we should define an interface for colors and not passing just `css` :(

webapp/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,51 @@
...
+  play: IconPlay,
+  search: IconSearch,
+  users: IconUsers,
+};
+
+type IconPropsType = SVGPropsType & {
+  type: IconType;
+};
+
+export function Icon({ type, ...restProps }: IconPropsType) {
```
vdedios, line 48: I've been playing around with this and I think importing icons as React components could be problematic. The ideal situation would be if we could expose only props that we want to be configured: (type, size, color etc.) But importing SVG like this forces you to expose all svg properties as props which we don't want... (see the spreading of `restProps` :sweat: )

Maybe there is a workaround but I haven't been able to do it, the ideal would be something like:

```tsx
export function Icon({ type, color, size }: IconProps) {
```

The only way I know to solve this is just to make our custom tsx and inject only desired proporties to SVG html element

vdedios (changes requested):
Warning!

It seems that not all `svg files` are the same, for example `chat` does not have the same style properties and html tags as `login`. If you try to import it in `App.tsxt` it does not get modified with `color:green` style for example

webapp/src/App.tsx
gbudau: Done :+1:

webapp/src/components/Icon/Icon.tsx
gbudau: Good catch. I've resized the icons to the same size (24px).

webapp/src/App.tsx
gbudau: Here you mean like an enum for colors? Or we could pass hex values?

webapp/src/components/Icon/Icon.tsx
gbudau: Good point. I've added the `color` and `size` types to the `IconPropsType` and removed the `SVGPropsType`.
Now we can pass only those props to the SVG component.
What do you think about this option?

gbudau:
@vdedios 
Thank you for the feedback :heart:
I think I've addressed all your suggestions.

webapp/src/components/Icon/Icon.tsx
gbudau: I've also added the `className` prop to style using CSS classes and `title` for accessibility

webapp/src/App.tsx
fizcamposancos: We can talk about the folder structure during next meeting, because I have some doubts. 

- A folfer for each single page?
- Inside this folder, at least 2 files: page.tsx and pase.css?
- So, a .css for each page?
- Variables in css like, for example, `--main-bg-color: black`, would they be in each file or just in the index.css?
- Where should we define our basic pallette and fonts?

gbudau: @fizcamposancos, sure, we could talk about it.

Currently planning to do something like this:

```
/src
- /pages
--  App.tsx
--  /Landing
---   Landing.tsx
---   Landing.css
---   index.tsx
---   /components
----   ...
--  /SomePage
---   SomePage.tsx
---   SomePage.css
---   index.tsx
---   ...
- /shared
--  /components
---   /Icon
----   Icon.tsx
----   index.tsx
----   /assets
-----    ...
---  /Button
----   Button.tsx
----   Button.css
----   index.tsx
- /assets
--  ...
...
```

The `pages` directory is for the views, the `shared` is for the shared components and files, and the `assets` is for other files.

What do we think about this structure?

webapp/src/components/Icon/Icon.tsx
gbudau: As discussed on Slack, we will expose only the `type`, `color`, and `size` for now.

webapp/src/App.tsx
vdedios: I really like that structure @gbudau! we can integrate a first approach like that and discuss in the next meeting about details. What @fizcamposancos says about `css` is also very important :+1:

vdedios (approved):
Thank you!

fizcamposancos (approved):
Nice job!!

webapp/src/App.tsx
Damrod: AFAIK we can use `@import "xxx.css";`in css, in case we want to reduce code duplication, and still have one .css file per component. Although I'm not sure how much do we need a particular css per component.

webapp/src/shared/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { ReactComponent as ArrowBack } from './assets/icons/arrow-back.svg';
+import { ReactComponent as ArrowForward } from './assets/icons/arrow-forward.svg';
+import { ReactComponent as Chat } from './assets/icons/chat.svg';
+import { ReactComponent as Edit } from './assets/icons/edit.svg';
+import { ReactComponent as File } from './assets/icons/file.svg';
+import { ReactComponent as Login } from './assets/icons/login.svg';
+import { ReactComponent as Logout } from './assets/icons/logout.svg';
+import { ReactComponent as Play } from './assets/icons/play.svg';
+import { ReactComponent as Search } from './assets/icons/search.svg';
+import { ReactComponent as Users } from './assets/icons/users.svg';
```
Damrod, line 11: I guess we cannot pack all of these imports into a barrel for consistency, because all of them have the same name and they are svgs?

Damrod (approved):
Looks good to me! Thank you

webapp/src/shared/components/Icon/Icon.tsx
gbudau: Done :+1:

webapp/src/shared/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,52 @@
...
+  LOGOUT: Logout,
+  PLAY: Play,
+  SEARCH: Search,
+  USERS: Users,
+};
+
+interface IconProps {
+  type: IconType;
+  size?: string | number | undefined;
+  color?: string | undefined;
```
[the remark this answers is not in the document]
gbudau: @vdedios These articles explain it better than I could:
https://react-typescript-cheatsheet.netlify.app/docs/basic/troubleshooting/types/#enum-types
https://www.executeprogram.com/blog/typescript-features-to-avoid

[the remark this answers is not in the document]
vdedios: Ok I see your point!! I'd say it is a good fit for the ColorType but what about the numeric values for the IconSize? I think we need an enum for that. We can make it constant to avoid passing any numeric value (https://fettblog.eu/tidy-typescript-avoid-enums/#prefer-union-types)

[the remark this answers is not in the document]
vdedios: I've seen you already solved the IconSizes passing it as string but I still believe using an enum for it makes more sense. Passing just a value does not give much semantic meaning (is it small, is it large?) and, also, imagine we want to update the sizes of our icon because we realize that the size of `24px` is too small. We'll have to update everywere we used that size to the new value which is painfull. With an enum you only have to update the value in one place...

I'd say using union string types makes sense for color **denominations** because they are not very prone to change, but for sizes, color codes etc. makes more sense to use variables.

webapp/src/shared/types/colors.ts
```diff
@@ -0,0 +1,9 @@
+export type Color =
+  | '--color-warning'
+  | '--color-submit'
+  | '--color-online'
+  | '--color-offline'
+  | '--color-light'
+  | '--color-background'
+  | '--color-dark'
+  | '--color-game';
```
vdedios, line 9: I think is strange to have color codes as css variables in tsx code we are sacrifiying legibility just to avoid importing this in a few places and saving a few lines of bundled `js` code. I really love the idea of union types but I think we should use them only when they add real value without substracting

```suggestion
export default enum Color = {
  WARNING =  '--color-warning',
   SUBMIT = '--color-submit',
  ONLINE =  '--color-online',
  OFFLINE = '--color-offline',
  LIGHT = '--color-light',
  BACKGROUND = '--color-background',
  DARK = '--color-dark',
  GAME = '--color-game'
}
```

or, if you really hate enums :smile: we could use a simple `js` object

```ts
export default Colors = {
  WARNING :  '--color-warning',
  SUBMIT : '--color-submit',
  ONLINE :  '--color-online',
  OFFLINE : '--color-offline',
  LIGHT : '--color-light',
  BACKGROUND : '--color-background',
  DARK : '--color-dark',
  GAME : '--color-game'
}
```

webapp/src/shared/components/Icon/Icon.tsx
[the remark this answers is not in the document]
gbudau: Thank you, good points. I've made the change :+1:

gbudau:
Thank you all :tada:

PR #84, opened 2022-06-27
#83/Create React folder structure

### LINKED ISSUE #83: ⚛️ Create React folder structure

gbudau:
Update index.css
Add color variables, fonts and fonts classes
Update libraries to the latest versions
Create a placeholder 'Landing' page to test colors and fonts.

closes #83


webapp/src/index.css
```diff
@@ -1,13 +1,118 @@
...
+    url('./assets/fonts/dm-sans-v11-latin-regular.woff') format('woff');
+}
+
+/* dm-sans-500 - latin */
+@font-face {
+  font-family: 'DM Sans';
+  font-style: normal;
+  font-weight: 500;
+  font-display: swap;
+  src: url('./assets/fonts/dm-sans-v11-latin-500.woff2') format('woff2'),
```
vdedios, line 52: What do you think about, instead of us storing the fonts, using Google Fonts :yum:  -> https://fonts.google.com/specimen/DM+Sans?query=dm+sans

vdedios: We can import this in `<head> -> <link>` https://github.com/42AGV/ft_transcendence/blob/c9e78797fd34c6d6fb162f590c6458b662cc4de4/webapp/public/index.html

webapp/src/index.tsx
```diff
@@ -1,14 +1,15 @@
+import { createRoot } from 'react-dom/client';
 import React from 'react';
-import ReactDOM from 'react-dom';
 import './index.css';
-import App from './App';
+import Landing from './pages/Landing';
```
fizcamposancos, line 4: If you are importing `Landing `from `./pages/Landing`, what's the point of having an index.tsx file in the Landing folder importing and exporting it?

gbudau: Good point. I've added that file to not have to type Landing twice.
```
import Landing from './pages/Landing/Landing';\
```


Maybe adding a barrel file to each component it's a bit too much. We could add one barrel file inside the components directory to import them from one single file in one line.

```
import { Landing, SomePage, SomeOtherPage } from './pages';
```

What do we think about this?

fizcamposancos: Ok, I got it. Yes, the idea of the barrel in the components folder sounds good. And one all the pages as well. I thought from that structure that there is going to be an index file for each page
/src
- /pages
--  /Landing
---   index.tsx 
---   Landing.tsx
---   Landing.css

Good 👍

webapp/src/index.css
```diff
@@ -1,13 +1,88 @@
...
+  font-weight: 700;
+}
+
+.text-style-4-regular {
+  font-size: 1rem;
+  font-family: 'DM Sans', sans-serif;
+  font-weight: 400;
+}
+
+.text-style-4-medium {
```
Damrod, line 66: I don't get it: we created the landing to test colors and fonts (and sizes), but we are only using one combination in the page?

webapp/public/index.html
```diff
@@ -24,7 +25,10 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>React App</title>
+    <title>ft_transcendence</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=DM+Sans:wght@400;500;700&display=swap" rel="stylesheet">
```
Damrod, line 31: we're linking against some external font we retrieve from google api?
Would it make sense to download those fonts and keep them in assets, not to rely in availability/licensing/not changeanbility of external services?

webapp/public/index.html
```diff
@@ -2,14 +2,15 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <!--Twitter, CC BY 4.0 <https://creativecommons.org/licenses/by/4.0>, via Wikimedia Commons-->
+    <link rel="icon" type="image/svg+xml" href="%PUBLIC_URL%/assets/images/favicon.svg">
```
Damrod, line 6: Do you also see this line of html emphasized with a transparentish yellowish overlay here in github? Is it something we have done? Is it the way my browser is rendering this page? Or github?

Damrod (approved):
thank you very much!

vdedios (approved):
LGTM!

webapp/public/index.html
vdedios: Do you trust yourself more than Google? 😄

vdedios: Apart from jokes, is very usual to store this kind of things in CDNs near to users (as well as images) and not directly bundled with the code, and is very common to import fonts from outside sources. I'd say importing this is safe because is an open font, and Google services are not very likely to be down

webapp/public/index.html
gbudau: Yes, you're right. I think this is how Github renders HTML comments.

webapp/public/index.html
gbudau: Both are good choices, and both have pros and cons. We could change it in the future if there are any issues :+1:

webapp/src/index.css
gbudau: The `Landing` page is a placeholder we could use as an example of how to structure the pages. I didn't want to add too much content since we will replace it with the actual content and styles when we start implementing the pages.

gbudau:
Thank you!

PR #88, opened 2022-07-03
#55 Create status UI component

### LINKED ISSUE #55: ⚛️ Create status UI component

gbudau:
closes #55


webapp/src/shared/components/Status/Status.css
```diff
@@ -0,0 +1,36 @@
...
+.status-text-online {
+  color: var(--color-online);
+}
+
+.status-text-playing {
+  background: var(--gradient-game);
+  background-clip: text;
+  -webkit-background-clip: text;
+  -webkit-text-fill-color: transparent;
+}
```
gbudau, line 36: https://github.com/42AGV/ft_transcendence/blob/92a129709eb7e559fa4e1ecb253278405c917801/webapp/src/shared/components/Text/Text.css#L1-L6

This class is the same as the one from the `Text` component.
Is it a good idea to move it to index.css?

```
.status-text-playing 
.text-color-game { 	  
  background: var(--gradient-game); 	  
  background-clip: text; 	  
  -webkit-background-clip: text; 	  
  -webkit-text-fill-color: transparent; 	  
}
```

fizcamposancos: Yes, I think so, as we have other text styles and colors, this way would be accesible to all the components, and no need to copy and paste it. It is not exaclty a color neither a style but I think we can move it there.

fizcamposancos (approved):
Very good implementation, an limitation of props with other props and the use of `never`.

webapp/src/index.css
```diff
@@ -85,31 +85,39 @@ body {
...
-.color-background {
+.text-color-background {
   color: var(--color-background);
 }
 
-.color-dark {
+.text-color-dark {
   color: var(--color-dark);
 }
+
```
Damrod, line 116: Why don't we like this in `webapp/src/shared/components/Text/Text.css` anymore?

Damrod (approved):
Looks good to me (but you will have conflicts with fportela's button)

webapp/src/index.css
gbudau: We move it here to reduce code duplication between the `Icon`, `Text`, and `Status` components because they share some parts.

webapp/src/shared/components/Status/Status.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: StatusProps) {
+  const variant = createVariant({ offline, online, playing });
+
+  return (
+    <div className="status">
+      <div className={`status-icon status-icon-${variant}`} />
+      <div className={`text-color-${variant} caption-regular`}>{variant}</div>
+    </div>
+  );
+}
```
vdedios, line 55: I think you may over engineer the solution because this could be easily solved in this case with string literal types. This would imply changing how we pass the prop to the component:

```tsx
      <Status variant='online' />
```

I know you've came up to this solution because in the issue we (I :sweat: ) described to have boolean props. But after seeing your implementation I think string literal types are a really better fit here. What do you think?

This being said, whatever you decide I'll agree with it!

```suggestion
type StatusType = 'online' | 'offline' | 'playing';

type StatusProps = {
    variant: StatusType
}

export default function Status({
    variant
}: StatusProps) {

  return (
    <div className="status">
      <div className={`status-icon status-icon-${variant}`} />
      <div className={`text-color-${variant} caption-regular`}>{variant}</div>
    </div>
  );
}
```

vdedios (approved):
LGTM :rocket: !!

webapp/src/shared/components/Status/Status.tsx
gbudau: You are right, and this is a lot easier to read :100:. Thank you.

gbudau:
Thank you :pray:

PR #92, opened 2022-07-08
Avatar UI component

### LINKED ISSUE #60: ⚛️ Create avatar UI component

vdedios:
closes #60

image

### DIFF
--- a/webapp/src/index.css
+++ b/webapp/src/index.css
@@ -59,37 +59,37 @@ body {
 }
 
 .heading-bold {
-  font-size: 1.5rem;
+  font-size: 2rem;
   font-family: var(--primary-font);
   font-weight: 700;
 }
 
 .subheading-bold {
-  font-size: 1.25rem;
+  font-size: 1.5rem;
   font-family: var(--primary-font);
   font-weight: 700;
 }
 
 .paragraph-regular {
-  font-size: 1rem;
+  font-size: 1.25rem;
   font-family: var(--primary-font);
   font-weight: 400;
 }
 
 .paragraph-medium {
-  font-size: 1rem;
+  font-size: 1.25rem;
   font-family: var(--primary-font);
   font-weight: 500;
 }
 
 .paragraph-bold {
-  font-size: 1rem;
+  font-size: 1.25rem;
   font-family: var(--primary-font);
   font-weight: 700;
 }
 
 .caption-regular {
-  font-size: 0.875rem;
+  font-size: 1rem;
   font-family: var(--primary-font);
   font-weight: 400;
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -10,6 +10,8 @@ import {
   TextColor,
   TextVariant,
   TextWeight,
+  SmallAvatar,
+  LargeAvatar
 } from '../../shared/components';
 import { Color } from '../../shared/types';
 import './Landing.css';
@@ -19,6 +21,8 @@ export default function Landing() {
   const btn_link = (): void => {
     window.location.href = 'http://google.com';
   };
+
+    const randomAvatar = 'https://i.pravatar.cc/1000'
   return (
     <section className="landing">
       <Status variant="playing" />
@@ -56,6 +60,12 @@ export default function Landing() {
         title="Users"
         urlNavigation="/users"
       />
+			<SmallAvatar url={randomAvatar} />
+			<LargeAvatar url={randomAvatar} />
+			<SmallAvatar url={randomAvatar} status='offline'/>
+			<LargeAvatar url={randomAvatar} edit status='online' caption='level 21'/>
+			<SmallAvatar url={randomAvatar} status='playing'/>
+			<LargeAvatar url={randomAvatar} edit status='playing' caption='level 21'/>
     </section>
   );
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -30,7 +30,7 @@ export default function Landing() {
         color={TextColor.LIGHT}
         weight={TextWeight.BOLD}
       >
-      Play online pong with your friends
+      Landing 🚀
       </Text>
       <Button variant={ButtonVariant.WARNING} onClick={btn_action}>
         Alert
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -25,16 +25,22 @@ export default function Landing() {
     const randomAvatar = 'https://i.pravatar.cc/1000'
   return (
     <section className="landing">
-      <Status variant="playing" />
-      <Button variant={ButtonVariant.SUBMIT} onClick={btn_action}>
-        ALERT
+      <Text
+        variant={TextVariant.TITLE}
+        color={TextColor.GAME}
+        weight={TextWeight.MEDIUM}
+      >
+        Landing 🚀
+      </Text>
+      <Button variant={ButtonVariant.WARNING} onClick={btn_action}>
+        Alert
       </Button>
       <Button
         variant={ButtonVariant.WARNING}
         iconVariant={IconVariant.LOGOUT}
         disabled
       >
-        LOGOUT BUTTON
+        Logout
       </Button>
       <Button
         variant={ButtonVariant.SUBMIT}
@@ -48,18 +54,12 @@ export default function Landing() {
         color={Color.ONLINE}
         size={IconSize.LARGE}
       />
-      <Text
-        variant={TextVariant.TITLE}
-        color={TextColor.GAME}
-        weight={TextWeight.MEDIUM}
-      >
-        Landing 🚀
-      </Text>
       <NavigationItem
         iconVariant={IconVariant.USERS}
         title="Users"
         urlNavigation="/users"
       />
+      <Status variant="playing" />
 			<SmallAvatar url={randomAvatar} />
 			<LargeAvatar url={randomAvatar} />
 			<SmallAvatar url={randomAvatar} status='offline'/>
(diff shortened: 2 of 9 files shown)


webapp/src/index.css
```diff
@@ -50,37 +50,37 @@ body {
 }
 
 .heading-bold {
-  font-size: 1.5rem;
+  font-size: 2rem;
```
vdedios, line 53: Adjust text sizes, in desktop where to small

webapp/src/pages/Landing/Landing.tsx
```diff
@@ -6,18 +6,28 @@ import {
...
   TextVariant,
   TextWeight,
+  SmallAvatar,
+  LargeAvatar
 } from '../../shared/components';
 import { Color } from '../../shared/types';
 import './Landing.css';
 
 export default function Landing() {
+	const randomAvatar = 'https://i.pravatar.cc/1000'
```
vdedios, line 16: just for testing purposes

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
...
+    </div>
+  );
+}
+
+export function LargeAvatar({
+  url,
+  status,
+  edit = false,
+  caption,
+}: LargeAvatarProps) {
```
vdedios, line 31: Better to have to differentiate types of avatars since each of them have different options. This way we don't have to have weird logic to parse them

webapp/src/shared/components/Icon/Icon.tsx
```diff
@@ -45,8 +45,8 @@ const Icons: { [key in IconVariant]: SVGComponent } = {
 };
 
 export enum IconSize {
-  SMALL = '24px',
-  LARGE = '48px',
+  SMALL = '1.25rem',
```
vdedios, line 48: Adjust icon size, in desktop where too big

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
+import './Avatar.css';
+
+import { Icon, IconVariant, IconSize } from '../index';
+import { Color } from '../../../shared/types';
+
+type AvatarStatus = 'online' | 'offline' | 'playing';
```
vdedios, line 6: This type is temporal until we have `status` component integrated. The idea is to be able to import `StatusType` here (and everywhere using statuses) so we have a centralized defintion of it

vdedios:
:warning:  Will solve conflicts when https://github.com/42AGV/ft_transcendence/pull/90 gets integrated

webapp/src/index.css
gbudau: The font styles match the mobile text styles from Figma with a base of 16px.
What do you think about using a media query for desktop and changing the base font-size?

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+type LargeAvatarProps = AvatarProps & {
+  edit?: boolean;
+  caption?: string;
+};
+
+export function SmallAvatar({ url, status }: AvatarProps) {
+  return (
+    <div className={`avatar-small__image-wrapper  avatar-status--${status}`}>
```
gbudau, line 20:
```suggestion
  const statusClassName = status ? `avatar-status--${status}` : '';

  return (
    <div className={`avatar-small__image-wrapper ${statusClassName}`}>
```

The status is `undefined` when not set and adds a class of `avatar-status--undefined`

What do you think about setting it to an empty string in this case?

gbudau (approved):
Very nice! Thank you.
LGTM :rocket:

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      )}
+      <div className={`avatar-large__image-wrapper  avatar-status--${status}`}>
+        <img src={url} alt={url} className="avatar-large__image" />
+      </div>
+      {caption && (
+        <caption className="avatar-large__caption caption-regular">
+          {caption}
+        </caption>
+      )}
```
fizcamposancos, line 50: Why `<caption>` and not other text tag? Or eve our text component?

fizcamposancos (approved):
Looks good!!!

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
+import './Avatar.css';
+
+import { Icon, IconVariant, IconSize } from '../index';
+import { Color } from '../../../shared/types';
```
Damrod, line 4: should this be `import { Color } from '../../shared/types';` ??

webapp/src/shared/components/Avatar/Avatar.tsx
gbudau: In Firefox and Chromium, there is this console warning:
```
Warning: validateDOMNesting(...): <caption> cannot appear as a child of <div>
```

Maybe we could use the [figure](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure) tag here with [figcaption](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption)?

webapp/src/shared/components/Icon/Icon.tsx
gbudau: In Firefox, there is this console warning, but it works ok from what I can tell:

```
Unexpected value 1.25rem parsing height attribute.                   react-dom.development.js:798
Unexpected value 1.25rem parsing width attribute.                    react-dom.development.js:798
```

webapp/src/shared/components/Avatar/Avatar.css
```diff
@@ -0,0 +1,50 @@
...
+
+.avatar-small__image-wrapper, .avatar-large__image-wrapper {
+    border-radius: 50%;
+    display: flex;
+    justify-content: center;
+    margin: 0.25rem;
+}
+
+.avatar-large__image {
+    width: 200px;
```
gbudau, line 29:
```suggestion
    width: 200px;
    height: 200px;
```
From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img)
> Use both [width](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width) and [height](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-height) to set the intrinsic size of the image, allowing it to take up space before it loads, to mitigate content layout shifts.

Same for the `.avatar-small__image` class.

webapp/src/shared/components/Avatar/Avatar.tsx
vdedios: Yes, good catch, I'll fix it :ok_hand:

webapp/src/shared/components/Avatar/Avatar.css
vdedios: good catch! reading the documentation we also could to get rid off `px` units: 

> [height](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-height)
> : The intrinsic height of the image, in pixels. Must be an integer without a unit.

I'll remove it

webapp/src/shared/components/Avatar/Avatar.tsx
vdedios: not, because `shared` folder is three levels under this path. But I see what you mean and we could have `../../types`

that being said I'll keep it as is because seeing this is imported from a shared folder gives more information

webapp/src/shared/components/Avatar/Avatar.css
vdedios: I say nothing about `px` it is only for the `img` tag properties itself

vdedios:
> The font styles match the mobile text styles from Figma with a base of 16px.
> What do you think about using a media query for desktop and changing the base font-size?
> 
> I think this is already like this on the main branch, so we might not need to change it.

About this, I'll rebase main. I adjusted a little bit the sizes because, despite having them defined in Figma, they were a little bit small once you saw them rendered in a webpage

webapp/src/shared/components/Icon/Icon.tsx
vdedios: I added this so we could have scaling depending on the device which makes sense because we don't want to see icons with the same size in all screens.

Maybe we can just inline css in the Icon component. It's not the best practise but, since we don't have any more styles here it seems to be a good fit. Something like:

```tsx
  <IconTag className={color} style={{width: size, height: size}} />;
```

webapp/src/index.css
vdedios: https://github.com/42AGV/ft_transcendence/issues/101

I think finally the best option is to update everything in this task, thanks for the observation!

webapp/src/shared/components/Avatar/Avatar.tsx
vdedios: done!

webapp/src/index.css
gbudau: Thank you, good catch! We could use classes instead of hardcoding the sizes in the Text component.

If the sizes are small only for the desktop, we could use media queries and match the values with the ones we have for desktop text in Figma.

Either setting a bigger base font
```
@media (min-width: 768px) {
  :root {
    font-size: 20px 
  }
}
```


Or setting the desktop size in each text class

```
@media (min-width: 768px) {
  .title-bold {
    font-size: 3rem;
  }
   
  .subtitle-bold {	  
    font-size: 2.5rem;	    
  }

  ...
}
```

And if the sizes are smaller for both mobile and desktop, we could update both default sizes in the index.css and Figma, and I think we also have to use media queries in this case.

vdedios: Nice!! If you want, you can add that comment to the issue so this info don't get lost

webapp/src/shared/components/Icon/Icon.tsx
gbudau: According to [this](https://stackoverflow.com/a/46337386) SO comment, Firefox does not support rem for SVGs.
Maybe we could wrap the Icon component in a div and apply the size with rem to the div instead because the SVGs can scale to the size of their container.

What do you think?

Issue #62, opened 2022-06-25
⚛️ Create row UI component


vdedios:

gbudau:
This component could have the following props:

- avatar: Avatar component (optional)
- title: Text component (optional)
- subtitle: Text component (optional)
- icon: Icon component (optional)

An example could be something like:

```tsx
<Row avatar={avatar} title='username' subtitle='description' icon={icon} />
```

Damrod:
some minor adjustment to row component

PR #93, opened 2022-07-09
#62/create row UI component

Damrod:
we need a row component

### DIFF
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -67,7 +67,7 @@ export default function Landing() {
 			<SmallAvatar url={randomAvatar} status='playing'/>
 			<LargeAvatar url={randomAvatar} edit status='playing' caption='level 21'/>
       <Row iconVariant={IconVariant.ARROW_FORWARD} avProps={{url:randomAvatar, status:'online'}} >
-        {'asd'} {'dffw3'}</Row>
+        {['John Doe','level 3']}</Row>
     </section>
   );
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -66,9 +66,8 @@ export default function Landing() {
 			<LargeAvatar url={randomAvatar} edit status='online' caption='level 21'/>
 			<SmallAvatar url={randomAvatar} status='playing'/>
 			<LargeAvatar url={randomAvatar} edit status='playing' caption='level 21'/>
-      <Row iconVariant={IconVariant.ARROW_FORWARD} avProps={{url:randomAvatar, status:'online'}} >
+      <Row iconVariant={IconVariant.ARROW_FORWARD} avProps={{url:randomAvatar, status:'online'}} onClick={btn_action}>
         {['John Doe','level 3']}</Row>
-      <h4>h4</h4>
     </section>
   );
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -12,7 +12,7 @@ import {
   SmallAvatar,
   LargeAvatar,
   NavigationBar,
-  Row
+  Row,
 } from '../../shared/components';
 import { Color } from '../../shared/types';
 import './Landing.css';
@@ -68,8 +68,13 @@ export default function Landing() {
         status="playing"
         caption="level 21"
       />
-      <Row iconVariant={IconVariant.ARROW_FORWARD} avProps={{url:randomAvatar, status:'online'}} onClick={btn_action}>
-        {['John Doe','level 3']}</Row>
+      <Row
+        iconVariant={IconVariant.ARROW_FORWARD}
+        avatarProps={{ url: randomAvatar, status: 'online' }}
+        onClick={btn_action}
+        title="John Doe"
+        subtitle="level 3"
+      />
     </section>
   );
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -66,7 +66,8 @@ export default function Landing() {
 			<LargeAvatar url={randomAvatar} edit status='online' caption='level 21'/>
 			<SmallAvatar url={randomAvatar} status='playing'/>
 			<LargeAvatar url={randomAvatar} edit status='playing' caption='level 21'/>
-      <Row iconVariant={IconVariant.USERS}>Row</Row>
+      <Row iconVariant={IconVariant.ARROW_FORWARD} avProps={{url:randomAvatar, status:'online'}} >
+        {'asd'} {'dffw3'}</Row>
     </section>
   );
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -18,8 +18,8 @@ import { Color } from '../../shared/types';
 import './Landing.css';
 
 export default function Landing() {
-  const btn_action = (): void => alert('This is an alert');
-  const btn_link = (): void => {
+  const buttonAction = (): void => alert('This is an alert');
+  const buttonLink = (): void => {
     window.location.href = 'http://google.com';
   };
 
@@ -34,7 +34,7 @@ export default function Landing() {
       >
         Landing 🚀
       </Text>
-      <Button variant={ButtonVariant.WARNING} onClick={btn_action}>
+      <Button variant={ButtonVariant.WARNING} onClick={buttonAction}>
         Alert
       </Button>
       <Button
@@ -47,7 +47,7 @@ export default function Landing() {
       <Button
         variant={ButtonVariant.SUBMIT}
         iconVariant={IconVariant.ARROW_FORWARD}
-        onClick={btn_link}
+        onClick={buttonLink}
       >
         To Google
       </Button>
@@ -70,8 +70,8 @@ export default function Landing() {
       />
       <Row
         iconVariant={IconVariant.ARROW_FORWARD}
-        avatarProps={{ url: randomAvatar, status: 'online' }}
-        onClick={btn_action}
+        avatar={<SmallAvatar url={randomAvatar} status="offline" />}
+        onClick={buttonAction}
         title="John Doe"
         subtitle="level 3"
       />
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -11,7 +11,7 @@ import {
   TextVariant,
   TextWeight,
   SmallAvatar,
-  LargeAvatar
+  LargeAvatar, Row
 } from '../../shared/components';
 import { Color } from '../../shared/types';
 import './Landing.css';
@@ -66,6 +66,7 @@ export default function Landing() {
 			<LargeAvatar url={randomAvatar} edit status='online' caption='level 21'/>
 			<SmallAvatar url={randomAvatar} status='playing'/>
 			<LargeAvatar url={randomAvatar} edit status='playing' caption='level 21'/>
+      <Row iconVariant={IconVariant.USERS}>Row</Row>
     </section>
   );
 }
--- a/webapp/src/pages/Landing/Landing.tsx
+++ b/webapp/src/pages/Landing/Landing.tsx
@@ -4,14 +4,15 @@ import {
   Icon,
   IconSize,
   IconVariant,
-  NavigationItem,
   Status,
   Text,
   TextColor,
   TextVariant,
   TextWeight,
   SmallAvatar,
-  LargeAvatar, Row
+  LargeAvatar,
+  NavigationBar,
+  Row
 } from '../../shared/components';
 import { Color } from '../../shared/types';
 import './Landing.css';
@@ -22,15 +23,16 @@ export default function Landing() {
     window.location.href = 'http://google.com';
   };
 
-    const randomAvatar = 'https://i.pravatar.cc/1000'
+  const randomAvatar = 'https://i.pravatar.cc/1000';
   return (
     <section className="landing">
+      <NavigationBar />
       <Text
         variant={TextVariant.TITLE}
         color={TextColor.LIGHT}
         weight={TextWeight.BOLD}
       >
-      Landing 🚀
+        Landing 🚀
       </Text>
       <Button variant={ButtonVariant.WARNING} onClick={btn_action}>
         Alert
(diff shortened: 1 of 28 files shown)


webapp/src/shared/components/Row/Row.css
```diff
@@ -0,0 +1,49 @@
...
+  border: 0;
+  position: relative;
+  align-items: center;
+  cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  min-width: 230px;
+}
+
+.row_wrapper {
```
gbudau, line 16: suggestion: Consider removing the `row-wrapper` div, and use the flexbox properties and padding, something like this:

```
.row {
  width: 100%;
  min-width: fit-content;
  padding: 16px 28px;
  border-radius: 8px;
  color: var(--color-light);
  background-color: var(--color-dark);
  border: 0;
  cursor: pointer;
  display: flex;
  align-items: center;
  column-gap: 20px;
}

.row_text_wrapper > * {
  margin: 0;
  text-align: left;
  white-space: no-wrap;
}

.row-icon {
  margin-left: auto;
}
```

The avatar and the text are aligned by flexbox and `margin-left: auto;` pushes the icon to the [right](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Aligning_Items_in_a_Flex_Container#using_auto_margins_for_main_axis_alignment).

If we add the `className` prop to the `Text` component, we could also use clases instead of the "exotic" `.row_text_wrapper > *` selector:
```
.row_text_title,
.row_text_subtitle {
  margin: 0;
  text-align: left;
  white-space: no-wrap;
}
```

gbudau (approved):
Thank you for this feature, and nicely done! I left some comments about what we could improve, but overall looks good to me!

webapp/src/shared/components/Text/Text.tsx
```diff
@@ -28,21 +29,29 @@ export enum TextWeight {
...
+  parent_class?: string;
   weight?: TextWeight;
   children: string;
 };
 
 export default function Text({
   variant,
-  color = TextColor.DARK,
+  color,
+  parent_class,
```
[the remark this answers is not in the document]
Damrod: I think you should merge first #96, and then we should revisit this, because we WILL have conflicts, and some of your changes are a implementing, in a different way, some of my changes.

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,54 @@
...
+import './Row.css';
+import { SmallAvatar, AvatarProps } from '../Avatar/Avatar';
+import { TextVariant, TextWeight } from '../Text/Text';
+import { default as Text } from '../Text/Text';
+
+type RowProps = {
+  iconVariant?: IconVariant;
+  onClick?: () => void;
+  avProps?: AvatarProps;
+  children?: string[];
```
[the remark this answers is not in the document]
Damrod: To be honest, I think passing these as children is one of the very few good ideas I had.

We could, in the future, pass an array of more than two strings, flex the height of a row, change the way we print this in lines 26 to 40 in Row.tsx to a loop, and fit a flexible number of lines. We could even pass not just lines of text, but some other html tag (<table> <rh> <rd>) , but then we would have to provide behaviour on how to actually handle the output in such cases.

But as you decided with the navigation bar to make such changes as needed, maybe it's the same case here.

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,54 @@
+import Icon, { IconVariant, IconSize } from '../Icon/Icon';
+import './Row.css';
+import { SmallAvatar, AvatarProps } from '../Avatar/Avatar';
+import { TextVariant, TextWeight } from '../Text/Text';
+import { default as Text } from '../Text/Text';
+
+type RowProps = {
+  iconVariant?: IconVariant;
+  onClick?: () => void;
+  avProps?: AvatarProps;
```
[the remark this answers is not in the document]
vdedios: agree with this, I think we should have descriptive names

webapp/src/shared/components/Text/Text.tsx
```diff
@@ -28,21 +29,29 @@ export enum TextWeight {
 type TextProps = {
   variant: TextVariant;
   color?: TextColor;
+  parent_class?: string;
```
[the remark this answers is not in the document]
vdedios: agree with this

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -4,7 +4,7 @@ import { Icon, IconVariant, IconSize } from '../index';
 import { StatusVariant } from '../Status/Status';
 import { Color } from '../../types';
 
-type AvatarProps = {
+export type AvatarProps = {
```
vdedios, line 7: Why do you need to export this? the url is just a string and the status is a `StatusVariant` which is a string type limited to a set of values. This is, you don't need to import this to pass the value of the stauts, you just need to pass one of allowed values of it. This is indeed one of the benefits of creating types like that, you don't have to import the type itself because it's just a string

webapp/src/shared/components/Row/Row.css
vdedios: Agree with this

webapp/src/shared/components/Row/Row.tsx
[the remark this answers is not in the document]
vdedios: I'd say (as explained in a comment above) passing an array here could be a little overkill for now. I see your point but we don't have such limitations for now, so I'd say it's better for now to just use props for now and evaluate in the future if we should upgrade this component when we meet such needs

webapp/src/shared/components/Row/Row.css
Damrod: I'm trying this, literally, and it doesn't work. I think we still need, at least, `row_text_wrapper`, in order to have `flex-direction: column;` maybe ?

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,54 @@
...
+  avProps,
+  children,
+}: RowProps) {
+  return (
+    <button className={`row paragraph-regular`} onClick={onClick}>
+      <div className={`row_wrapper`}>
+        {avProps && <SmallAvatar {...avProps} />}
+        {children && (
+          <div className={`row_text_wrapper`}>
+            {children[0] && (
```
vdedios: Same as before, `children[0]` is not very legibile :(  Apart from this, for me, an array should be a data structure to use when all components had the same relevance. Here the position `0` have a different relevance from `1`, they represent different things (title and subtitle) and we should take that into account.

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,54 @@
...
+                variant={TextVariant.PARAGRAPH}
+                weight={TextWeight.MEDIUM}
+                parent_class="row_text_wrapper"
+              >
+                {children[0]}
+              </Text>
+            )}
+            {children[1] && (
+              <Text
+                variant={TextVariant.CAPTION}
```
vdedios: I think subtitle is not a caption but a regular paragraph, please review text sizes/types from Figma -> https://www.figma.com/file/efdjuSsiJqDyZbOxs9UTI7/App?node-id=159%3A32

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <button className={`row paragraph-regular`} onClick={onClick}>
+      <div className={`row_wrapper`}>
+        {avProps && <SmallAvatar {...avProps} />}
+        {children && (
+          <div className={`row_text_wrapper`}>
+            {children[0] && (
+              <Text
+                variant={TextVariant.PARAGRAPH}
+                weight={TextWeight.MEDIUM}
```
vdedios: I think title is not a medium paragraph but a bold subheading, please review text sizes/types from Figma -> https://www.figma.com/file/efdjuSsiJqDyZbOxs9UTI7/App?node-id=159%3A32

webapp/src/shared/components/Row/Row.tsx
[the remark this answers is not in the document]
Damrod: done

webapp/src/shared/components/Row/Row.tsx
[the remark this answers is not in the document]
Damrod: Anyway, done

webapp/src/shared/components/Text/Text.tsx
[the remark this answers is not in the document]
Damrod: This one will wait for #96.

webapp/src/shared/components/Text/Text.tsx
```diff
@@ -28,21 +29,29 @@ export enum TextWeight {
...
   children,
 }: TextProps) {
   const { tag, size } = variant;
   const TextTag = tag as React.ElementType;
+  const classes = [
+    !color && !parent_class ? TextColor.DARK : color,
+    parent_class,
+  ]
+    .join(' ')
+    .trim();
```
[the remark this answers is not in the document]
Damrod: This one will wait for #96

webapp/src/shared/components/Row/Row.css
Damrod: Should be working now. Any more comments on this topic?

webapp/src/shared/components/Text/Text.tsx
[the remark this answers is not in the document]
vdedios: Also, I think we should avoid passing classes to components, we should respect as much as we can the interfaces exposed by the components (aka props) because they define how the component, from how we conceived it, can be modified. If we pass classes we kind of bypassing those restrictions and let te component be modified how we want.

In this case I'd say you could simply just create a `div`wrapper around the text component and apply all styling you need to it.

vdedios (changes requested):
I think this is a very good implementation which need still a few tweaks! :muscle:

webapp/src/shared/components/Row/Row.tsx
Damrod: Yes, I understand that, but it looks bad. Maybe we have to wait to have the text sizes fixed?

webapp/src/pages/Landing/Landing.tsx
```diff
@@ -67,6 +68,8 @@ export default function Landing() {
         status="playing"
         caption="level 21"
       />
+      <Row iconVariant={IconVariant.ARROW_FORWARD} avProps={{url:randomAvatar, status:'online'}} onClick={btn_action}>
```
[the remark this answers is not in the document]
Damrod: As mentioned in issue 63, I think passing components around has very weak typing, and I'm not sure it's a good idea.

webapp/src/pages/Landing/Landing.tsx
```diff
@@ -67,6 +68,8 @@ export default function Landing() {
...
+        {['John Doe','level 3']}</Row>
```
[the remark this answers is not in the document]
Damrod: This is outdated.

webapp/src/shared/components/Row/Row.tsx
Damrod: Also, plain no.
We agreed on figma because it looks good. Title is set to 'subheading', but the size of the font as displayed does not match what we have set for subheading. I have matched the text sizes, that looks to me like it's what we agreed on, not the text styles as we have them described in the top left corner. If there are inconsistencies, the displayed value should be prioritized.

Screenshot from 2022-07-17 23-10-17

Screenshot from 2022-07-17 23-10-04

Screenshot from 2022-07-17 23-09-46

webapp/src/shared/components/Row/Row.tsx
Damrod: Please, read above.

webapp/src/shared/components/Text/Text.tsx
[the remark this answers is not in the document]
Damrod: We have a mexican standoff here:

gbudau:
> praise: I like how you use the avatar props for this component.
I obviously don't dislike the idea.

webapp/src/shared/components/Row/Row.tsx
Damrod: There is room for negotiation in the weight part of this style, but it's tightly coupled to size, and as long as you don't agree that your opinion, as written above, on the text size is wrong, we are blocked.

webapp/src/shared/components/Text/Text.tsx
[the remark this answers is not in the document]
gbudau: There are pros and cons to both options :smiley: 

If we don't add a `className` prop to the components, we have to create wrappers to apply custom styling and positioning. It might make the code a bit less legible.

If we pass a `className` prop, we don't have to create wrappers, it might make the code more legible, but we can shoot ourselves in the foot.

webapp/src/shared/components/Row/Row.css
gbudau: Thank you! Looks good to me.
We could remove the `height: 82px;` from the `.row'` class to allow for flexible height when there is only a title or subtitle.

webapp/src/shared/components/Row/Row.tsx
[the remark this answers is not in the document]
gbudau: Thank you, it's not a bad idea. The suggestion was more from a naming point of view. I was concerned that after some time, it is very likely that we will forget what `children[0]`, `children[1]`, etc.. stands for.
We could revisit this later if the need arises.

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,58 @@
...
+  avatarStatus,
+  title,
+  subtitle,
+}: RowProps) {
+  return (
+    <button className={`row paragraph-regular`} onClick={onClick}>
+      {avatarUrl &&
+        ((avatarStatus && (
+          <SmallAvatar url={avatarUrl} status={avatarStatus} />
+        )) || <SmallAvatar url={avatarUrl} />)}
```
Damrod, line 30: I do this because otherwise we have this:
Screenshot from 2022-07-18 01-48-50

```
<figure class="avatar-small__image-wrapper avatar-status--undefined">...</figure>
```
Which doesn't seem very safe as a className to me. But maybe it's worth it to have simpler logic, and strange classNames that will probably (although not necesarilly) not exist.

Damrod: Ah look, we have this classname anyway...

fizcamposancos (approved):
Beside from all discussions, I think the solution is good. Well done!!

webapp/src/shared/components/Row/Row.tsx
vdedios: Thanks for updating this :grinning:  We have indeed a task to revisit font sizes -> https://github.com/42AGV/ft_transcendence/issues/101 because, as you say, there are some sizes which look bad. But I think naming would still be the same due to what it represents.

webapp/src/shared/components/Text/Text.tsx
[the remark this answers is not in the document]
vdedios: Let me explain this from another point of view:

If we enable classes to be passed around we'll be opening the component to be modified anywhere just to solve a particular case. For me, having a `div` around a specific component in a few places does not worsen legibility :stuck_out_tongue_closed_eyes: 

Another reason for not passing around classes is we would have css for the same component defined in several places (some styling defined near to it, other styling near to the component which needs it to behave different) You may have problems identifying sometimes where that CSS comes from.

In component libraries, such as Material UI, what they expose is [a utility](https://mui.com/system/styled/) to apply styles to components in order to create new ones, variations. But when you have a component, you want to respect as much as you can exposed interfaces. That component has been conceived to behave in a specific way.

If you need to start tweaking or hacking it, it may be a sign of a need to create a new variantion component.

webapp/src/shared/components/Row/Row.css
```diff
@@ -0,0 +1,26 @@
+.row {
+  display: flex;
+  width: 100%;
+  min-width: fit-content;
+  height: 82px;
+  padding: 16px 28px;
+  border-radius: 8px;
```
vdedios, line 7: I think we could use `rem` units better, this component may be bigger in desktop and, using this units, it would scale automatically depending on base font size.

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -0,0 +1,55 @@
...
+  avatarStatus?: StatusVariant;
+  title?: string;
+  subtitle?: string;
+};
+
+export default function Row({
+  iconVariant,
+  onClick,
+  avatarUrl,
+  avatarStatus,
```
vdedios, line 21: If we finally go with this option, I'd group related props at least. This is -> `AvatarProps`

vdedios (approved):
Just made a few improvement comments and a few more still to be discussed by the team. But I think this is very nice now! Good job :rocket:

webapp/src/pages/Landing/Landing.tsx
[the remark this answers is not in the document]
vdedios: (This is being discussed in Slack if we don't get a clear answer in a reasonable amount of time, I'd say we could merge it and fix it in the future so @Damrod don't get blocked)

PR #94, opened 2022-07-10
#58 Create navigation bar UI component

### LINKED ISSUE #58: ⚛️ Create navigation bar UI component

gbudau:
closes #58


webapp/src/shared/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,49 @@
...
+              iconVariant={IconVariant.CHAT}
+              title="Chat"
+              urlNavigation="/chat"
+            />
+          </li>
+        )}
+      </ul>
+    </nav>
+  );
+}
```
vdedios, line 49: Nice! but I think that, as soon as we start adding options, this won't scale. What about trying to group options, for example:

1. We could define what a navigation option should have:

```tsx
type NavOptionContent = {
  icon: IconVariant;
  title: string;
  url: string;
};
```

2. After that, lets add the real options

```tsx
export const users: NavOptionContent = { icon: IconVariant.USERS, title: 'Users', url: '/users' };
export const play: NavOptionContent = { icon: IconVariant.PLAY, title: 'Play', url: '/play' };
export const chat: NavOptionContent = { icon: IconVariant.CHAT, title: 'Chat', url: '/chat' };
```

3. Now we define what navigation options could we have and, also, the component props which can be an array so we can easily iterate through them:

```tsx
type NavOption = typeof users | typeof play | typeof chat;

type NavigationBarProps = {
  options: NavOption[];
};
```

4. Finally, we can glue all together in the component:

```tsx
export default function NavigationBar({ options }: NavigationBarProps) {
  return (
    <nav className="navigation-bar">
      <ul>
        {options.map((option: NavOption) => (
          <li>
            <NavigationItem
              iconVariant={option.icon}
              title={option.title}
              urlNavigation={option.url}
            />
          </li>
        ))}
      </ul>
    </nav>
  );
}
```


This way we avoid modifying the component whenever we have new options as well as keep it simple and full of logic.

The usage would change a bit from what you've done, but still very similar and comprehensive:

```tsx
import { NavigationBar, users, play, chat } from '../../shared/components';

// ...

<NavigationBar options={[users, play, chat]} />
```

vdedios (changes requested):
I think you should review how the component it's implemented because, when adding more and more options it won't scale. I've wrote down a suggestion, but feel free to do it your own way :wink:

webapp/src/shared/components/NavigationBar/NavigationBar.tsx
gbudau: I like having an array of items content and mapping it to print the list items.
I'm not a fan of exporting constants and passing the options as an array of those constants.

Could we make this component take no props for the items?

vdedios (approved):
LGTM!

webapp/src/shared/components/NavigationBar/NavigationBar.tsx
vdedios: This was disscussed by chat, we won't add for now the dynamic behaviour to the navigationBar until we have a scenario where we need it

gbudau:
Thank you! :rocket:

PR #96, opened 2022-07-11
#59 Create header UI component

### LINKED ISSUE #59: ⚛️ Create header UI component

gbudau:
mobile

<br>

desktop

close #59


webapp/src/shared/components/Header/Header.tsx
```diff
@@ -0,0 +1,50 @@
...
+          </Link>
+        ) : (
+          <Link to={urlNavigation}>
+            <img
+              src={navigation}
+              alt="avatar"
+              width="50px"
+              height="50px"
+              style={{ borderRadius: '50%' }}
+            />
```
gbudau, line 40: Temporary solution, until we merge #92

vdedios: I think you can use the real Avatar component, it has been integrated!

webapp/src/shared/components/Header/Header.tsx
```diff
@@ -0,0 +1,42 @@
...
+              variant={navigationFigure as IconVariant}
+              size={IconSize.SMALL}
+              color={Color.LIGHT}
+            />
+          ) : (
+            <SmallAvatar url={navigationFigure} />
+          )}
+        </Link>
+      </div>
+      <div className="header-text">{children}</div>
```
Damrod, line 36: I don't think this looks good.
We have a seriffed font for everything else, but this header has its own style. Somehow it is not acquiring the `--primary-font:'DM Sans', sans-serif;` style from the .root, but if you hardcode it as `font-family:'DM Sans;'` you can see the difference. Why don't you use some of our text components? Why did you create them for otherwise?

Damrod (changes requested):
As I mentioned, I think the children styling needs some rework.

webapp/src/shared/components/Header/Header.tsx
gbudau: Well spotted, thank you! I think it's fixed now.

vdedios: There was already a task to fix that -> https://github.com/42AGV/ft_transcendence/issues/101. I have checked the points fixed by @gbudau

webapp/src/shared/components/Text/Text.tsx
```diff
@@ -32,17 +33,13 @@ type TextProps = {
...
-}: TextProps) {
-  const { tag, size } = variant;
+export default function Text({ variant, color, weight, children }: TextProps) {
+  const { tag, className } = variant;
   const TextTag = tag as React.ElementType;
+  const textClassName = `${className} ${color ? color : ''}`;
 
   return (
-    <TextTag style={{ fontSize: size, fontWeight: weight }} className={color}>
+    <TextTag style={{ fontWeight: weight }} className={textClassName}>
```
Damrod, line 42: if weigth is undefined, what will this expand to?
I think it would be
```font-weight: undefined;```

File: webapp/src/shared/components/Header/Header.tsx [hunk 4]
```diff
@@ -0,0 +1,51 @@
...
+            />
+          ) : (
+            <SmallAvatar url={navigationFigure} />
+          )}
+        </Link>
+      </div>
+      <div className="header-text">
+        <Text
+          variant={TextVariant.HEADING}
+          color={TextColor.LIGHT}
```
Damrod, line 40: We have been doing this so far, but I'm not sure it is a good idea.
We are hardcoding the colors. What if we want to change all the colors? Something like a night mode. Wouldn't it be better to just pass down the css values for vars like 'foreground color' and 'background color'?

webapp/src/shared/components/Text/Text.tsx
gbudau: I think if the weight is undefined, React does not set an inline `font-weight` for this element.

webapp/src/shared/components/Header/Header.tsx
gbudau: I don't understand. Could you please clarify what needs to be done? Thank you.

Damrod: So, we are hardcoding in several places these values, inside different components: `TextColor.LIGHT` and `TextColor.DARK`. I have a problem with LIGHT and DARK, because we are impliying that dark has to be dark (as close-ish to #0000000 ) and light has to be light (as close-ish to #ffffff), and I don't think we should bind ourselves to such a contract, because we may want to swap (all darks become lights, and all lights become darks).

I thought there was a problem in the dependency tree for styles, in the sense that we would set dark for some text, and then change root variables, and not have them change in our components, but I have noticed this is not happening. As things stand we have: 

```
.text-color-light {
color: var(--color-light)
} --- takes the value of `--color-light : yy` from ---->
.root {
color-dark:xx;
color-light:yy
}
```
```
.text-color-dark {
color: var(--color-dark)
} --- takes the value of `--color-dark : xx` from ---->
.root {
color-dark:xx;
color-light:yy
}
```

which is fine. If we want to change all darks, we change color-dark:xx in .root, and it works.

I think I still have a problem with dark and light. I'd rather name them like something along the lines of `light` -> `foreground`, `background` -> `background0`, and `dark` -> `background1` . But this is a concern I am just raising now, and it's outside of the scope of this PR.

Damrod (approved):
LGTM!

webapp/src/shared/components/Header/Header.tsx
gbudau: You are right. We could improve the naming of the color variables. What do we think about adding primary, secondary, tertiary, neutrals (white, black, and greys), and accent colors?
Then we could have context, like --color-primary-background, --color-secondary--background, --color-accent-submit, --color-accent-warning, --color-neutral-light, or something like that.
 Let's make a new task for this.

fizcamposancos (commented):
Good job, but I've found a problem with the text of the central position of the header. If the text in the middle section of the header is long, or if you reducing the size ot the window or zooming in the site, the text pushes the status outside ot the window and a horizontal scroll bar appears to reach that content, but the content outside the main window is not well visible.
This is due to `.header-text > * {
  margin: 0;
  white-space: nowrap;
}`
We can leave it like that, which if fine if we all know it and we all agree, or fix it,  ajust it to multiple lines, reduce text size, wrap it with `text-overflow: ellipsis;` or something similar, limit text length, ...

gbudau:
@fizcamposancos, well spotted, and thank you. I think it's mostly fixed now. It still "breaks" after a minimum resolution (below ~250-300px), but I think this is normal.

fizcamposancos (approved):
Now It's good enough for me. Good job

webapp/src/shared/components/Header/Header.tsx
```diff
@@ -0,0 +1,51 @@
+import { Link } from 'react-router-dom';
+import { Color } from '../../types';
+import { SmallAvatar } from '../Avatar/Avatar';
+import Icon, { IconSize, IconVariant } from '../Icon/Icon';
+import Status, { StatusVariant } from '../Status/Status';
+import Text, { TextColor, TextVariant, TextWeight } from '../Text/Text';
+import './Header.css';
+
+type HeaderProps = {
+  navigationFigure: string;
```
vdedios, line 10: Nice!! but I'd explicitly add `IconVariant` so it is more clear what options can we pass around

```suggestion
  navigationFigure: IconVariant | string;
```

webapp/src/pages/Users/Users.css
```diff
@@ -2,13 +2,18 @@
   margin: 0;
   padding: 0;
   min-height: 100vh;
+  background-color: var(--color-background);
+  color: var(--color-light);
+}
+
+.users-wrapper {
+  width: 80%;
```
vdedios, line 10: This is just a comment about what is about to come when we implement screens, but since it Figma we are using grids to organize the content, maybe it worth to evaluate if we want to use them to organize content or just flex. For me, any of them is good.

vdedios (approved):
Very nice! This time, not many comments to make :p

webapp/src/pages/Users/Users.css
gbudau: Good point! I vote for a grid layout to match the design in Figma. Flexbox is also a good choice.
Also, these pages are temporary. We will probably start from scratch when we start implementing the pages.

gbudau:
Thank you all :bow:

Issue #63, opened 2022-06-25
⚛️ Create rows list UI component


vdedios:

gbudau:
This component could have the following props:

- rows: array of Rows components

An example could be something like:

```tsx
<RowsList rows={rows} />
```

Damrod:
> This component could have the following props:
> 
> * rows: array of Rows components
> 
> An example could be something like:
> 
> ```tsx
> <RowsList rows={rows} />
> ```

We want this to take an array of RowProps that we internally manage and convert into rows or literally something like JSX.Element[]
```
const Rows : JSX.Element[] = [<Row
    iconVariant={IconVariant.ARROW_FORWARD}
    avatarProps={{ url: randomAvatar, status: 'online' }}
    onClick={btn_action}
    title="John Doe"
    subtitle="level 3"
  />, <Row
    iconVariant={IconVariant.ARROW_FORWARD}
    avatarProps={{ url: randomAvatar, status: 'offline' }}
    onClick={btn_action}
    title="Jane Doe"
    subtitle="level 99"
  /> ];
```
?


The first one fits more the description of this ticket, but the second one doesn't provide stront type checking.

gbudau:
I'd say it is probably better to pass an array of `RowProps`, as you are suggesting.

PR #102, opened 2022-07-17
#63/rows list UI component

Damrod:
We need to see several rows in one go


webapp/src/shared/components/Row/Row.tsx
```diff
@@ -2,29 +2,26 @@ import Icon, { IconVariant, IconSize } from '../Icon/Icon';
...
-import { StatusVariant } from '../Status/Status';
+import { AvatarProps, SmallAvatar } from '../Avatar/Avatar';
 
-type RowProps = {
+export type RowProps = {
   iconVariant?: IconVariant;
   onClick?: () => void;
-  avatarUrl?: string;
-  avatarStatus?: StatusVariant;
+  avatarProps?: AvatarProps;
```
gbudau, line 10: praise: I like this more, thank you!

webapp/src/shared/components/Row/Row.css
```diff
@@ -2,7 +2,6 @@
   display: flex;
   width: 100%;
   min-width: fit-content;
-  height: 4.5rem;
   padding: 0.9rem 1.56rem;
   border-radius: 0.44rem;
```
gbudau, line 6: praise: Looks good. Thank you for removing the height.

suggestion: I think these rem values are using 20px as a base, and for the mobile version, the base is 16px
```
  padding: 1rem 1.75rem;
  border-radius: 0.5rem;
  column-gap: 1.25rem;
```
The `column-gap` in Figma was 22px. I think, for consistency, we could use a multiple of 4. We could change it to 20px

webapp/src/shared/components/RowsList/RowsList.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export type RowsListProps = {
+  rows?: RowProps[];
+};
+
+export default function RowsList({ rows }: RowsListProps) {
+  return (
+    <div className="rows-list-wrapper">
+      {rows && rows.map((rowProps) => <Row {...rowProps} />)}
+    </div>
```
gbudau, line 12: suggestion: Here we could use the unordered list and list items tags to improve the semantics.
Also, when mapping over an array, [React docs](https://reactjs.org/docs/lists-and-keys.html#keys) recommends adding a `key` property to the items.

Something like this:
```typescript
   <ul className="rows-list">
      {rows &&
        rows.map((rowProps) => (
          <li className="rows-list-item" key={rowProps.title}>
            <Row {...rowProps} />
          </li>
        ))}
    </ul>
```

```css
.rows-list {
  width: 100%;
  list-style: none;
  margin: 0;
  padding: 0;
}

.rows-list-item:not(:last-child) {
  margin-bottom: 1.75rem;
}
```

gbudau (approved):
Thank you, LGTM!

webapp/src/shared/components/RowsList/RowsList.tsx
Damrod: better now?

Damrod: Maybe `title` isn't stable enough as a key, since we may have duplicated titles, I think.

gbudau: You are right :100:
Also, the `title` is optional, and it could be `undefined`.
Any ideas on what we could use as a `key`?

Damrod: maybe we should add a id prop to Row, and use the uuid of the user?

gbudau: Sure, using the user UUID seems like a good option to me. Maybe we could name the prop `key` because `id` has a special purpose in HTML.
Or we could also use `id` without passing it as a prop to the `button` tag inside the Row component.

vdedios: good catch! I like the idea of adding an addtional `id` prop to the row component :+1:

vdedios (approved):
very nice, LGTM! :rocket:

webapp/src/shared/components/RowsList/RowsList.tsx
fizcamposancos: A key in the row component is a good idea and using user UUID seems good as well

Damrod: Case in point for the :not(:lastchild:) pseudo selector here https://dev.to/puputtallulah/stop-using-ids-and-classes-in-every-html-element-3k2c

Damrod:
I made [these](https://github.com/42AGV/ft_transcendence/pull/102/commits/194567c15de53295548fca5bd815b0fb437bcc1c) changes, but I'm not particularly happy about them. I don't like the key in the RowProps. I'll rework it tomorrow, to follow a pattern similar than the one we have in the backend with userDto, and userEntity.

webapp/src/shared/components/RowsList/RowsList.tsx
Damrod: Do we like where the key is [now](https://github.com/42AGV/ft_transcendence/pull/102/commits/37b6d66f8b2f3c95a1ddb736af586e0657429517)?

gbudau: I like the idea of adding the `key` only to the RowListProps :+1:
We could improve the naming of the types. What do you think about `RowItem` and `rowProps`?

RowEntity -> RowItem
rowDto -> rowProps

Another option is to create a type that extends the RowProps type:

```
type RowItem = RowProps & {
  key: string;
};
```

Then we could use destructuring while mapping the rows:
```
        rows.map((rowListProps) => {
          const { key, ...rowProps } = rowListProps;
          return (
            <li className="rows-list-item" key={key}>
              <Row {...rowProps} />
            </li>
          );
        }
```

Damrod: Now?

Damrod:
This is refusing to even run macos workflows.
I think we may need to remove them, as gbudau has in his latest draft pr.

webapp/src/shared/components/RowsList/RowsList.tsx
gbudau: @Damrod, thank you. LGTM :rocket:

vdedios (approved):

fizcamposancos (approved):
Nice job. LGTM!

Issue #101, opened 2022-07-13
⚛️ Revisit font sizes


vdedios:
- [x] Fix broken fonts in `Text` component

And, also:
 
- [x] Unify font sizes from Figma to application.
- [x] Add the `line-height` property to the text classes to match the sizes from Figma
- [x] Get rid off harcoded text font sizes in `Text` component

gbudau:
Thank you, good catch! We could use classes instead of hardcoding the sizes in the Text component.

If the sizes are small only for the desktop, we could use media queries and match the values with the ones we have for desktop text in Figma.

Either setting a bigger base font
```
@media (min-width: 768px) {
  :root {
    font-size: 20px 
  }
}
```


Or setting the desktop size in each text class

```
@media (min-width: 768px) {
  .title-bold {
    font-size: 3rem;
  }
   
  .subtitle-bold {	  
    font-size: 2.5rem;	    
  }

  ...
}
```

And if the sizes are smaller for both mobile and desktop, we could update both default sizes in the index.css and Figma, and I think we also have to use media queries in this case.

vdedios:
Part of this task have been solved in https://github.com/42AGV/ft_transcendence/pull/96#event-7005111752

gbudau:
- [ ] Add the `line-height` property from Figma to the text classes from index.css

PR #104, opened 2022-07-20
#61/create form input component

### LINKED ISSUE #61: ⚛️ Create form input UI component

fizcamposancos:
closes #61 
It may close that issue but I would open at least two new tasks, something like:
- Create search users form component
- Create edit user form component

This is just an uncontrolled input field to be used in the other forms.
Please, let me know if you agree with this


webapp/src/shared/components/Input/Input.css
```diff
@@ -0,0 +1,39 @@
...
+
+.input-container-light {
+  background-color: var(--color-light);
+}
+
+.input-form {
+  border: 0;
+  width: 100%;
+}
+
```
Damrod, line 25: maybe we would like to control the color of the placeholder text? Right now it's the default color...

webapp/src/shared/components/Input/Input.tsx
```diff
@@ -0,0 +1,41 @@
...
+  label?: string;
+  iconVariant?: IconVariant;
+  input?: string;
+};
+
+export default function Input({
+  color,
+  label,
+  iconVariant,
+  input,
```
Damrod, line 21: in the component input, everything is input. Maybe `placeholder` would be a name that more accurately would describe what is it that it's actually doing?

webapp/src/shared/components/Input/Input.tsx
```diff
@@ -0,0 +1,41 @@
+import { Color } from '../../types';
+import Icon, { IconSize, IconVariant } from '../Icon/Icon';
+import './Input.css';
+
+export enum InputColor {
+  LIGHT = 'light',
```
Damrod, line 6: I'm not sure why do we need this class. We already have a enum for colors. I see that you want to put logic on top in order to transform the color from the prop into different colors, but I don't understand why would we need a special separate symbol. Also, it's ambiguous. Input color? Color of the text that is the input or of the background of the input text? Or of the component named 'input' that contains both? It's confusing.

Damrod (approved):
Besides the comment, LGTM

webapp/src/shared/components/Input/Input.tsx
gbudau: What do we think about naming it to `InputVariant`?

webapp/src/shared/components/Input/Input.tsx
```diff
@@ -0,0 +1,41 @@
...
+            variant={iconVariant}
+            size={IconSize.LARGE}
+            color={color === InputColor.DARK ? Color.LIGHT : Color.DARK}
+          ></Icon>
+        )}
+        <input
+          className={`input-form input-form-${color} paragraph-regular `}
+          placeholder={input}
+        />
+      </div>
```
gbudau, line 38: praise: I like how you have done this part. Brilliant idea to place the icon next to the input inside a container and match the colors.

webapp/src/shared/components/Input/Input.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export default function Input({
+  color,
+  label,
+  iconVariant,
+  input,
+}: InputProps) {
+  return (
+    <div className="form-container subheading-bold">
+      {label && <label>{label}</label>}
```
gbudau, line 25:
```suggestion
    <label className="form-container subheading-bold">
      {label}
```

suggestion: Here we could use the `label` tag directly as a wrapper to associate it with the `input`.

webapp/src/shared/components/Input/Input.css
```diff
@@ -0,0 +1,39 @@
+.form-container {
+  width: 312px;
+}
```
gbudau, line 3:
```suggestion
.input {
  width: 100%;
  min-width: 200px;
}
```

suggestion: We could rename this class to `input` to match the naming we have currently (CSS classes match the name of the component).

question: What do we think about adding a width of 100% if we want to allow flexible size? Then we could control the width of this component using a wrapper tag.

webapp/src/shared/components/Input/Input.css
```diff
@@ -0,0 +1,39 @@
...
+
+.input-container {
+  margin-top: 0.3rem;
+  display: flex;
```
gbudau, line 7:
```suggestion
  display: flex;
  align-items: center;
```
suggestion: This centers the icon and the input horizontally inside the input container.

webapp/src/shared/components/Input/Input.tsx
```diff
@@ -0,0 +1,41 @@
+import { Color } from '../../types';
+import Icon, { IconSize, IconVariant } from '../Icon/Icon';
+import './Input.css';
+
+export enum InputColor {
+  LIGHT = 'light',
+  DARK = 'dark',
+}
+
+type InputProps = {
```
gbudau, line 10: question: What do we think about adding a `value` and `onChange` optional props to allow creating controlled inputs?

Then we could simply pass them to the input tag:
```
<input ... onChange={onChange} value={value} />
```

gbudau (approved):
Thank you, very nice implementation. Looks good to me.

webapp/src/shared/components/Input/Input.css
fizcamposancos: If we want to have different colors for the placeholder in the dark or light scenario, we can add something like that, sure.

webapp/src/shared/components/Input/Input.tsx
fizcamposancos: I stuck to the issue names but you're right, it makes more sense.

webapp/src/shared/components/Input/Input.tsx
fizcamposancos: Yes, again I tried to follow the issue scheme, but you are right, we can rename it to something better and try to take advantage of the enum color we already have. I will have a look

webapp/src/shared/components/Input/Input.tsx
fizcamposancos: Yes, it looks good. I'm changing it. And regarding the spacing, add it to #101 and I will also try to have a look

webapp/src/shared/components/Input/Input.css
fizcamposancos: I change the name and take note for future CSS classes.
Regarding the width, I like the idea, it adds more flexibility for future designs or uses

webapp/src/shared/components/Input/Input.tsx
vdedios: Agree with @gbudau and another option for use could be [useRef](https://es.reactjs.org/docs/hooks-reference.html#useref) instead of names. Names could end up being problematic if we have inputs with the same name, but ref identifies them unequivocally

webapp/src/shared/components/Input/Input.css
```diff
@@ -0,0 +1,50 @@
+.input {
```
vdedios, line 1: just in case it gets override for some reason, what about defining default color as white? (right now labels are white because they inherit the color from root)

```suggestion
.input {
  color: var(--color-light); 
```

vdedios (approved):
Nice and agree with the idea of having this base component and extending to two specific types. Just a comment about it, I'd place those inside this folder and create an `index.ts` to export all inputs together. Very nice!!

webapp/src/shared/components/Input/Input.css
fizcamposancos: Yes, good point. Done

PR #108, opened 2022-07-29
Create users page

Damrod:
We need a page with users


fizcamposancos:
No se ha vosotros pero a mi no se me ve bien la página. No me sale ningún usuario y tampoco me carga el avatar

Damrod:
You need to login first. http://localhost:3000/api/v1/auth/login , and then the 42intra authentication

webapp/src/pages/Users/Users.css
```diff
@@ -1,21 +1,92 @@
 .users {
   margin: 0;
-  padding: 0;
+  padding: var(--padding-top-bottom) var(--padding-sides);
   min-height: 100vh;
   background-color: var(--color-background);
   color: var(--color-light);
+  grid-template-rows: auto auto 1fr auto;
```
gbudau, line 7:
```suggestion
  height: 100vh;
  background-color: var(--color-background);
  color: var(--color-light);
  display: grid;
  grid-template-columns: 1fr;
  grid-template-rows: auto auto 1fr auto;
  gap: var(--row-gap) var(--column-gap);
```

suggestion: We could set the `height` to `100vh` and use a grid display here. This way, the `users-rows` will occupy the remaining space automatically, and we could remove the `margin` and `height` of the items (`users-search`, `users-rows` and `users-navigation`)

webapp/src/pages/Users/Users.tsx
```diff
@@ -1,74 +1,106 @@
...
   RowsList,
+  RowsListProps,
+  SmallAvatar,
 } from '../../shared/components';
-import './Users.css';
+import React from 'react';
+import { USER_URL, USERS_EP_URL, USERS_URL } from '../../shared/urls';
+import { Link } from 'react-router-dom';
+
+class User {
```
gbudau, line 20: suggestion: We could convert this to an interface because we don't use the constructor anywhere. In the `transcendence-app`, we define it as a class because the swagger module requires classes to generate documentation automatically.

webapp/src/pages/Users/Users.tsx
```diff
@@ -1,73 +1,99 @@
...
+      });
+    });
+    return rows;
   };
-  const randomAvatar = 'https://i.pravatar.cc/1000';
-  const randomAvatar2 = 'https://i.pravatar.cc/2000';
-  const randomAvatar3 = 'https://i.pravatar.cc/3000';
+
+  useEffect(() => {
+    const fetchUsersList = async () => {
```
gbudau, line 70: Good job for connecting it to the backend :rocket:

gbudau (approved):
Thank you, looks very good! :clap:

webapp/src/pages/Users/Users.css
Damrod: what do you think about setting the number of columns in mobile to 4? Not that we need them here, but it works equally, and like that we can use this style sheet as boilerplate for most of the rest of the pages

gbudau: Sure, as you see fit.

webapp/src/pages/Users/Users.tsx
Damrod: done

webapp/src/pages/Users/Users.tsx
Damrod: thank you

webapp/src/pages/Users/Users.css
Damrod: No, it's paying for something we are not using, and the change will be trivial, when we do need it. I prefer your choice.

PR #111, opened 2022-07-30
Add a Docker volume to store transcendence-app data

gbudau:
Update transcendence-app Dockerfile
Add `TRANSCENDENCE_APP_DATA` environment variable
Set by default to the /var/lib/transcendence-app directory

Please add `TRANSCENDENCE_APP_DATA=/var/lib/transcendence-app` to the transcendence-app `.env` file

Subtask of #74

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,11 @@ services:
     depends_on:
       - "db"
       - "memcached"
+    build:
+      args:
+        - TRANSCENDENCE_APP_DATA=${TRANSCENDENCE_APP_DATA:-/var/lib/transcendence-app}
+    volumes:
+      - transcendence-app-data:${TRANSCENDENCE_APP_DATA:-/var/lib/transcendence-app}
 
   db:
     image: postgres:14.4-alpine3.16
@@ -30,6 +35,8 @@ services:
 
   memcached:
     image: memcached:1.6.15-alpine3.16
+    restart: always
 
 volumes:
   transcendence-db:
+  transcendence-app-data:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -18,6 +18,11 @@ services:
     depends_on:
       - "db"
       - "memcached"
+    build:
+      args:
+        - TRANSCENDENCE_APP_DATA=${TRANSCENDENCE_APP_DATA:-/var/lib/transcendence-app}
```
gbudau, line 23: This syntax is to set a default value in case this variable is undefined

vdedios (approved):
👌👌👌

fizcamposancos (approved):
LGTM! :+1:

docker-compose.yml
Damrod: Why do we even have to set it if there is a default and that is the default we all need to set it to?

Damrod (approved):
LGTM!

docker-compose.yml
gbudau: The arg from here is an option if someone prefers to store the data in another directory inside the Docker container. Since it's not a confidential setting, we set it by default to `/var/lib/transcendence-app`. It's similar to how PostgreSQL has a `PGDATA` variable which defines where it stores the data.

With Docker, the application works without setting this variable in the `.env` file of transcendence-app because there is the `ENV` directive in the Dockerfile, and Docker exposes it to the container. But if someone wants to run the application without Docker, this variable is required. I hope it makes sense.

gbudau:
Thank you! :tada:

PR #112, opened 2022-07-30
#74/Add local file interceptor and api file decorator

gbudau:
Subtask of #74


transcendence-app/src/user/decorators/api-file.decorator.ts
```diff
@@ -0,0 +1,17 @@
...
+        type: 'object',
+        properties: {
+          [fileName]: {
+            type: 'string',
+            format: 'binary',
+          },
+        },
+      },
+    })(target, propertyKey, descriptor);
+  };
```
gbudau, line 17: This decorator is for documenting an endpoint that accepts multipart/form-data using OpenAPI specification
More info here: https://stackoverflow.com/questions/65082784/how-do-i-annotate-an-endpoint-in-nestjs-for-openapi-that-takes-multipart-form-da

transcendence-app/src/shared/local-file/local-file.interceptor.ts
```diff
@@ -0,0 +1,51 @@
...
+interface LocalFileInterceptorOptions {
+  fieldName: string;
+  path?: string;
+  fileFilter?: MulterOptions['fileFilter'];
+  limits?: MulterOptions['limits'];
+}
+
+function LocalFileInterceptor(
+  options: LocalFileInterceptorOptions,
+): Type<NestInterceptor> {
```
gbudau, line 18: TLDR: It's extending the `FileInterceptor` to use a default directory for storing the files and to optionally add a subpath, filter the file by MIME type or limit by file size.

More info about how this works -> https://wanago.io/2021/11/08/api-nestjs-uploading-files-to-server/ under 'Extending the FileInterceptor'

transcendence-app/src/user/user.controller.ts
```diff
@@ -26,6 +27,27 @@ import {
...
+        ),
+        false,
+      );
+    }
+    callback(null, true);
+  },
+  limits: {
+    fileSize: Math.pow(1024, 2), // 1MB
+  },
+});
```
gbudau, line 50: This is an instance of the `LocalFileInterceptor`, which allows only JPEG files under 1MB.

We save this into a variable to allow importing it from the unit tests

transcendence-app/src/user/decorators/api-file.decorator.ts
```diff
@@ -0,0 +1,17 @@
+import { ApiBody } from '@nestjs/swagger';
+
+export const ApiFile =
+  (fileName = 'file'): MethodDecorator =>
+  (target, propertyKey, descriptor) => {
+    ApiBody({
+      schema: {
+        type: 'object',
+        properties: {
+          [fileName]: {
```
Damrod, line 10: I like this way of mapping objects uniquely to data. I wonder if we have more use cases for this.

transcendence-app/src/user/user.controller.ts
Damrod: Why only jpg? Are we against any other image format? Anyway, we can discuss this later.

Damrod (approved):
LGTM!
mixin is an interesting concept I had not bumped into...

transcendence-app/src/user/user.controller.ts
gbudau: We could allow more image formats :+1:

gbudau: Update: I've added a constant with an array of allowed formats. For now it contains `image/jpeg`, `image/png` and `image/webp`. This way, we can easily add/remove other image formats in the future.

gbudau:
Thank you :+1:

Issue #52, opened 2022-06-11
Create logs and tests and read from config in db repository


vdedios:
Some comments:

- Investigate if makes sense to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message with nestjs logger

PR #113, opened 2022-07-30
#52/Read from config in PostgreSQL connection

gbudau:
Add environment variables for PostgreSQL connection
Add docker-compose.test.yml for e2e tests
Update package.json add pretest and post:test scripts to start and stop a database container for testing
Move auth and user e2e tests to transcendence-app/test directory
Remove app.e2e-spec.ts

**To test this PR, please copy your transcendence-app `.env` file to `.env.development`**

Subtask of #52
Subtask of #74


transcendence-app/src/main.ts
```diff
@@ -19,6 +19,7 @@ async function bootstrap() {
       transformOptions: {
         enableImplicitConversion: true,
       },
+      forbidUnknownValues: true,
```
gbudau, line 22: The [docs](https://github.com/typestack/class-validator#passing-options) recommend this option.

transcendence-app/package.json
```diff
@@ -19,7 +19,9 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json"
+    "pretest:e2e": "docker compose -f docker-compose.test.yml up -d",
+    "test:e2e": "jest --config ./test/jest-e2e.json; ret=$?; npm run post:test:e2e; exit $ret",
+    "post:test:e2e": "docker compose -f docker-compose.test.yml down -v"
```
gbudau, line 24: Workaround to execute the ["post" script](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts) even if the tests fail.

transcendence-app/.env.test
```diff
@@ -0,0 +1,19 @@
...
+MEMCACHED_PORT=11212
+TRANSCENDENCE_APP_DATA=/tmp
+POSTGRES_HOST=localhost
+POSTGRES_PORT=5433
+POSTGRES_DB=ft_transcendence
+POSTGRES_USER=postgres
+POSTGRES_PASSWORD=postgres
+
+# UUIDs format, generate with uuidgen
+SESSION_SECRET=80fbbf31-7083-419a-8fe1-23f1e10aae80
```
Damrod, line 18: I don't get why is .env.test different from .env.sample, in the sense that UUIDs can be hardcoded.

transcendence-app/src/config/env.validation.ts
```diff
@@ -49,11 +66,10 @@ export class EnvironmentVariables {
 }
 
 export function validate(config: Record<string, unknown>) {
-  const validatedConfig = plainToClass(EnvironmentVariables, config, {
-    enableImplicitConversion: true,
-  });
+  const validatedConfig = plainToClass(EnvironmentVariables, config);
   const errors = validateSync(validatedConfig, {
     skipMissingProperties: false,
+    forbidUnknownValues: true,
```
Damrod, line 72: just a question:
Doesn't the fact that all the attributes in EnvironmentVariables are marked as `!` required, overlap with this forbidUnknownValues?

transcendence-app/test/auth/auth.e2e-spec.ts
```diff
@@ -1,11 +1,11 @@
...
+import { AuthController } from '../../src/auth/auth.controller';
 import { HttpStatus, INestApplication } from '@nestjs/common';
-import { AuthModule } from './auth.module';
+import { AuthModule } from '../../src/auth/auth.module';
 import * as request from 'supertest';
 import { ConfigModule, ConfigService } from '@nestjs/config';
-import { validate } from '../config/env.validation';
-import { OAuth42Config } from './oauth42-config.interface';
+import { validate } from '../../src/config/env.validation';
+import { OAuth42Config } from '../../src/auth/oauth42-config.interface';
```
Damrod, line 8: I think we will need soon (or already?) some way to shorten the import paths. Something like [this](https://getfrontend.tips/shorten-import-paths-in-typescript/)

Damrod (approved):
LGTM! Thank you!

transcendence-app/test/auth/auth.e2e-spec.ts
gbudau: This is interesting, thank you. We could make a new task for this because it requires changing almost all the files if we want to use this syntax.

transcendence-app/src/config/env.validation.ts
gbudau: The `!` is the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-). This operator tells TypeScript that these values can't be null or undefined because they are validated here at runtime.

transcendence-app/.env.test
gbudau: Do you mean we could remove the `.env.sample` and have only `.env.test` since we already use it for the e2e tests?

transcendence-app/src/shared/db/postgres/postgresConnection.provider.ts
```diff
@@ -1,20 +1,41 @@
...
+
+    this.pool.on('error', (err: Error) => this.logger.error(err.message));
+  }
+
+  async onModuleInit() {
+    try {
+      await this.pool.query('SELECT 1');
+    } catch (err) {
+      this.logger.error(err);
+    }
```
vdedios, line 34: I assume this is to ensure a good first connection?

transcendence-app/test/user/user.e2e-spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        transformOptions: {
+          enableImplicitConversion: true,
+        },
+        forbidUnknownValues: true,
+      }),
+    );
+    await app.init();
+  });
+
+  it('Creates an user [POST /]', async () => {
```
vdedios, line 54:
```suggestion
  it('Creates a user [POST /]', async () => {
```

transcendence-app/src/shared/db/postgres/postgresConnection.provider.ts
```diff
@@ -1,20 +1,41 @@
...
+  }
+
+  async onModuleDestroy() {
+    await this.pool.end();
```
vdedios, line 38: Maybe, if we have this, we could get rid off end() method. I think is not being called anywhere

vdedios (approved):
Very nice!

transcendence-app/src/shared/db/postgres/postgresConnection.provider.ts
gbudau: Yes, if there is an error with the database connection, we log it here.

transcendence-app/src/shared/db/postgres/postgresConnection.provider.ts
gbudau: Done, thanks

transcendence-app/.env.test
Damrod: discussed on slack

gbudau: We decided to remove the step to set the UUIDs for the `SESSION_SECRET` and the `MEMCACHED_SECRET` from the [installing](https://github.com/42AGV/ft_transcendence#installing) documentation.
For development and testing, these values can be hardcoded for simplicity.

We will have to add this step to the documentation for [deployment](https://github.com/42AGV/ft_transcendence#-deployment-) to a production environment.

gbudau:
Thank you! :tada:

PR #115, opened 2022-07-30
#114/Fix transcendence-app logout endpoint

### LINKED ISSUE #114: Fix transcendence-app logout endpoint
The `req.logout()` method got an update in [passport 0.6.0](https://medium.com/passportjs/fixing-session-fixation-b2b68619c51d) and now is async and has a callback parameter

gbudau:
The req.logout() method got an update. Now is async and takes a callback as an argument
Update Oauth42Guard -> Throw ServiceUnavailableException and log the error
Add a guard to the logout endpoint to return Forbidden if the user is not authenticated
Update @types/passport to the latest version

close #114

### DIFF
--- a/transcendence-app/src/auth/oauth42.guard.ts
+++ b/transcendence-app/src/auth/oauth42.guard.ts
@@ -17,11 +17,9 @@ export class OAuth42Guard extends AuthGuard('oauth42') {
     return result;
   }
 
-  handleRequest(err: Error, user: any) {
+  handleRequest<User>(err: Error, user: User) {
     if (err || !user) {
-      if (err) {
-        this.logger.error(err.message);
-      }
+      err && this.logger.error(err.message);
       throw new ServiceUnavailableException();
     }
     return user;
--- a/transcendence-app/src/auth/oauth42.guard.ts
+++ b/transcendence-app/src/auth/oauth42.guard.ts
@@ -19,8 +19,10 @@ export class OAuth42Guard extends AuthGuard('oauth42') {
 
   handleRequest(err: Error, user: any) {
     if (err || !user) {
-      this.logger.error(err.message);
-      throw new ServiceUnavailableException(err.message);
+      if (err) {
+        this.logger.error(err.message);
+      }
+      throw new ServiceUnavailableException();
     }
     return user;
   }
--- a/transcendence-app/src/auth/oauth42.guard.ts
+++ b/transcendence-app/src/auth/oauth42.guard.ts
@@ -1,12 +1,15 @@
 import {
-  BadGatewayException,
   ExecutionContext,
   Injectable,
+  Logger,
+  ServiceUnavailableException,
 } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 
 @Injectable()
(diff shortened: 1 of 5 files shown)


fizcamposancos (approved):
Good catch, now it looks better

Damrod:
```
ft_transcendence-transcendence-app-1  | src/auth/auth.controller.ts:39:16 - error TS2554: Expected 0 arguments, but got 1.
ft_transcendence-transcendence-app-1  | 
ft_transcendence-transcendence-app-1  |  39     req.logout(function (err: Error) {
ft_transcendence-transcendence-app-1  |                    ~~~~~~~~~~~~~~~~~~~~~~~
ft_transcendence-transcendence-app-1  |  40       if (err) {
ft_transcendence-transcendence-app-1  |     ~~~~~~~~~~~~~~~~
ft_transcendence-transcendence-app-1  | ... 
ft_transcendence-transcendence-app-1  |  42       }
ft_transcendence-transcendence-app-1  |     ~~~~~~~
ft_transcendence-transcendence-app-1  |  43     });
ft_transcendence-transcendence-app-1  |     ~~~~~
ft_transcendence-transcendence-app-1  | 
```

Please, take a look at this compiler error I'm seeing.

Damrod (approved):
LGTM!

gbudau:
@Damrod, sometimes Docker doesn't build the images even if you pass the `--build` flag :pensive:. The solution is what you did :+1:, first run `docker compose down`, then `docker compose up --build -d`. And if there is an issue with the volumes, `docker compose down -v` to remove the old volumes.

gbudau:
After some investigation, it looks like this happens because we have an anonymous volume for the node_modules directory of the transcendence-app. We use this volume to isolate that directory from the host. This allows the container to work even if the host doesn't have a node_module directory.

We can pass the `--renew-anon-volumes` flag (or `-V`). Otherwise, Docker will reuse the old volume with the node_modules, which contains the old packages.

```
docker compose up --build -d -V
```

transcendence-app/src/auth/oauth42.guard.ts
```diff
@@ -1,22 +1,28 @@
...
+
   async canActivate(context: ExecutionContext) {
     const result = (await super.canActivate(context)) as boolean;
     const request = context.switchToHttp().getRequest();
     await super.logIn(request);
     return result;
   }
 
-  handleRequest(err: any, user: any) {
+  handleRequest(err: Error, user: any) {
```
vdedios, line 20: User can be any too?

vdedios (approved):
Thanks for taking the effort in splitting reviews, this is much more comprehensive to review

transcendence-app/src/auth/oauth42.guard.ts
gbudau: I think it could be `any`, yes.

The definition of the interface:
```
handleRequest<TUser = any>(err: any, user: any, info: any, context: ExecutionContext, status?: any): TUser;
```

I've changed it to `handleRequest<User>(err: Error, user: User)`, and I don't know if it's doing anything for the typing system. We could leave it as documentation.

gbudau:
Thank you all :tada:

PR #119, opened 2022-08-04
#52/Connect transcendence-app with db

### LINKED ISSUE #21: Create a db method adapter to get and filter users from DB

### LINKED ISSUE #52: Create logs and tests and read from config in db repository

gbudau:
close #21
close #52


transcendence-app/src/shared/db/models.ts
```diff
@@ -10,11 +10,11 @@ export interface BaseEntity {
 
 export interface Query {
   text: string;
-  values: Array<string | null>;
+  values: any[];
```
gbudau, line 13: I think `any[]` is a more suited type for this property because it could represent an array with strings, numbers, Dates, nulls, etc...

For example, if there is an entity like this:

```
{
  id: string
  name: string;
  age: number;
  created_at: Date;
  someOtherProp: null;
  ...
}
```

transcendence-app/src/shared/db/postgres/utils.ts
```diff
@@ -21,21 +25,18 @@ export const makeQuery = async <T>(
 ): Promise<T[] | null> => {
   try {
     const res = await pool.query(query);
-    return res?.rows?.length ? (res.rows as T[]) : null;
-  } catch (e) {
-    // TODO temp solution until we implement logging
-    throw new Error(stringifyPostgresError(e));
+    return res.rows;
+  } catch (err) {
+    if (isCriticalDatabaseError(err)) {
```
gbudau, line 30: Here we log database errors only for critical cases:
FATAL -> Reports an error that caused the current session to abort.
PANIC -> Reports an error that caused all sessions to abort.

We don't log errors for these cases:
ERROR -> Reports an error that caused the current command to abort.

https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS

transcendence-app/src/user/user.domain.ts
```diff
@@ -1,9 +1,21 @@
...
+  id: string;
+  createdAt: Date;
+
   constructor(
-    public username: string,
-    public email: string,
-    public avatar: string | null,
-    public id: string,
-    public created_at: number,
-  ) {}
```
gbudau, line 8: The [Swagger CLI](https://docs.nestjs.com/openapi/cli-plugin) plugin doesn't work with this class syntax. I've changed to a more verbose format to make it work and automatically detect the scheme of the class.

Alternatively, we could remove the constructor because I don't think we use it anywhere.

And we could define this class like this:
```typescript
export class User {	 
  username!: string;
  email!: string;
  avatar!: string | null;
  id!: string;
  createdAt!: Date;
}
```

db/setup/schema.sql
```diff
@@ -2,10 +2,10 @@ CREATE DATABASE ft_transcendence;
...
-    USERNAME    VARCHAR(20)                     NOT NULL UNIQUE,
-    EMAIL       VARCHAR(50)                     NOT NULL UNIQUE,
-    AVATAR      VARCHAR(150),
-    CREATED_AT  TIMESTAMPTZ                     NOT NULL DEFAULT now()
+CREATE TABLE IF NOT EXISTS Users(
+  "id" UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+  "username" VARCHAR(20) NOT NULL UNIQUE,
+  "email" VARCHAR(50) NOT NULL UNIQUE,
+  "avatar" VARCHAR(150),
+  "createdAt" TIMESTAMPTZ NOT NULL DEFAULT NOW()
```
gbudau, line 10: Here we quote the column names to have the same naming style between SQL and TypeScript.

The downside is that when we do raw queries to the database, we have to also quote the names, for example:

```sql
SELECT "id", "email" FROM Users WHERE "username" = 'example';
```

db/setup/schema.sql
```diff
@@ -2,10 +2,10 @@ CREATE DATABASE ft_transcendence;
...
 \c ft_transcendence;
 
-CREATE TABLE IF NOT EXISTS users(
-    ID          UUID DEFAULT gen_random_uuid()  PRIMARY KEY,
-    USERNAME    VARCHAR(20)                     NOT NULL UNIQUE,
-    EMAIL       VARCHAR(50)                     NOT NULL UNIQUE,
-    AVATAR      VARCHAR(150),
-    CREATED_AT  TIMESTAMPTZ                     NOT NULL DEFAULT now()
+CREATE TABLE IF NOT EXISTS Users(
+  "id" UUID DEFAULT gen_random_uuid() PRIMARY KEY,
```
Damrod, line 6: I was concerned about getting collisions of uuids when we don't check that we already have one like the one we just created, but [this](https://stackoverflow.com/questions/18708496/how-to-be-sure-uuid-is-not-duplicated) convinced me that this is not likely at all.

transcendence-app/nest-cli.json
```diff
@@ -8,7 +8,8 @@
         "name": "@nestjs/swagger",
         "options": {
           "classValidatorShim": true,
-          "introspectComments": true
+          "introspectComments": true,
+          "dtoFileNameSuffix": [".dto.ts", ".entity.ts", ".domain.ts"]
```
Damrod, line 12: (I thought I wrote this on Friday, but somehow my review has dissapeared) Anyway, in this case, I was commenting that [dtoFileNameSuffix](https://docs.nestjs.com/openapi/cli-plugin) seems to be assigning, by default, the suffixes that you are explicitly adding here.

> The Swagger plugin will automatically:
> Please, note that your filenames must have one of the following suffixes: ['.dto.ts', '.entity.ts'] (e.g., create-user.dto.ts) in order to be analysed by the plugin.
>
> f you are using a different suffix, you can adjust the plugin's behavior by specifying the dtoFileNameSuffix option (see below).

Are you sure we need to be explicit here?

transcendence-app/src/user/user.controller.spec.ts
```diff
@@ -47,24 +56,46 @@ describe('User Controller unit tests', () => {
...
+        const user = await controller.addUser(testUserDto);
+        expect(user.id).toEqual(testUserId);
+        expect(user.username).toEqual(testUserDto.username);
+      });
+    });
 
-  it('returns an existing user', async () => {
-    const user = await controller.getUserById(testUserId);
-    expect(user.id).toEqual(testUserId);
+    describe('when user with username exists', () => {
```
Damrod, line 69: I think the choice of not allowing duplicated usernames is slightly questionable, but I don't care enough to stop the merge.

transcendence-app/test/auth/auth.e2e-spec.ts
```diff
@@ -25,19 +30,44 @@ describe('Auth Controller', () => {
...
+        }),
+      }),
+    );
+    app.use(passport.initialize());
+    app.use(passport.session());
     app.init();
   });
 
-  it('should redirect to 42 oauth page', async () => {
+  it('Login [GET /login]', async () => {
```
Damrod, line 57: I'm not sure that the description string you have chosen is more descriptive that the string it used to have.

transcendence-app/test/auth/auth.e2e-spec.ts
```diff
@@ -25,19 +30,44 @@ describe('Auth Controller', () => {
...
     const response = await request(app.getHttpServer())
       .get('/auth/login')
       .expect(HttpStatus.FOUND);
     const authUrl = config.get('FORTYTWO_APP_AUTHORIZATION_URL');
     expect(response.headers['location']).toMatch(new RegExp(`^${authUrl}?`));
   });
 
+  it('Logout [DELETE /logout]', async () => {
```
Damrod, line 65: Why do we get forbidden when we attempt to log out here?

Damrod (approved):
Besides the minor comments, it looks really good thank you!

transcendence-app/nest-cli.json
gbudau: Yes. Once we set some custom values, it overwrites the default ones, and we have to assign them again.

transcendence-app/test/auth/auth.e2e-spec.ts
gbudau: Fixed, thanks

transcendence-app/test/auth/auth.e2e-spec.ts
gbudau: We added the authenticated guard to this route in #115.
Now unauthenticated users accessing this route will get a Forbidden(403) response.

This test makes an HTTP request without valid cookies (as an unauthenticated user), hence the Forbidden response.

gbudau:
Thank you :pray:

transcendence-app/src/user/user.controller.spec.ts
gbudau: We could discuss these options at the next meeting. I'll add a point in Notion.

PR #122, opened 2022-08-09
#75/Fetch and persist user image from 42 intranet on registration

### LINKED ISSUE #74: Implement an user storage to save user images

### LINKED ISSUE #75: Fetch and persist user image from 42 intranet on registration

gbudau:
```mermaid
erDiagram
    User {
        string username
        string email
        Date createdAt
    }
    LocalFile {
        string filename
        string path
        string mimetype
        number size
    }
    User ||--o| Avatar : has
    Avatar ||--|| LocalFile : is
  
 ```

**This PR adds a new table to the database and updates the `Users` table. Because we don't have migration scripts, for now, to test this, please remove Docker volumes and restart the containers.**

```
docker compose down -v
docker compose up -d --build
```

close #74
close #75

### CODE REVIEW

**Comment by Damrod** [comment]
So, there is no GET `http://localhost:3000/api/v1/users/${user-uuid}/avatar`  endpoint yet. I've been connecting `docker exec -it ft_transcendence-db-1 bash` to the running container after logging in, expecting to see the avatar file at /var/lib/ft_transcendence but no luck. Is this not implemented yet? How can I check if the avatar file is there?

File: transcendence-app/src/auth/oauth42.strategy.ts [hunk 1]
```diff
@@ -25,13 +28,30 @@ export class OAuth42Strategy extends PassportStrategy(Strategy, 'oauth42') {
...
+    );
     const dbUser = await this.userService.retrieveUserWithUserName(
       userDto.username,
     );
     if (dbUser) {
       return dbUser;
     }
-    return this.userService.addUser(userDto);
+
+    // Download the avatar and save the avatar metadata and the user in the database
```
Damrod, line 41: Couldn't we make one or two private, properly named, methods in this OAuth42Strategy class for these avatar file operations, instead of just leaving some comments?

transcendence-app/src/shared/db/postgres/postgresConnection.provider.ts
```diff
@@ -45,4 +46,27 @@ export class PostgresPool implements OnModuleInit, OnModuleDestroy {
...
+      } catch (err) {
+        await client.query('ROLLBACK');
+        throw err;
+      } finally {
+        client.release();
+      }
+      return data;
+    } catch (err) {
+      if (isCriticalDatabaseError(err)) {
+        this.logger.error(err.message);
```
Damrod, line 67: I don't get why do we only log 'FATAL' or 'PANIC' severity level errors.

transcendence-app/src/shared/local-file/infrastructure/db/local-file.entity.ts
```diff
@@ -0,0 +1,19 @@
+import { BaseEntity } from '../../../db/models';
+
+export enum LocalFileKeys {
+  ID = '"id"',
```
Damrod, line 4: I understand why do we have this, but we have to admit that it's very funky.

transcendence-app/src/shared/local-file/infrastructure/db/postgres/local-file.postgres.repository.ts
```diff
@@ -0,0 +1,39 @@
+import { Injectable } from '@nestjs/common';
+
+import { BasePostgresRepository } from '../../../../../shared/db/postgres/postgres.repository';
+import { table } from '../../../../../shared/db/models';
+import { LocalFileEntity, LocalFileKeys } from '../local-file.entity';
+import { PostgresPool } from '../../../../../shared/db/postgres/postgresConnection.provider';
```
Damrod, line 6: '../../../../../' <- this is still horrible, but it's also out of the scope of this PR.

transcendence-app/src/shared/local-file/local-file.service.ts
```diff
@@ -0,0 +1,90 @@
...
+    const file = await this.repository.deleteById(id);
+
+    if (!file) {
+      return null;
+    }
+    this.deleteFileData(file.path);
+    return file;
+  }
+
+  updateFileById(
```
Damrod, line 84: we don't use this method.

transcendence-app/src/user/api42.service.ts
```diff
@@ -25,6 +25,15 @@ export class Api42Service {
     return data;
   }
 
+  downloadUserAvatar(accessToken: string, url: string) {
+    return lastValueFrom(
```
Damrod, line 29: > Converts an observable to a promise by subscribing to the observable, waiting for it to complete, and resolving the returned promise with the last value from the observed stream.
If the observable stream completes before any values were emitted, the returned promise will reject with EmptyError or will resolve with the default value if a default was specified.
If the observable stream emits an error, the returned promise will reject with that error.
WARNING: Only use this with observables you know will complete. If the source observable does not complete, you will end up with a promise that is hung up, and potentially all of the state of an async function hanging out in memory. To avoid this situation, look into adding something like timeout, take, takeWhile, or takeUntil amongst others.

Specially:

> **WARNING: Only use this with observables you know will complete**. If the source observable does not complete, you will end up with a promise that is hung up, and potentially all of the state of an async function hanging out in memory

Are we sure lastValueFrom is the best option for this?

>  To avoid this situation, look into adding something like timeout, take, takeWhile, or takeUntil amongst others.

Do we know a request to the 42Api will complete?

db/setup/schema.sql
```diff
@@ -2,10 +2,18 @@ CREATE DATABASE ft_transcendence;
 
 \c ft_transcendence;
 
+CREATE TABLE IF NOT EXISTS LocalFile(
+  "id" UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+  "filename" VARCHAR(100) NOT NULL UNIQUE,
+  "path" VARCHAR(255) NOT NULL UNIQUE,
+  "mimetype" VARCHAR(255) NOT NULL,
+  "size" BIGINT NOT NULL
```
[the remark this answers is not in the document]
Damrod: Maybe it would be useful to have the size before we upload it to our service, as validation, so that users can't upload monstrously large images?

Damrod (approved):
Besides the comments, it's great, thank you!

gbudau:
> What is the responsibility of each of the containers that run when we do `docker-compose up --build`?

Currently, there are five services:

`nginx` -> reverse proxy, proxies requests to `/` route to the `webapp` service and to `/api` route to `transcendence-app` service.
The only entrypoint to our infrastructure. Accessible via port 3000 in development and 80 in production.

`webapp` -> Single page application. In development it uses webpack dev server for hot reloading, and in production it uses an nginx server to serve the data.

`transcendence-app` -> API server.
Stores files at `/var/lib/transcendence-app-data` which is a docker volume `transcendence-app-data` for file persistance.
Uses the `memcached` service to cache the session data.
Uses the `db` service to persist data.

`memcached` -> memory caching.

`db` -> The database of our application. Stores data in a docker volume `transcendence-db` for persistance.

```mermaid
graph TB;
  user["fa:fa-user user"]
  nginx{"fa:fa-server nginx"};
  webapp("fa:fa-server webapp");
  transcendence-app("fa:fa-server transcendence-app");
  transcendence-app-data[("fa:fa-folder transcendence-app-data")];
  memcached("fa:fa-server memcached");
  db("fa:fa-database db");
  transcendence-db[("fa:fa-folder transcendence-db")];

  user --> |HTTP request| nginx
  nginx --> |HTTP response| user

  nginx-->|/| webapp;
  nginx-->| /api | transcendence-app;

  transcendence-app-->db;
  transcendence-app-->| session store |memcached
  transcendence-app-->| /var/lib/transcendence-app |transcendence-app-data;

  db-->|/var/lib/postgresql/data|transcendence-db;
```

transcendence-app/src/auth/oauth42.strategy.ts
gbudau: Done, :+1:

transcendence-app/src/user/api42.service.ts
gbudau: Good catch!
I've added a timeout to the HttpModule. It should be good now.
Should we make it an environment variable or keep it as a constant?

transcendence-app/src/shared/local-file/local-file.service.ts
gbudau: Fixed, thanks.

transcendence-app/src/shared/db/postgres/postgresConnection.provider.ts
gbudau: https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS

The 'pg' driver throws errors in many cases. For example, when the query has a syntax error or failed because of some constrain.

Let's take an example of a user who wants to update his email address and select an email address already in use. On the Users database table, the email column has a UNIQUE constrain. In this case, the 'pg' driver throws an error.

If we add logging to every case, it could get noisy and hard to notice critical errors. Also, the database container already logs these cases.

Here we log unrecoverable errors, like if the database connection is down or it can't save data because there is no space left. Anyway, this is my opinion. If we decide that it's an improvement to log every error, we could remove this condition. Or maybe log all the errors in the development environment and leave this condition for the production environment.

db/setup/schema.sql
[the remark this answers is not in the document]
gbudau: I've added a `createdAt` column for this table :+1: 
We use the `size` to set the `Content-Length` response header.

Adding a maximum size is also a good idea. We check this when a user uploads his avatar using the `AvatarFileInterceptor` from `UserController`.

gbudau:
Thank you for the feedback and suggestions :rocket:

Issue #123, opened 2022-08-09
Create controllers to upload and fetch the user avatar


gbudau:
PUT [/api/v1/users/avatar](http://localhost:3000/api#/users/UserController_uploadAvatar) -> update the avatar of the currently authenticated user

GET [/api/v1/users/avatar](http://localhost:3000/api#/users/UserController_getCurrentUserAvatar) -> get the avatar of the currently authenticated user

GET [/api/v1/users/{uuid}/avatar](http://localhost:3000/api#/users/UserController_getAvatar) -> get the avatar of the user with `uuid`

![image](https://user-images.githubusercontent.com/54190400/181372282-a54ae9f3-6ffb-4034-bf3d-ab15bafa0c51.png)

PR #125, opened 2022-08-10
#123/Create controllers to fetch the user avatar

gbudau:
GET [/api/v1/users/avatar](http://localhost:3000/api#/users/UserController_getCurrentUserAvatar) -> get the avatar of the currently authenticated user
GET [/api/v1/users/{uuid}/avatar](http://localhost:3000/api#/users/UserController_getAvatar) -> get the avatar of the user with uuid

image

Subtask of #123

### DIFF
--- a/transcendence-app/src/user/user.controller.ts
+++ b/transcendence-app/src/user/user.controller.ts
@@ -10,6 +10,7 @@ import {
   Post,
   Query,
   ServiceUnavailableException,
+  StreamableFile,
   UnprocessableEntityException,
   UseGuards,
 } from '@nestjs/common';
@@ -22,6 +23,7 @@ import {
   ApiForbiddenResponse,
   ApiNotFoundResponse,
   ApiOkResponse,
+  ApiProduces,
   ApiServiceUnavailableResponse,
   ApiTags,
   ApiUnprocessableEntityResponse,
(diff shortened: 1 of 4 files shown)


Damrod:
Do we need, again to remove the docker volumes?
```
ft_transcendence-transcendence-app-1  | src/shared/local-file/local-file.interceptor.ts:5:29 - error TS7016: Could not find a declaration file for module 'multer'. '/transcendence-app/node_modules/multer/index.js' implicitly has an 'any' type.
ft_transcendence-transcendence-app-1  |   Try `npm i --save-dev @types/multer` if it exists or add a new declaration (.d.ts) file containing `declare module 'multer';`
ft_transcendence-transcendence-app-1  | 
ft_transcendence-transcendence-app-1  | 5 import { diskStorage } from 'multer';
ft_transcendence-transcendence-app-1  |                               ~~~~~~~~
ft_transcendence-transcendence-app-1  | 
ft_transcendence-transcendence-app-1  | [12:51:15 PM] Found 1 error. Watching for file changes.
ft_transcendence-transcendence-app-1  | 
```

gbudau:
I think only the anonymous volumes, the `-V` flag

```
docker compose up --build -d -V
```

Damrod:
Now it finally works, but I had to manually remove all volumes, all containers, and rerun them all.

transcendence-app/src/user/user.controller.ts
```diff
@@ -22,6 +23,7 @@ import {
   ApiForbiddenResponse,
   ApiNotFoundResponse,
   ApiOkResponse,
+  ApiProduces,
```
Damrod, line 26: It's a bit outside of the scope of this PR, but I have just noticed we don't use the `AvatarFileInterceptor` . Are we planning to use it in the future?

Damrod: In fact we are not using the local File interceptor (which seems strange to me...)
What am I missing?

gbudau: You are right. We don't use file interceptors in this PR. We use the `AvatarFileInterceptor` in #126 to save the avatar on disk and check mime type and size.

Damrod: It's PR !126 for ticket #123. That's confusing.

gbudau:
Thank you!

fizcamposancos (commented):
LGTM!

PR #126, opened 2022-08-10
#123/Create controller to update the user avatar

gbudau:
PUT [/api/v1/users/avatar](http://localhost:3000/api#/users/UserController_uploadAvatar) -> update the avatar of the currently authenticated user

image

Subtask of #123

### DIFF
--- a/transcendence-app/src/user/user.service.ts
+++ b/transcendence-app/src/user/user.service.ts
@@ -60,25 +60,26 @@ export class UserService {
     return this.userRepository.updateById(userId, updateUserDto);
   }
 
-  async addAvatar(
+  private async addAvatarAndUpdateUser(
     user: User,
     newAvatarFileDto: LocalFileDto,
   ): Promise<StreamableFile | null> {
-    // If the user doesn't have an avatar yet, save the avatar file in the database and update the user avatarId
-    if (user.avatarId === null) {
-      const updatedUser = await this.userRepository.addAvatarAndUpdateUser(
-        { id: uuidv4(), createdAt: new Date(Date.now()), ...newAvatarFileDto },
-        user,
-      );
-      if (!updatedUser) {
-        this.localFileService.deleteFileData(newAvatarFileDto.path);
-        return null;
-      }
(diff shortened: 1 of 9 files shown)


transcendence-app/src/user/user.service.ts
```diff
@@ -53,4 +59,78 @@ export class UserService {
...
+      disposition: `inline; filename="${fileDto.filename}"`,
+      length: fileDto.size,
+    });
+  }
+
+  async validateAvatarType(path: string): Promise<boolean | undefined> {
+    /**
+     * Import 'file-type' ES-Module in CommonJS Node.js module
+     */
+    const { fileTypeFromFile } = await (eval('import("file-type")') as Promise<
```
Damrod, line 125: >Warning: Executing JavaScript from a string is an enormous security risk. It is far too easy for a bad actor to run arbitrary code when you use eval(). See Never use eval()! , below.

[Here.](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/eval#never_use_eval!)

Damrod:
Why do we need to use eval there?

transcendence-app/src/user/user.service.ts
gbudau: The `file-type` library uses ES modules, and the NestJS compiles into CommonJS modules, which converts the `import` syntax to `require`.

We use `eval` to import the library dynamically at runtime.

gbudau: Maybe we could use something else :thinking:
I will do some research.

Damrod (approved):
Cuando estés conforme, mergeas.

transcendence-app/src/user/user.service.ts
gbudau: Fixed, now we use `Function()` instead of `eval()`. Thank you :rocket:

gbudau:
Thank you! :tada:

fizcamposancos (commented):
LGTM!

PR #128, opened 2022-08-13
#62/Rows should conditionally be a link, not just a button

Damrod:


webapp/src/pages/Users/Users.tsx
```diff
@@ -18,38 +18,35 @@ import {
...
-  username: string;
-  email: string;
-  avatarId: string | null;
-  id: string;
-  createdAt: Date;
+  readonly username: string;
+  readonly email: string;
+  readonly avatarId: string | null;
+  readonly id: string;
+  readonly createdAt: Date;
```
gbudau, line 25: :+1:

webapp/src/pages/Users/Users.tsx
```diff
@@ -18,38 +18,35 @@ import {
...
-        },
-        title: user.username,
-        subtitle: 'level x',
-        key: user.id,
-      });
-    });
-    return rows;
+    const users = (await fetch(url).then((response) => response.json())) ?? [];
+    return {
+      rows: users.map((user: User) => {
```
gbudau, line 34: praise: Well done! I like this more.

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -3,10 +3,12 @@ import './Row.css';
 import { TextColor, TextVariant, TextWeight } from '../Text/Text';
 import { default as Text } from '../Text/Text';
 import { AvatarProps, SmallAvatar } from '../Avatar/Avatar';
+import { Link } from 'react-router-dom';
 
 export type RowProps = {
   iconVariant?: IconVariant;
   onClick?: () => void;
+  url?: string;
```
gbudau, line 11: suggestion: Here we could make this component require one of the `url` or `onClick` and disallow both or none.
https://stackoverflow.com/a/61281828

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -15,38 +17,46 @@ export type RowProps = {
...
-            {subtitle}
-          </Text>
-        )}
-      </div>
-      {iconVariant && (
-        <div className="row-icon">
-          {<Icon variant={iconVariant} size={IconSize.SMALL} />}
-        </div>
+  const RowChildren: JSX.Element[] = [
+    (avatarProps && <SmallAvatar {...avatarProps} />) || <div />,
```
gbudau, line 26: suggestion: Here we could use a [fragment](https://reactjs.org/docs/fragments.html#short-syntax). This way, we could omit the empty divs.

gbudau (approved):
Thank you! LGTM :rocket:

webapp/src/shared/components/Row/Row.tsx
Damrod: Does this solution work for you?

fizcamposancos (commented):
I like it more now

Issue #69, opened 2022-06-25
⚛️ Create user UI page


vdedios:

Damrod:
With larger resolutions, this doesn't look figma accurate: the main grid row gap should be in percent, and the avatar should be aligned to the top left of its cells, not centered

PR #129, opened 2022-08-13
#69/create user IU page

Damrod:
we need a user IU page [here](http://localhost:3000/user)


webapp/src/shared/components/Avatar/Avatar.css
```diff
@@ -31,14 +32,14 @@
 }
 
 .avatar-large__image {
-  width: 200px;
-  height: 200px;
+  width: 10rem;
```
Damrod, line 35: he cambiado esto porque creo que era una conversion de 10 rem de cuando nuestro texto en desktop era 20, y además con el antiguo tamaño, no tenía nada que ver con figma.

webapp/src/pages/UserPage/UserPage.css
```diff
@@ -0,0 +1,84 @@
...
+  }
+  .user-page .header {
+    grid-column: 1 / var(--column-end);
+  }
+
+  .user-wrapper {
+    height: 10.2rem;
+    grid-column: 3 / calc(var(--column-end) - 1);
+    grid-row: 2/3;
+    display: grid;
```
Damrod, line 51: He hecho una grid dentro de una celda de otra grid... a lo mejor no es la mejor idea, pero lo podemos discutir.

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+                : WILDCARD_AVATAR_URL
+            }
+            caption="level 4"
+          />
+        ) : (
+          <></>
+        )}
+        <div className="user-text">
+          <Text
+            variant={TextVariant.PARAGRAPH} // this size doesn't look like in figma
```
Damrod, line 56: // this size doesn't look like in figma, but it's set to figma's setting

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+            color={TextColor.LIGHT} // at least for desktop
+            weight={TextWeight.MEDIUM}
+          >
+            {me.username}
+          </Text>
+          <Text
+            variant={TextVariant.PARAGRAPH} // this size doesn't look like in figma
```
Damrod, line 63: // this size doesn't look like in figma, but it's set to figma's setting

webapp/src/shared/components/Avatar/Avatar.css
Damrod, line 35: I changed this to match what we actually have in figma.

webapp/src/shared/components/Header/Header.tsx
```diff
@@ -5,34 +5,52 @@ import Icon, { IconSize, IconVariant } from '../Icon/Icon';
...
 
-type HeaderProps = {
+type HeaderCommon = {
   navigationFigure: IconVariant | string;
-  navigationUrl: string;
   statusVariant?: StatusVariant;
   children: string;
 };
 
+type HeaderLinkProps = HeaderCommon & {
```
Damrod, line 16: We need a button to go back here, so I changed it on the template of the change I also made for the row component.

webapp/src/pages/Users/Users.tsx
```diff
@@ -8,56 +8,43 @@ import {
...
-              : WILDCARD_AVATAR_URL,
-            status: 'offline',
-          },
-          url: `${USERS_URL}/${user.username}`,
-          title: user.username,
-          subtitle: 'level x',
-          key: user.id,
-        };
-      }),
+      rows: useData<User[]>(USERS_EP_URL, [], instanceOfUserArray).map(
```
Damrod, line 28: I'm not sure it's a good idea to use this hook, as we may want to keep some state here for the search use feature. Also we'd have conflicts with search users form mr

webapp/src/pages/UserPage/UserPage.tsx
Damrod: Well, I found where is the inconsistency:
Screenshot from 2022-08-15 14-24-01
Screenshot from 2022-08-15 14-23-47
Screenshot from 2022-08-15 14-22-37

For desktop, for our src code as it is: font-size is set to 18, paragraph is set to 1rem, which means that paragraph is set to 18 font size.
For figma, caption is set to 18, while caption is set to 21.

As a solution, I have to say that I think font size should stay at 18 for desktop, and it should stay in paragraph. Figma paragraph should match 18 font size for paragraph, and we should rework figma layouts where it doesn't look okay: change some of the texts from paragraph to subheading, and so on. I think this missmatch is exclusive to desktop.

webapp/src/pages/UserPage/UserPage.css
gbudau: I think this is a good approach :+1:. There is nothing wrong with having a grid/flexbox container inside another grid/flexbox item.

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+  TextVariant,
+  TextWeight,
+} from '../../shared/components';
+import { USER_URL, USERS_EP_URL, WILDCARD_AVATAR_URL } from '../../shared/urls';
+import { instanceOfUser, User } from '../../shared/types';
+import { useData } from '../../shared/hooks/UseData';
+import { goBack, logout } from '../../shared/callbacks';
+import { useEffect, useState } from 'react';
+
+export default function UserPage() {
```
gbudau, line 20: praise: Great job! I like how you implemented this page.

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const [me, setMe] = useState<User>({
+    username: '',
+    email: '',
+    avatarId: null,
+    id: '',
+    createdAt: new Date(Date.now()),
+  });
```
gbudau, line 27: thought: In the first render of this page, the user has empty info. We could assign the `me` state to `null` and return the `user-page` or a `Loading` component (for now, a loading… text, until we create this component, it could be a spinner or something like that)

```typescript
  const [me, setMe] = useState<User | null>(null);

  return me === null ? (
    <div>Loading...</div>
  ) : (
    <div className="user-page">
 ...
    </div>
  );
```

webapp/src/shared/types.ts
```diff
@@ -7,3 +7,35 @@ export enum Color {
...
+
+export interface User {
+  readonly username: string;
+  readonly email: string;
+  readonly avatarId: string | null;
+  readonly id: string;
+  readonly createdAt: Date;
+} /* This could be autogenerated - it's provisionally here, since it's a shared type */
+
+export function instanceOfUser(value: object): boolean {
```
gbudau, line 19: thought: Since we are the ones developing the back-end, maybe we could trust ourselves to return valid HTTP status codes and response types (or maybe not? :sweat_smile:)

sugestion: We could check if the response has a valid status code ([Response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok)) and simplify the error checking
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Client-side_web_APIs/Fetching_data#the_can_store

gbudau (approved):
Thank you for this feature. I like how you implement it :rocket:

webapp/src/pages/UserPage/UserPage.css
```diff
@@ -0,0 +1,84 @@
...
+
+  .user-page .button {
+    grid-column: 1 / var(--column-end);
+  }
+}
+
+@media all and (min-width: 1200px) {
+  .user-page-navigation {
+  }
+}
```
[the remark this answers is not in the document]
Damrod: done

webapp/src/pages/UserPage/UserPage.css
```diff
@@ -0,0 +1,84 @@
+.user-page {
```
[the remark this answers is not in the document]
Damrod: done

webapp/src/shared/types.ts
Damrod: What do we even need the openApi generator for, then?

Damrod: I think we should use it, even if we are the ones developing the backend. Not because we don't trust ourselves, but because it's good practice.

webapp/src/pages/UserPage/UserPage.tsx
Damrod: done

webapp/src/shared/callbacks.ts
```diff
@@ -0,0 +1,15 @@
...
+  })
+    .catch((e) => console.error(e))
+    .finally(() => {
+      window.location.href = HOST_URL;
+    });
+};
+
+export const goBack = () => {
+  window.history.back();
+};
```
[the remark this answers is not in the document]
Damrod: done. Do you like it like this? Why do you think this is better than just window.history.back() ?

webapp/src/shared/types.ts
gbudau: Ok, good point :+1:

webapp/src/shared/callbacks.ts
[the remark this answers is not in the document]
gbudau: Looks great, thank you. The [documentation](https://reactrouter.com/docs/en/v6/getting-started/concepts#history-object) of React Router recommends not using `window.history` or `window.location` directly for navigating inside the application.

PR #133, opened 2022-08-15
#132/generate webapp interfaces using typescript fetch openapi generator

### LINKED ISSUE #132: Generate webapp interfaces using typescript-fetch openapi-generator
https://openapi-generator.tech/docs/generators/typescript-fetch/

gbudau:
close #132


Damrod:
Is everything in the folder `webapp/src/shared/generated/` generated? If it is, as I would assume: from the same instance where the content of that folder was created from, could I erase everything there, run the script, and have the same? If this is the way it's supposed to work, maybe we should document it (add something to the readme).
Besides this, do you think it's outside the scope of this pr to remove the generated types I provisionally added to webapp/shared/types.ts to see the generated code in action?

Thx for the feature! Looks great!

Damrod (approved):
Besides the comment, lgtm!

gbudau:
> Is everything in the folder webapp/src/shared/generated/ generated? If it is, as I would assume: from the same instance where the content of that folder was created from, could I erase everything there, run the script, and have the same? If this is the way it's supposed to work, maybe we should document it (add something to the readme).
Besides this, do you think it's outside the scope of this pr to remove the generated types I provisionally added to webapp/shared/types.ts to see the generated code in action?

Yes, the script generates all the files from the `webapp/src/shared/generated/` directory. Good point. I'll update the README.md with this info and update the types to use the generated ones.

gbudau:
Thank you! :rocket:

fizcamposancos (commented):
i would need further explanations about this tool. But so far it looks great

PR #134, opened 2022-08-15
user/:id page

Damrod:
we need a ui to see user pages other than for self

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -24,7 +24,8 @@ root.render(
         <Route path={USERS_URL} element={<Users />} />
         <Route path={PLAY_URL} element={<Play />} />
         <Route path={CHAT_URL} element={<Chat />} />
-        <Route path={USER_URL} element={<UserPage />} />
+        <Route path={USER_URL} element={<UserPage isMe={true} />} />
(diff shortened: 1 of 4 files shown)


webapp/src/index.tsx
```diff
@@ -24,7 +24,8 @@ root.render(
         <Route path={USERS_URL} element={<Users />} />
         <Route path={PLAY_URL} element={<Play />} />
         <Route path={CHAT_URL} element={<Chat />} />
-        <Route path={USER_URL} element={<UserPage />} />
+        <Route path={USER_URL} element={<UserPage isMe={true} />} />
+        <Route path={`${USERS_URL}/:id`} element={<UserPage isMe={false} />} />
```
gbudau, line 28: Very cool :rocket:

gbudau (approved):
I like how you solve this by adding a prop to the `UserPage`!
LGTM :tada:

webapp/src/index.tsx
Damrod: thx!

Damrod: thx!

fizcamposancos (commented):
Good job!!

Issue #64, opened 2022-06-25
⚛️ Create editable avatar UI component


vdedios:

Damrod:
We still need to take the coordinates from the user dragging, and save it somewhere.

Damrod:
Done

PR #137, opened 2022-08-16
#64/create editable avatar component

Damrod:
Crete editable Avatar draft


webapp/src/shared/components/Avatar/EditableAvatar.tsx
```diff
@@ -0,0 +1,48 @@
+import './EditableAvatar.css';
+import { AvatarProps } from './Avatar';
+import useDrag from '../../hooks/UseDrag';
+import { Text, TextColor, TextVariant, TextWeight } from '../index';
+
+export function EditableAvatar({ url }: AvatarProps) {
```
gbudau, line 6: What do we think about cropping the avatar before upload? This way, we could use our current Avatar components and the API endpoints as they are.

I am concerned that with this approach, we save frontend logic in the backend. We have to persist the position of how the user 'crop' their avatar in the database, and when displaying it, we have to use that position again.

Damrod: We query the BE too get all user data. X and y of the Avatar would be just two more number fields: doesn't seem like an overkill. We only need to modify the patch negros for users, and add two fields . And like this, it's reversible, if the user devices he wants a different cropping.

Damrod: About keeping the Avatar the way it currently is: I don't know about yours, but mine is completely deformed. The way it is only works if your picture has some particular proportions.

Damrod: Also: we don't really save logic
 We save data that we intend to persist, as we should. It just happens to be about the display of things.

fizcamposancos (commented):
In Chrome it works well but in Firefox it doesn't. I think it is because of the browser that thinks you are going to drag the picture elsewhere and changes from "our" drag to "its" drag.
Good work!!

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -7,20 +7,40 @@ import { Color } from '../../types';
...
+export function SmallAvatar({
+  url,
+  status,
+  XCoordinate,
+  YCoordinate,
+}: AvatarProps) {
+  const position =
+    XCoordinate && YCoordinate
+      ? {
+          objectPosition: `${XCoordinate * 0.139}px ${YCoordinate * 0.139}px`,
```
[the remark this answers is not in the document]
Damrod: These strange numbers were found experimentally via trial and error: I tried to see the same image (for usefulness I used a chess like grid, with different numbers and colors on each cell) in the three sizes of Avatar we have, 
so that they would look the same way with the same x y coordinates. These coefficients make the three be in sync. When you set the position from the editable Avatar component, you can freely pass the same coordinates to Avatar small and large, and they'll look just like editable Avatar.

webapp/src/shared/components/Avatar/EditableAvatar.tsx
gbudau: I don't understand why we want to make the cropping reversible. If the user wants different cropping, they could upload the image again.

Anyway, even if I think the current solution is not optimal, it's pretty creative and could work.

gbudau (approved):
Thank you! LGTM :rocket:

webapp/src/shared/hooks/UseDrag.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const { isDragging } = dragInfo;
+  const handleMouseDown = ({ clientX, clientY }: React.MouseEvent) => {
+    if (!isDragging)
+      setDragInfo({
+        ...dragInfo,
+        isDragging: true,
+        origin: { x: clientX, y: clientY },
+      });
+  };
```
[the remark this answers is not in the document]
fizcamposancos: Yes, now it works smoothly

webapp/src/shared/components/Avatar/EditableAvatar.css
```diff
@@ -0,0 +1,36 @@
...
+.editable-avatar-wrapper {
+  margin: -0.5rem;
+  background-color: var(--color-light);
+  width: fit-content;
+  clip-path: inset(0.55rem 0.55rem 0.55rem 0.55rem round 0.55rem);
+  border: none;
+}
+
+.editable-avatar__image {
+  user-drag: none; /* TODO: none of these work for firefox: research */
```
fizcamposancos, line 18: Despite the comment and that VSCode has a curly yellow line underneath these css properties, I can see no difference between Firefx and Chrome behaviors .

fizcamposancos (approved):
Good work!!

webapp/src/shared/components/Avatar/EditableAvatar.css
Damrod: There was:
as it was, I was disabling all dragging of all elements in the page, not disabling dragging only for the component that needs to have its dragging not intervened by the browser.
But it's fixed now, so should be okay.

webapp/src/shared/hooks/UseDrag.tsx
```diff
@@ -0,0 +1,66 @@
...
+      const { translation } = dragInfo;
+      setDragInfo({
+        ...dragInfo,
+        isDragging: false,
+        lastTranslation: { x: translation.x, y: translation.y },
+      });
+    }
+  };
+
+  const picturePosition = {
```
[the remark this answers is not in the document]
Damrod: done! Thx!

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -31,9 +51,19 @@ export function LargeAvatar({
...
   caption,
+  XCoordinate,
+  YCoordinate,
 }: LargeAvatarProps) {
   const statusClass = status ? `avatar-status--${status}` : '';
 
+  const position =
+    XCoordinate && YCoordinate
+      ? {
+          objectPosition: `${XCoordinate * 0.55}px ${YCoordinate * 0.55}px`,
```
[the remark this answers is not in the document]
Damrod: As mentioned above: 
> These strange numbers were found experimentally via trial and error[...]

PR #138, opened 2022-08-17
#71/Create edit user page

### LINKED ISSUE #71: ⚛️ Create /user/edit UI page

fizcamposancos:
Closes #71 

Now works with a confirmation window and an alert. We can change it to a better approach if you want.

### DIFF
--- a/webapp/src/pages/EditUserPage/EditUserPage.css
+++ b/webapp/src/pages/EditUserPage/EditUserPage.css
@@ -5,35 +5,23 @@
   background-color: var(--color-background);
   color: var(--color-light);
   display: grid;
-  grid-column-gap: var(--column-gap);
+  gap: 4rem var(--column-gap);
   grid-template-columns: 1fr;
-  grid-template-rows: auto 1fr auto;
-  grid-row-gap: 4rem;
+  grid-template-rows: auto auto 1fr auto;
   overflow: scroll;
 }
-.edit-user-container {
-  height: 100%;
-}
-.form-container {
-  height: 24.5rem;
-  display: flex;
-  flex-direction: column;
-  justify-content: space-between;
+
+.edit-user-form {
+  min-width: 15rem;
+  max-width: 20rem;
+  justify-self: center;
 }
 
-.form-container .edit-user-form {
-  width: 23rem;
-  height: 100vh;
-  color: var(--color-light);
-  display: flex;
-  flex-direction: column;
-  justify-content: space-between;
-  align-items: center;
+.edit-user-form-button {
+  justify-self: center;
+  align-self: end;
 }
 
-.form-container .edit-user-form .inputs-container {
-  width: 100%;
-}
 
 @media all and (min-width: 768px) {
   .edit-user-page {
@@ -48,15 +36,24 @@
   }
 
   .avatar-large {
-    grid-column: 3/6;
+    grid-column: 2/6;
     grid-row: 2/3;
     align-self: start;
   }
-  .form-container {
+  .edit-user-form {
     grid-column: 7/12;
   }
-  .form-container .edit-user-form .button {
-    display: flex;
-    align-self: flex-start;
+  .edit-user-form-button {
+    grid-column: 5/9;
+  }
+}
+
+@media (min-width: 1024px) {
+  .avatar-large {
+    grid-column: 3/6;
+  }
+
+  .edit-user-form {
+    grid-column: 7/11;
   }
 }
--- a/webapp/src/pages/EditUserPage/EditUserPage.css
+++ b/webapp/src/pages/EditUserPage/EditUserPage.css
@@ -1,0 +1,49 @@
+.edit-user-page {
+  margin: 0;
+  padding: var(--padding-top-bottom) var(--padding-sides);
+  height: 100vh;
+  background-color: var(--color-background);
+  color: var(--color-light);
+  display: grid;
+  grid-column-gap: var(--column-gap);
+  grid-template-columns: 1fr;
+  grid-template-rows: auto 1fr auto;
+  grid-row-gap: 4rem;
+  overflow: scroll;
+}
+.edit-user-container {
+  height: 100%;
+}
+.form-container {
+  height: 24.5rem;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+}
+
+@media all and (min-width: 768px) {
+  .edit-user-page {
+    display: grid;
+    grid-template-columns: repeat(var(--column-number), 1fr);
+    grid-template-rows: auto 1fr auto;
+    grid-row-gap: 8rem;
+  }
+  .edit-user-page .header {
+    grid-row: 1;
+    grid-column: 1 / var(--column-end);
+  }
+
+  .avatar-large {
+    grid-column: 3/6;
+    grid-row: 2/3;
+  }
+
+  .form-container {
+    grid-column: 5/10;
+    height: 4.8rem;
+  }
+  .form-container .edit-user-form .button {
+    grid-column: 3/7;
+    grid-row: 2;
+  }
+}
--- a/webapp/src/pages/EditUserPage/EditUserPage.css
+++ b/webapp/src/pages/EditUserPage/EditUserPage.css
@@ -22,7 +22,6 @@
   align-self: end;
 }
 
-
 @media all and (min-width: 768px) {
   .edit-user-page {
     display: grid;
--- a/webapp/src/pages/EditUserPage/EditUserPage.css
+++ b/webapp/src/pages/EditUserPage/EditUserPage.css
@@ -21,6 +21,20 @@
   justify-content: space-between;
 }
 
+.form-container .edit-user-form {
+  width: 23rem;
+  height: 100vh;
+  color: var(--color-light);
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.form-container .edit-user-form .inputs-container {
+  width: 100%;
+}
+
 @media all and (min-width: 768px) {
   .edit-user-page {
     display: grid;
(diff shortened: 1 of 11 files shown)


webapp/src/shared/components/Input/EditUserForm.tsx
```diff
@@ -2,39 +2,85 @@ import { useState } from 'react';
...
 
-export default function EditUserForm() {
-  const [fullName, setFullName] = useState('');
-  const [email, setEmail] = useState('');
+type EditUserProps = {
+  origFullName: string;
+  origEmail: string;
+};
+
+export default function EditUserForm({
```
Damrod, line 10: Just like we have the generic wrapper around fetch for GET requests that is the hook UseData, maybe we could have a generic wrapper around other request types? Maybe no...

Damrod: Maybe instead we can make generic the EditUserForm itself, into EditForm, by passing from the parent component, the request options, the url endpoint, and the placeholders ... ?
Or maybe it's not to be made generic. Not sure.

webapp/src/shared/components/Input/EditUserForm.tsx
```diff
@@ -2,39 +2,85 @@ import { useState } from 'react';
...
+      method: 'PATCH',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        fullName,
+        email,
+      }),
+    };
+    await fetch('http://localhost:3000/api/v1/users', requestOptions);
```
Damrod, line 28: praise:
thanks for connecting it to the backend!

gbudau, line 28: suggestion (non-blocking): This might be out of the scope of this PR.

We could use the UsersApi service from the generated OpenAPI spec.
We could have a status message which shows after the user submits the form (or a Notification component?).

Something like this:

<details>

```diff
diff --git a/webapp/src/shared/components/Input/EditUserForm.tsx b/webapp/src/shared/components/Input/EditUserForm.tsx
index 6dbe974..5b6b0b9 100644
--- a/webapp/src/shared/components/Input/EditUserForm.tsx
+++ b/webapp/src/shared/components/Input/EditUserForm.tsx
@@ -1,49 +1,61 @@
-import { useState } from 'react';
-import { Button, ButtonVariant, Input, InputVariant } from '../';
+import { useEffect, useState } from 'react';
+import {
+  Button,
+  ButtonVariant,
+  Input,
+  InputVariant,
+  Text,
+  TextColor,
+  TextVariant,
+} from '../';
+import { Configuration, UsersApi } from '../../generated';
 import './EditUserForm.css';
 
+// We could move this part to another file and import it from there. shared/services/ApiService.ts?
+const config = new Configuration({ basePath: window.location.origin });
+const usersApi = new UsersApi(config);
+
 type EditUserProps = {
   origFullName: string;
   origEmail: string;
 };
 
+type EditUserFormStatus = {
+  type: 'success' | 'error';
+  message: string;
+};
+
 export default function EditUserForm({
   origFullName,
   origEmail,
 }: EditUserProps) {
   const [fullName, setFullName] = useState(origFullName);
   const [email, setEmail] = useState(origEmail);
+  const [status, setStatus] = useState<EditUserFormStatus | null>(null);
 
   async function updateData() {
-    const requestOptions = {
-      method: 'PATCH',
-      headers: {
-        'Content-Type': 'application/json',
-      },
-      body: JSON.stringify({
-        fullName,
-        email,
-      }),
-    };
-    await fetch('http://localhost:3000/api/v1/users', requestOptions);
+    try {
+      await usersApi.userControllerUpdateCurrentUser({
+        updateUserDto: { fullName, email },
+      });
+      setStatus({ type: 'success', message: 'Update successfully' });
+    } catch {
+      setStatus({ type: 'error', message: 'Invalid name or email' });
+    }
   }
 
   const handleOnSubmit = (e: React.FormEvent<HTMLFormElement>): void => {
     e.preventDefault();
-    let txt;
-    if (
-      window.confirm(
-        `Your name: "${fullName}" and your email: "${email}".\n
-        Is this correct?`,
-      )
-    ) {
-      txt = 'Your data has been saved';
-      updateData();
-    } else {
-      txt = 'Canceled!';
-    }
-    alert(txt);
+    updateData();
   };
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setStatus(null);
+    }, 3000);
+    return () => clearTimeout(timer);
+  }, [status]);
+
   return (
     <>
       <form
@@ -73,6 +85,16 @@ export default function EditUserForm({
             }}
           />
         </div>
+        {status && (
+          <Text
+            variant={TextVariant.PARAGRAPH}
+            color={
+              status.type === 'success' ? TextColor.ONLINE : TextColor.OFFLINE
+            }
+          >
+            {status.message}
+          </Text>
+        )}
       </form>
       <div className="edit-user-form-button">
         <Button

```

</detais>

webapp/src/shared/components/Input/EditUserForm.tsx
```diff
@@ -2,39 +2,85 @@ import { useState } from 'react';
 import { Button, ButtonVariant, Input, InputVariant } from '../';
 import './EditUserForm.css';
 
-export default function EditUserForm() {
-  const [fullName, setFullName] = useState('');
-  const [email, setEmail] = useState('');
+type EditUserProps = {
+  origFullName: string;
+  origEmail: string;
+};
```
gbudau, line 8: Good idea to add these props :+1:

webapp/src/pages/EditUserPage/EditUserPage.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.edit-user-form-button {
+  justify-self: center;
+  align-self: end;
+}
+
+@media all and (min-width: 768px) {
+  .edit-user-page {
+    display: grid;
```
Damrod, line 27: if you are not overriding the options of the mobile version, it's useless to set it inside the media query. I.e.: we already set `edit-user-page` to `display: grid` in line 7. We don't need to do it again.

gbudau (approved):
Well done! Thank you for this feature. LGTM :rocket:

webapp/src/pages/EditUserPage/EditUserPage.css
```diff
@@ -0,0 +1,58 @@
...
+    grid-template-columns: repeat(var(--column-number), 1fr);
+    grid-template-rows: auto 1fr auto;
+    grid-row-gap: 8rem;
```
Damrod, line 30: I used this ` gap: 4rem var(--column-gap);` in the user page in the beginning because I thought it was convenient, but the semantics of the gap (that you cannot configure independently on a particular per-gap basis) do not mesh well with the one-time nature of this particular gap. Also, as it is now, it is off sync with the users page. I would match it with what you have there:
Five rows in total, the second one replaces this gap, the fourth one replaces your second gap.
(template should be something like     `grid-template-rows: auto 20% auto 1fr auto`). Then you'd have to fix the respective rows of the actual classes you have in the selectors below...

webapp/src/pages/EditUserPage/EditUserPage.css
```diff
@@ -0,0 +1,58 @@
...
+  }
+  .edit-user-form {
+    grid-column: 7/12;
+  }
+  .edit-user-form-button {
+    grid-column: 5/9;
+  }
+}
+
+@media (min-width: 1024px) {
```
Damrod, line 50: we kind of agreed that the bigger version would start at 1200px... do you mind?

webapp/src/pages/EditUserPage/EditUserPage.css
fizcamposancos: Yes, you're right! Thnaks

webapp/src/pages/EditUserPage/EditUserPage.css
fizcamposancos: The result is the same but the code is more consistent. Done

webapp/src/pages/EditUserPage/EditUserPage.css
fizcamposancos: No problem. Done!

webapp/src/pages/EditUserPage/EditUserPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+  TextWeight,
+} from '../../shared/components';
+import { USERS_EP_URL, WILDCARD_AVATAR_URL } from '../../shared/urls';
+import { useData } from '../../shared/hooks/UseData';
+import { goBack } from '../../shared/callbacks';
+import { useEffect, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { instanceOfUser, User } from '../../shared/generated';
+
+export default function EditUserPage() {
```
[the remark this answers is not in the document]
fizcamposancos: Thanks, I am learning from the best colleagues.

webapp/src/pages/EditUserPage/EditUserPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+    </div>
+  ) : (
+    <div className="edit-user-page">
+      <Header
+        navigationFigure={IconVariant.ARROW_BACK}
+        onClick={goBack(navigate)}
+      >
+        edit profile
+      </Header>
+      <div className="avatar-large">
```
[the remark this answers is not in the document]
fizcamposancos: Yes, in fact I had some issues with this naming and I thought I've changed it. Anyway, I change it now

webapp/src/shared/components/Input/EditUserForm.tsx
fizcamposancos: Thanks

webapp/src/shared/components/Input/EditUserForm.tsx
fizcamposancos: We can talk about it. Anyway I going to double-check what we can do

webapp/src/shared/components/Input/EditUserForm.tsx
fizcamposancos: Thanks

Damrod (approved):
Lgtm, thank you!

webapp/src/shared/components/Input/EditUserForm.tsx
fizcamposancos: If you don't mind I am going to create another issuer and I'll try it

webapp/src/shared/components/Input/EditUserForm.tsx
fizcamposancos: And I'll create an issue in order not to forget it

PR #139, opened 2022-08-19
#68/create draft UI chat page

Damrod:
a generic dispatch page solution is considered as a good choice for chats plays and users pages


webapp/src/pages/Users/Users.tsx
```diff
@@ -1,66 +1,34 @@
...
-      </div>
-      <div className="users-navigation">
-        <NavigationBar />
-      </div>
-    </div>
+    <DispatchPage
+      dataValidator={instanceOfUser}
+      genericEndpointUrl={USERS_EP_URL}
+      dataMapper={mapUserToRow}
+    />
```
gbudau, line 32: question: If I understand correctly, we will use the same layout for the Users, Play and Chat pages?

thought: As long as we keep this in mind, it will allow us to have a consistent UI between these pages and reduce code duplication.

gbudau (approved):
I like how you solve this. LGTM :rocket:

webapp/src/pages/Users/Users.tsx
Damrod: That is the general idea, yes. But who knows, we'll see as we go forward.

fizcamposancos (approved):
Nice! Good approach for the common parts of the pages

PR #145, opened 2022-08-19
#80/Create react context to protect routes

### LINKED ISSUE #79: ⚛️ Create api service to make requests to backend
As part of this task:

- [x] https://github.com/42AGV/ft_transcendence/issues/81

### LINKED ISSUE #80: ⚛️ Create React context to protect routes

### LINKED ISSUE #81: ⚛️ Investigate if we could have a container (context) to hold data in the application

gbudau:
close #79
close #80
close #81


webapp/src/index.tsx
```diff
@@ -28,18 +30,62 @@ const developmentMode = process.env.NODE_ENV === 'development';
...
+          <Route
+            path={PLAY_URL}
+            element={
+              <RequireAuth>
+                <Play />
+              </RequireAuth>
+            }
+          />
+          <Route
+            path={CHAT_URL}
```
Damrod, line 53: Screenshot from 2022-08-22 15-00-49
There is error in the console when entering chat or play routes.

webapp/src/context/auth-context.tsx
```diff
@@ -0,0 +1,40 @@
...
+  useEffect(() => {
+    setIsLoading(true);
+    usersApi
+      .userControllerGetCurrentUser()
+      .then((currentUser) => setUser(currentUser))
+      .catch(() => setUser(null))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const logout = async (callback: VoidFunction) => {
```
Damrod, line 26: Twenty lines of code, two ways of writing async functions.

webapp/src/pages/Landing/Landing.tsx
```diff
@@ -38,7 +48,9 @@ export default function Landing() {
       <div className="landing-login-button">
         <Button
           variant={ButtonVariant.SUBMIT}
-          onClick={() => (window.location.href = LOGIN_EP_URL)}
+          onClick={() =>
+            window.location.replace(`${LOGIN_EP_URL}?state=${from}`)
```
Damrod, line 52: why is `location.replace(...)` a better idiom than `location.href`?

webapp/src/shared/components/RequireAuth/RequireAuth.tsx
```diff
@@ -0,0 +1,26 @@
...
+      <div className="require-auth">
+        <Text variant={TextVariant.SUBHEADING} weight={TextWeight.MEDIUM}>
+          Loading...
+        </Text>
+      </div>
+    );
+  }
+
+  if (!auth.user) {
+    return <Navigate to="/" state={{ from: location }} replace />;
```
Damrod, line 22: `HOST_URL` ?

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -88,7 +90,7 @@ export default function UserPage({ isMe = false }: UserPageProps) {
         <Button
           variant={ButtonVariant.WARNING}
           iconVariant={IconVariant.LOGOUT}
-          onClick={logout(navigate)}
+          onClick={() => logout(() => navigate('/'))}
```
Damrod, line 93: `HOST_URL`?

webapp/src/pages/Landing/Landing.tsx
gbudau: `location.replace()` redirects to the URL and replaces the current resource with the one at the provided URL, meaning the user won't be able to use the back button to navigate to it. (Source [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Location/replace))

I think this is better because after the user login, they can't go back to the login page if they press the back button.

webapp/src/context/auth-context.tsx
gbudau: I've changed it to use the `useData` hook. Please let me know what you think.

webapp/src/context/auth-context.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const AuthProvider = ({ children }: { children: ReactNode }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    setIsLoading(true);
+    usersApi
+      .userControllerGetCurrentUser()
+      .then((currentUser) => setUser(currentUser))
+      .catch(() => setUser(null))
```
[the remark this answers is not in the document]
gbudau: Done (added logging to the `useData` hook)

webapp/src/index.tsx
gbudau: This error was because the `DispatchPage` from `Play` and `Chat` pages had an undefined `genericEndpointUrl`.

I've marked these props as required and copied/pasted the code from the `Users` page until we implement them.

Not sure if it's the best solution, but it works. Please let me know what you think.

webapp/src/shared/components/RequireAuth/RequireAuth.css
```diff
@@ -0,0 +1,10 @@
+.require-auth {
+  min-height: 100vh;
+  background-color: var(--color-background);
+  color: var(--color-light);
```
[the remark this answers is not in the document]
gbudau: Since we use these styles in all the pages, I think we could add it to the `body` in the `index.css` and remove it from all the pages (including this one). What do we think about this?

fizcamposancos:
I don't know why, if something related with cache or it is just the way it works, but if I am in the UserPage, then logout with the swagger UI from another tab, come back to UserPage, I can click and enter EditUserPage and even try to change the fields and the get messages. If I reload the site, then I am redirected to the landing page. But it seems like an strange behavior

webapp/src/index.tsx
Damrod: sorry, this was my fault, then.
Anyway, thank you for fixing it.

webapp/src/context/auth-context.tsx
Damrod: I think it looks good.

Damrod (approved):
LGTM!

gbudau:
@fizcamposancos, I think this happens because when we access the logout endpoint from the Swagger or directly with a DELETE request to `/api/v1/auth/logout`, this logs out the user from the transcendence-app.

To log out from the webapp, we call the `logout` method from the `AuthContext`, which also sets the `user` to `null` inside the auth context.

I see two solutions:
- Do nothing and leave it like this, if the users log out without using the logout button from the webapp, it's their choice (and problem :smile:)
- Check if there was a 403 (Forbidden) response after each fetch to our API, and call `AuthContext.logout()` then redirect to the login page. Doable but a bit harder to maintain.
  Something like this:
  ```tsx
  const { logout } = useAuth();
  const { data, error } = useData(getUserById);
  
  if (error && isAuthError(error)) {
    logout();
    return <Navigate to={HOST_URL} state={{ from: location }} replace />;
  }
  ```

gbudau:
Thank you!
If we want to deal with the case when the user logs out from outside the webapp, and we like the above solution, or we have another solution, please let me know, and I will open a new issue and PR for it.

Issue #150, opened 2022-08-23
⚛️ Create edit avatar page


Damrod:
we need to let the users edit their avatar

PR #152, opened 2022-08-24
#64/add avatar coords fields to user

Damrod:
we need endpoints to store the clipping of the avatar the user selected


webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -12,28 +12,40 @@ import {
...
+        method: 'PATCH',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          avatarX: picturePosition.x,
+          avatarY: picturePosition.y,
+        }),
+      };
+      console.log(requestOptions);
```
Damrod, line 43: Remind the guys

webapp/src/pages/ComponentsBook/ComponentsExamples/HeaderExample.tsx
```diff
@@ -14,7 +15,7 @@ export const HeaderExample = () => (
       {placeholderTitle}
     </Header>
     <Header
-      navigationFigure={randomAvatar}
+      navigationFigure={JSON.stringify(avProps)}
```
Damrod, line 18: This is questionable, but it appeared to me as the simplest way to achieve the passing of the coordinates to the avatar, here, with the least amount of changes to the component as it is.

webapp/src/pages/EditAvatarPage/EditAvatarPage.css
```diff
@@ -14,9 +14,16 @@
   overflow: scroll;
 }
 
+.edit-avatar-page .editable-avatar {
+  align-self: flex-start;
+}
+
 .edit-avatar-page .row {
-  width: 18rem;
+  width: 15rem;
```
Damrod, line 22: I prefer it bigger, but this is the way it is right now in figma

webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -7,6 +7,7 @@ import { Link } from 'react-router-dom';
 
 export type AvatarProps = {
   url: string;
+  imgHash?: number /* TODO: fix auto reloads when avatar is updated, possibly with this */;
```
Damrod, line 10: When we update the avatar, your browser's cache will probably not update all other instances of avatars in the parent pages, unless you force a reload. This should be improved.

Damrod:
I have split this MR into yet another branch. #155 needs to be merged into this one, before this one can be merged into #150

gbudau (approved):
Thank you! LGTM :rocket:

webapp/src/pages/ComponentsBook/ComponentsExamples/HeaderExample.tsx
gbudau: question: What do we think about splitting this into two props: `iconVariant` and `avatarProps`? And require one or the other.
Or pass the JSX element directly?

I'm concerned that we add hacks on top of hacks (the previous way was also a hack :sweat_smile:) and risk creating something pretty hard to understand down the road.

Damrod: I think it's a good idea. Will do.

Damrod: How would you do the part where we already have two union types for link and button? If we add two variants, I'll have to add two variants for button and two variants for link. Isn't this making it very complicated very soon? (There is a technical term for this problem, but I cannot recall it)

gbudau: We could do something like this:

<details>

```diff
diff --git a/webapp/src/shared/components/Header/Header.tsx b/webapp/src/shared/components/Header/Header.tsx
index 848bdd6..012b056 100644
--- a/webapp/src/shared/components/Header/Header.tsx
+++ b/webapp/src/shared/components/Header/Header.tsx
@@ -8,25 +8,37 @@ import './Header.css';
 import React from 'react';
 
 type HeaderCommon = {
-  navigationFigure: IconVariant | string;
   statusVariant?: StatusVariant;
   children: string;
 };
 
-type HeaderLinkProps = HeaderCommon & {
+type HeaderLinkNavigation = {
   navigationUrl: string;
   onClick?: never;
 };
 
-type HeaderButtonProps = HeaderCommon & {
+type HeaderButtonNavigation = {
   navigationUrl?: never;
   onClick: React.MouseEventHandler<HTMLButtonElement>;
 };
 
-type HeaderProps = HeaderLinkProps | HeaderButtonProps;
+type HeaderIconFigure = {
+  iconVariant: IconVariant;
+  avatarProps?: never;
+};
+
+type HeaderAvatarFigure = {
+  iconVariant?: never;
+  avatarProps: AvatarProps;
+};
+
+type HeaderProps = HeaderCommon &
+  (HeaderLinkNavigation | HeaderButtonNavigation) &
+  (HeaderIconFigure | HeaderAvatarFigure);
 
 export default function Header({
-  navigationFigure,
+  iconVariant,
+  avatarProps,
   navigationUrl,
   onClick,
   statusVariant,
@@ -34,15 +46,10 @@ export default function Header({
 }: HeaderProps) {
   const lNavFigure = (
     <>
-      {navigationFigure in IconVariant ? (
-        <Icon
-          variant={navigationFigure as IconVariant}
-          size={IconSize.SMALL}
-          color={Color.LIGHT}
-        />
-      ) : (
-        <SmallAvatar {...(JSON.parse(navigationFigure) as AvatarProps)} />
+      {iconVariant && (
+        <Icon variant={iconVariant} size={IconSize.SMALL} color={Color.LIGHT} />
       )}
+      {avatarProps && <SmallAvatar {...avatarProps} />}
     </>
   );
   return (

```

</details>


You are right. It is a bit convoluted if we want to keep the types and conditional props.

Damrod: This should be fixed now (following your last proposition). Let me know if this works for you.

gbudau: Thank you! Looks very good!

webapp/src/pages/EditAvatarPage/EditAvatarPage.css
```diff
@@ -14,9 +14,25 @@
   overflow: scroll;
 }
 
+.edit-avatar-page .editable-avatar {
+  align-self: flex-start;
```
Damrod, line 18: for mobile looks much better as a grid, than as a flex with space between, so I'll not change it. I'm following figma.

PR #153, opened 2022-08-28
#146/Add pagination to the Users UI page

### LINKED ISSUE #146: ⚛️ Add pagination to the Users UI page
Currently, it displays max 20 users. We could either load more users on scrolling or add page buttons

gbudau:
[Infinite Scrolling With React](https://www.youtube.com/watch?v=NZKUirTtxcg)
[IntersectionObserver](https://www.youtube.com/watch?v=2IbRtjez6ag)

close #146

### DIFF
--- a/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
+++ b/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
@@ -30,11 +30,14 @@ export default function DispatchPage<T>({
   dataMapper,
   button,
 }: DispatchPageProps<T>) {
-  const [search, setSearch] = useState('');
-  const [offset, setOffset] = useState(0);
+  const [searchParams, setSearchParams] = useState<SearchFetchFnParams>({
+    search: '',
+    offset: 0,
+  });
+  const { search } = searchParams;
   const getData = useCallback(() => {
-    return fetchFn({ search, offset });
-  }, [fetchFn, search, offset]);
+    return fetchFn(searchParams);
+  }, [fetchFn, searchParams]);
   const { data, isLoading, hasMore } = UseSearch(search, getData);
   const observer = useRef<IntersectionObserver | null>(null);
 
@@ -48,7 +51,10 @@ export default function DispatchPage<T>({
       }
       observer.current = new IntersectionObserver((entries) => {
         if (entries[0].isIntersecting && hasMore) {
-          setOffset(data.length);
+          setSearchParams((prevSearchParams) => ({
+            ...prevSearchParams,
+            offset: data.length,
+          }));
         }
       });
       if (row) {
@@ -76,8 +82,7 @@ export default function DispatchPage<T>({
   };
 
   const handleSearch = (value: string) => {
-    setSearch(value);
-    setOffset(0);
+    setSearchParams({ search: value, offset: 0 });
   };
 
   return (
--- a/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
+++ b/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
@@ -91,14 +91,12 @@ export default function DispatchPage<T>({
         <SearchForm search={search} onSearchChange={handleSearch} />
       </div>
       <div className="dispatch-page-rows">
-        {dataValidator &&
-          instanceOfArrayTyped(data, dataValidator) &&
-          dataMapper && (
-            <RowsList
-              rows={mapDataToRows(dataMapper, data)}
-              lastRowRef={lastRowElementRef}
-            />
-          )}
+        {instanceOfArrayTyped(data, dataValidator) && (
+          <RowsList
+            rows={mapDataToRows(dataMapper, data)}
+            lastRowRef={lastRowElementRef}
+          />
+        )}
       </div>
       {button && <div className="dispatch-page-button">{button}</div>}
       <div className="dispatch-page-navigation">
--- a/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
+++ b/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
@@ -1,11 +1,12 @@
 import './DispatchPage.css';
-import { useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { Link } from 'react-router-dom';
 import { USER_URL, USERS_EP_URL } from '../../urls';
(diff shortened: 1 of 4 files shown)


webapp/src/shared/components/DispatchPage/DispatchPage.tsx
```diff
@@ -20,7 +30,33 @@ export default function DispatchPage<T>({
...
+
+  const lastRowElementRef = useCallback(
+    (row: HTMLLIElement) => {
+      if (isLoading) {
+        return;
+      }
+      if (observer.current) {
+        observer.current.disconnect();
+      }
+      observer.current = new IntersectionObserver((entries) => {
```
Damrod, line 49: It's a bit hard to follow. I'm assuming you are creating a last row item with a special behaviour, but I'm not sure yet. Do you have an easy way to test it or should I manually add dozens of users to the database until some offset and limit are meaningful?

gbudau: How it works:

We assign a callback ref to the last row item. React will call that callback whenever the ref gets attached to a different row. When the component mounts, it calls the callback with the element, and when it unmounts, it calls with null as the parameter.

This article helped me to understand how callback refs work: https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs.

Inside the callback ref, we create an `IntersectionObserver` and set it to observe the last row item.
The observer also has a callback and calls it when an observed element intersects with the viewport (when it appears or disappears from the screen).

```tsx
(entries) => {
        if (entries[0].isIntersecting && hasMore) {
          setOffset(data.length);
        }
      }
```

When the `offset` changes, the `getData` method updates, which triggers the `useEffect` from the `useData` hook.

I hope this helps, and sorry, it is a bit confusing.

gbudau: > Do you have an easy way to test it or should I manually add dozens of users to the database until some offset and limit are meaningful?

We could create a script to seed the database with random users. I will look into it.

gbudau: To test this until we implement a script to seed the database, please go to https://www.mockaroo.com/ea0e3c90. Download the `SeedUsers.sql` file, copy the file to db/setup/, then execute:
```sh
docker compose down -v && docker compose up -d --build -V
```

This will insert 1000 random users to the Users table

Damrod: as an extract of the errors when running the commands you mention I have this:

```
ft_transcendence-db-1                 | psql:/var/lib/postgresql/setup/SeedUsers.sql:987: ERROR:  relation "seedusers" does not exist
ft_transcendence-db-1                 | LINE 1: insert into SeedUsers ("username", "fullName", "email") valu...
```

Damrod: (I could fix it by changing the table name to users in mockaroo, before downloading the file, but maybe this is not the way you intended)

webapp/src/shared/components/DispatchPage/DispatchPage.tsx
```diff
@@ -39,6 +75,11 @@ export default function DispatchPage<T>({
     return array.every((element) => elementChecker(element));
   };
 
+  const handleSearch = (value: string) => {
+    setSearch(value);
```
Damrod, line 79: I'm struggling to understand when do we need useCallback and when we don't. My guess so far, as a rule of thumb, is: if we don't need a dependency array, we don't need usecallback. Also, if we don't need to provide referential stability to the component we are passing the callback to, we don't need usecallback.
Here, I see we don't have nor need a dependency array, but we do pass it to SearchForm, that will keep reredering everytime the component renders, because the callback will be a different object every time.
So I'm not sure. Can you help with this?

webapp/src/shared/components/DispatchPage/DispatchPage.tsx
```diff
@@ -20,7 +30,33 @@ export default function DispatchPage<T>({
   dataMapper,
   button,
 }: DispatchPageProps<T>) {
-  const [data, setData] = useState<T[]>([]);
+  const [search, setSearch] = useState('');
```
Damrod, line 33: since we already have the type, I think something like this may make it a bit cleaner:
```
diff --git a/webapp/src/shared/components/DispatchPage/DispatchPage.tsx b/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
index 30d284d..3eaecf9 100644
--- a/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
+++ b/webapp/src/shared/components/DispatchPage/DispatchPage.tsx
@@ -30,8 +30,11 @@ export default function DispatchPage<T>({
   dataMapper,
   button,
 }: DispatchPageProps<T>) {
-  const [search, setSearch] = useState('');
-  const [offset, setOffset] = useState(0);
+  const [searchParams, setSearchParams] = useState<SearchFetchFnParams>({
+    search: '',
+    offset: 0,
+  });
+  const { search, offset } = searchParams;
   const getData = useCallback(() => {
     return fetchFn({ search, offset });
   }, [fetchFn, search, offset]);
@@ -48,7 +51,7 @@ export default function DispatchPage<T>({
       }
       observer.current = new IntersectionObserver((entries) => {
         if (entries[0].isIntersecting && hasMore) {
-          setOffset(data.length);
+          setSearchParams({ ...searchParams, offset: data.length });
         }
       });
       if (row) {
@@ -76,8 +79,7 @@ export default function DispatchPage<T>({
   };
 
   const handleSearch = (value: string) => {
-    setSearch(value);
-    setOffset(0);
+    setSearchParams({ search: value, offset: 0 });
   };
 
   return (
```

Damrod (approved):
LGTM, thank you!

webapp/src/shared/components/DispatchPage/DispatchPage.tsx
gbudau: You are right. The table name should be `users`. Now it's fixed.

webapp/src/shared/components/DispatchPage/DispatchPage.tsx
gbudau: Thank you. Done

fizcamposancos (approved):
Not many changes, but very condensed. Good job!!!

gbudau:
Thank you! I've opened a new task (#156) to create a script to seed the development database.

PR #155, opened 2022-08-29
#150/edit avatar page 2

Damrod:
The avatar img in avatar page needs to refresh at the right times, needs to update the backend, and needs to stay in sync with the info provided by the backend.
The style needs to match figma.

### DIFF
--- a/webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
+++ b/webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
@@ -19,7 +19,8 @@ import { usersApi } from '../../shared/services/ApiService';
 import { EditableAvatar } from '../../shared/components/Avatar/EditableAvatar';
 import useDrag from '../../shared/hooks/UseDrag';
 
-export const EDITABLE_AVATAR_SCALE = 0.6896;
+export const EDITABLE_AVATAR_SCALE = 1.45;
+export const EDITABLE_AVATAR_SCALE_REVERSE = 1 / EDITABLE_AVATAR_SCALE;
 
 export default function EditAvatarPage() {
   const getCurrentUser = useCallback(
@@ -31,7 +32,7 @@ export default function EditAvatarPage() {
   const { data: user } = useData<User>(getCurrentUser);
 
   const { picturePosition, handleMouseDown, handleMouseMove, handleMouseUp } =
-    useDrag({ x: user?.avatarX ?? 0, y: user?.avatarY ?? 0 });
+    useDrag(user ? { x: user.avatarX, y: user.avatarY } : null);
 
   const navigate = useNavigate();
   const [selectedFile, setSelectedFile] = useState<File | null>(null);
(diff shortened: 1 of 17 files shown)


webapp/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -7,6 +7,7 @@ import { Link } from 'react-router-dom';
 
 export type AvatarProps = {
   url: string;
+  imgHash?: number /* TODO: fix auto reloads when avatar is updated, possibly with this */;
```
Damrod, line 10: When we update the avatar, your browser's cache will probably not update all other instances of avatars in the parent pages, unless you force a reload. This should be improved.

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -11,15 +12,72 @@ import {
...
+    }
+  }, [imgFile]);
+
+  const changeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event &&
+      event?.target &&
+      event?.target?.files?.[0] &&
+      setImgData({
+        imgFile: event?.target?.files[0],
+        imgName: event?.target?.files[0]?.name ?? null,
```
Damrod, line 77: I have to get the falsy types for imgName coherent. I have a small mess now.

webapp/src/shared/components/Avatar/EditableAvatar.tsx
```diff
@@ -54,13 +54,23 @@ export function EditableAvatar({
...
+          />
+        ) : (
+          <img
+            ref={ref}
+            src={url}
+            alt={url}
+            className="editable-avatar__image"
+            style={position}
+          />
+        )}
```
gbudau, line 73:
```suggestion
        <img
          onDragStart={(e) => e.preventDefault()}
          src={`${url}?${imgHash}`}
          alt={url}
          className="editable-avatar__image"
          style={position}
        />
```

suggestion: We could provide an inline `onDragStart` listener because react removes it automatically when the component unmounts and prevents memory leaks.

suggestion: We could change the type of the `imgHash` to string and provide an empty string as a default parameter.

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
gbudau: suggestion: We could do something like this:

```tsx
    if (event.target.files === null) {
      return;
    }

    setImgData({
      imgFile: event.target.files[0],
      imgName: event.target.files[0].name,
      imgHash: Date.now().toString(),
    });
```

gbudau (approved):
Very nice! Thank you! LGTM :rocket:

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
Damrod: done

PR #158, opened 2022-08-31
Editable avatar page and component adjustments

Damrod:
small style fixes for editable avatar (position nudge + pointer style) + removed hardcoded transformation in usedrag to make it reusable + make avatar image and position update when avatar updates position, for all pages


webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -26,7 +29,11 @@ type ImgData = {
 };
 
 export default function EditAvatarPage() {
-  const { user } = useAuth();
+  const getCurrentUser = useCallback(
+    () => usersApi.userControllerGetCurrentUser(),
+    [],
+  );
```
gbudau, line 35: praise: :+1:  Thank you!

suggestion (if-minor): Since it's better to fetch the current user on the pages that needs this data, what do we think about converting the user from `useAuth` to a boolean, something like `isAuthed`/`isLoggedIn`?

Damrod: Seems like a good idea to me.

gbudau (approved):
Thank you! LGTM :rocket:

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
Damrod: done!

PR #159, opened 2022-08-31
#156/Create a script to seed the development database with random data

### LINKED ISSUE #156: Create a script to seed the development database with random data
We need more data in the database to test some functionality of our app while developing.

gbudau:
To run the script, execute the following command from the root directory while the database service is up:

```sh
docker compose exec -it transcendence-app npx knex seed:run
```

close #156


transcendence-app/seeds/seed_users.ts
```diff
@@ -0,0 +1,32 @@
+import { Knex } from 'knex';
+import { faker } from '@faker-js/faker';
+
+const USERS_NUMBER = 5000;
+
+const createRandomUser = () => {
+  const firstName = faker.name.firstName();
+  const lastName = faker.name.lastName();
+  const username = faker.helpers.unique(faker.internet.userName, [
```
Damrod, line 9: Either we make the fullName unique, or we make the username editable. 
From the scale:
```
When logged in, the user has access to his admin panel where he can change informations about himself,
including its nickname (must be unique across the users of the website)
```

gbudau: Good point :+1:

We could make the username editable and save the id from the intranet to check if the user exists or not because currently, we use the username for this. We could add two more columns to the user table: `provider` and `providerId`. This way, we could also add more authentication options(github, google, etc.) in the future.

Damrod (approved):
LGTM!
And it works.
Thank you!

fizcamposancos (approved):
Good job!

gbudau:
Thank you! :tada:

PR #162, opened 2022-09-02
#160/Make usernames editable

### LINKED ISSUE #142: ⚛️ Use of UsersApi service from the generated OpenAPI spec.
Use the UsersApi service from the generated OpenAPI spec.
We could have a status message which shows after the user submits the form (or a Notification component?).

### LINKED ISSUE #160: Make usernames editable
We could save the id from the intranet to check if the user exists or not because currently, we use the username for this. We could add two more columns to the user table: `provider` and `providerId`. This way, we could also add more authentication options(github, google, etc.) in the future.

Subject:
```
A user must be able to choose a unique name that will be displayed on the website
```

Checklist:
```
When logged in, the user has access to his admin panel where he can change information about himself,
including its nickname (must be unique across the users of the website)
```

gbudau:
I've added a new table, `AuthProvider`. According to various StackOverflow posts, this is one of the best ways to implement it because a user could have multiple auth providers (for example, they could link their 42, google, github, etc. accounts). We probably won't implement this link to the various auth providers, but we might add another OAuth provider apart from 42.
https://stackoverflow.com/a/4793464
https://stackoverflow.com/a/7298769

To test these changes, please remove the volumes and rebuild the images first:
```sh
docker compose down -v && docker compose up -d --build -V
```
close #160
close #142


transcendence-app/src/auth/oauth42.strategy.ts
```diff
@@ -62,17 +67,25 @@ export class OAuth42Strategy extends PassportStrategy(Strategy, 'oauth42') {
...
     );
-    if (dbUser) {
-      return dbUser;
+    if (authProvider) {
+      return this.userService.retrieveUserWithId(authProvider.userId);
     }
 
     const avatarDto = await this.saveAvatar(accessToken, avatarUrl);
-    return this.createUser(avatarDto, userDto);
+    const user = await this.createUser(avatarDto, userDto);
```
gbudau, line 81: todo: Since we can have multiple auth providers, there could be a collision of the `username` or the `email`.
We could generate a random `username` (provider `username` + append `providerId` or an arbitrary string?) and allow duplicate emails.

webapp/src/shared/components/Input/EditUserForm.tsx
```diff
@@ -34,7 +38,7 @@ export default function EditUserForm({
     let txt;
     if (
       window.confirm(
-        `Your name: "${fullName}" and your email: "${email}".\n
+        `Your Username: "${username}", Full name: "${fullName}", and your email: "${email}".\n
```
Damrod, line 41: When we try to change our username to match an already existing username, the update will not work, as it should, but the alert will still say "Your data has been saved".
Maybe we should should make some warning message when the username the user tries to set already exists in the database?

webapp/src/shared/components/Input/EditUserForm.tsx
```diff
@@ -53,6 +57,16 @@ export default function EditUserForm({
         onSubmit={handleOnSubmit}
       >
         <div className="inputs-container">
+          <Input
```
Damrod, line 60: comment (non blocking):
We have to rethink the UI.

transcendence-app/src/shared/local-file/infrastructure/db/local-file.entity.ts
```diff
@@ -9,7 +7,7 @@ export enum LocalFileKeys {
   CREATED_AT = '"createdAt"',
 }
 
-export class LocalFileEntity implements BaseEntity {
+export class LocalFileEntity {
```
Damrod, line 10: Why do we have to give up on this particular inheritance hierarchy?

Damrod (approved):
Besides the comments, looks good to me thank you!

webapp/src/shared/components/Input/EditUserForm.tsx
gbudau: Thank you, I'll look into it.

One option is to check if the API call succeeded and set a valid status message or an invalid one.

For example:
Update successfully
The provided info is invalid, missing, or already exists.

If we want more detailed messages, we could check the error.response.status:
400 Bad Request -> Invalid or missing info
422 Unprocessable Entity -> Info already exists

And if we want even more detailed messages, we have to edit the response object and parse the database error message.

webapp/src/shared/components/Input/EditUserForm.tsx
gbudau: If we like how the form looks with this additional input, I could update Figma. If not, please suggest what we can improve.

transcendence-app/src/shared/local-file/infrastructure/db/local-file.entity.ts
gbudau: The `BaseEntity` had one optional field(`id`), which the `AuthProviderEntity` doesn't have. TypeScript was throwing some errors, and the easiest way to fix them was to remove the `BaseEntity` interface.

webapp/src/shared/components/Input/EditUserForm.tsx
Damrod: > 400 Bad Request -> Invalid or missing info
422 Unprocessable Entity

In my opinion this should be enough. Unprocessble entity could be an array of different bad input, which is not very specific (could be duplicated username, or malicious attempt at SQL injection detected, or non displayable characters, or too long data), but I don't think it's crucial to display complete error information here. Maybe I'm wrong.

webapp/src/shared/components/Input/EditUserForm.tsx
Damrod: Please, take a look at the ongoing discussions in figma/slack on the topic, and feel free to share your own ideas on the topic.

webapp/src/shared/components/Input/EditUserForm.tsx
gbudau: I made these changes. Please let me know what we think about this solution.

gbudau:
Thanks!

PR #175, opened 2022-09-09
#167/Fix logout from multiple tabs

### LINKED ISSUE #167: ⚛️ Fix logout from multiple tabs

gbudau:
How it works:

Each application instance creates a `BroadcastChannel` that subscribes to the `auth_channel`
Whenever the user logs out from one of the applications, it sends a message(the currently authenticated user object) to all the other applications subscribed to the `auth_channel`.
Whenever another application receives this message, it compares the users, and if they are the same, it logs out the user.

https://stackoverflow.com/a/58814352

close #167


webapp/src/shared/context/auth-context.tsx
```diff
@@ -28,6 +28,16 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
   }, []);
   const { data, isLoading: isDataLoading } = useData<User>(getCurrentUser);
   const navigate = useNavigate();
+  const authBroadcastChannel = useMemo(
+    () => new BroadcastChannel('auth_channel'),
+    [],
+  );
```
vdedios, line 34: Is this available to any webapp instance running on the same browser?

gbudau: Yes, each application running in a different tab or window subscribes to the `auth_channel` -> https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel#constructor

And they can send messages to each other.

fizcamposancos (approved):
Simple and complex at the time.
LGTM!!

gbudau:
Thanks!

PR #177, opened 2022-09-10
#172/create new login page

### LINKED ISSUE #172: ⚛️ Create new Login Page

fizcamposancos:
Closes #172
The design is not closed yet, so please comment, here or in figma, your preferences.


Damrod:
Could we force a gap in the gap annotation, please? AFAIK in figma it's 50 px, which could be a percent of 100vh of your chosing (5% ? 10% of 100vh ? ). Also, the page title (pong in 'playing' color), and the buttons should be aligned to the left.

Additionally, I'd say the color of the link is off. I'd say it's getting decorated by default. This line `text-decoration: none;` in the appropriate block in the style sheet may help.

login_page

webapp/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+    type: 'pending',
+    message: '',
+  });
+
+  async function login() {
+    try {
+      await authApi.authControllerLoginLocalUser({
+        loginUserDto: { username, password },
+      });
+      window.location.href = USERS_URL;
```
Damrod, line 37: what was the problem with navigate?

webapp/src/pages/LoginPage/LoginPage.css
```diff
@@ -0,0 +1,81 @@
...
+
+.login .landing-animation {
+  display: none;
+}
+
+/* Styles for viewports wider than 768 pixels. */
+@media (min-width: 768px) {
+  .login {
+    min-height: 20rem;
+    padding: 4rem var(--padding-sides) 8rem var(--padding-sides);
```
Damrod, line 63: maybe an `@import` between this css and the landing css would prevent the whole block of duplicated code we have?

webapp/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+          setStatus({ type: 'error', message: `${error.response.statusText}` });
+        }
+      } else if (error instanceof Error) {
+        setStatus({ type: 'error', message: `${error.message}` });
+      }
+    }
+  }
+  const handleOnSubmit = (e: React.FormEvent<HTMLFormElement>): void => {
+    e.preventDefault();
+    login();
```
Damrod, line 55: > Promise returned from `login()` is ignored 
What about we put a `.catch((e) => console.error(e))` attached to this `login()`?

webapp/src/pages/LoginPage/LoginPage.tsx
fizcamposancos: For some reason it navigates to the landing page and not to the users page, and if you refresh then it do goes to the users page

webapp/src/pages/LoginPage/LoginPage.tsx
fizcamposancos: Yes, you are right. Done

fizcamposancos:
> Could we force a gap between the title `PONG` and the button `Login with 42`, please? AFAIK in figma it's 50 px, which could be a percent of 100vh of your chosing (5% ? 10% of 100vh ? some rem? Id rather use the percent of 100vh, but it's your choice). Also, the page title (pong in 'playing' color), and the buttons should be aligned to the left (I think the title `PONG` is not in the right column, but I'm not sure).
> 
> Additionally, I'd say the color of the link is off. I'd say it's getting decorated by default, but, again, I'm not sure. This line `text-decoration: none;` in the appropriate block in the style sheet may help.
> 
> Attached image to illustrate my comments:
> 
> login_page

Regarding the gap between the title and the first button, I put 4rem, but if you think another value could fit better, please go ahead, I have no problem changing it.
Buttons and title are now aligned to the left. I followed the main landing page for the left column, and I cannot say if it is in the right position or not, I am not sure. In Figma there are two main images, and they fit with the page, but with intermediate widths…
The link color is on “porpoise” jeje, I set the color to `TextColor.ONLINE` on line 135. The `text-decoration: none;` is to remove the default underline of the links. But feel free to change it if you don't like it.

Damrod:
we're still able to navigate to the /login page, even after loggin in, either by pressing the back button, or by writing the url in the browser url bar. I assume this is not desired behaviour, but it's non blocking for this PR.

Damrod (approved):
You can merge if you want, but please, open tickets to recall that we have to fix the navigation back to the login page, and the non encrypted password we can plainly see in the request body upon login in.

Damrod:
login_scroll
I also have a problem with the scroll. The screen is tall enough to fit in one display the whole page, but there is still a scrollbar on the right. I think we should do something like `scroll: clip`, or make the height ot the page be specifically and explicitly 100vh to prevent this.

webapp/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -0,0 +1,142 @@
...
+            }}
+          />
+          <Input
+            variant={InputVariant.LIGHT}
+            placeholder="password"
+            value={password}
+            name="password"
+            type="password"
+            onChange={(e) => {
+              setPassword(e.target.value);
```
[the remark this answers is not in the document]
vdedios: Despite being scary, is not that problematic if using and encrypted connection such as HTTPS. All data will be encrypted and you wouldn't have to have any issues. This is a widely extended practise, if you fill Linkedin login form, and analyze the payload request, you'll see:
image

We could mitigate sending plain text pass by [hashing/encrypting them ](https://ticki.github.io/blog/you-are-probably-doing-login-systems-wrong/) but the attacker could still use the hashed pass to gain access into our system. And of course there will be more options that I don't know about.

All that being said, I admit it stills feels a little bit scary :cry:

vdedios:
Agree with @Damrod, there seems to be, still, some inconsistencies in the layout that need to be checked. In the mobile version (I have selected a Pixel 5 model in the browser):

1. In the landing page, the title is placed to the left while the buttons are centered. Compare  it with previous login version, the buttons are supossed to fill the width of the grid layout (same width as title) and everything is centered. 

2. In the login page, the content is aligned to the left. There's a larger padding on the right.

webapp/src/pages/LoginPage/LoginPage.tsx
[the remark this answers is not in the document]
fizcamposancos: I've opened a ticket about it so we can investigate and decide what to do.

PR #178, opened 2022-09-11
Fix landing pong animation

gbudau:
There was some space between the bottom paddle and the ball.


fizcamposancos (approved):
Now it seems to be perfect!!

PR #180, opened 2022-09-12
#179/create register page

### LINKED ISSUE #179: ⚛️ Create Register Page

fizcamposancos:
closes #179

With the design issues we are discussing in the login page #177


webapp/src/pages/RegisterPage/RegisterPage.tsx
```diff
@@ -0,0 +1,161 @@
...
+  TextWeight,
+} from '../../shared/components';
+import { ResponseError } from '../../shared/generated';
+import { authApi } from '../../shared/services/ApiService';
+import { SubmitStatus } from '../../shared/types';
+import { LOGIN_OPTIONS_URL } from '../../shared/urls';
+import './RegisterPage.css';
+
+export default function RegisterPage() {
+  const [username, setUsername] = useState('');
```
Damrod, line 20: couldn't we use something like this:
```
const [userRegistrationDto, setUserRegistrationDto] = useState<RegisterUserDto>({username: '', fullName: '', email: '', password: '', confirmationPassword: ''});
```
instead of these some 5 useState, since we already have the type anyway?

webapp/src/pages/RegisterPage/RegisterPage.css
```diff
@@ -0,0 +1,72 @@
...
+    grid-area: main;
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem var(--column-gap);
+  }
+
+  .register .landing-animation {
+    display: inline;
+  }
+}
```
Damrod, line 72: I have a problem, again, with the height of the page. I work on a small screen, and have to do quite some scrolling even in a page with not so much content as this one, but maybe there's no clean way around my scrolling...
Where even is the figma for this particular page? Did you wing it?

fizcamposancos: Yes, the ticket is open, once I fix the other one I will fix this as well. BTW I couldn't do what you suggested, `scroll: clip `, what did you mean?

webapp/src/pages/RegisterPage/RegisterPage.css
```diff
@@ -0,0 +1,72 @@
...
+  margin: 0;
+  padding: var(--padding-top-bottom) var(--padding-sides);
+  height: 100vh;
+  min-height: 40rem;
+  overflow: scroll;
+  background-color: var(--color-background);
+  color: var(--color-light);
+  display: grid;
+  grid-template-columns: repeat(var(--column-number), 1fr);
+  grid-template-rows: auto 1fr;
```
Damrod, line 11: have you noticed that we have been building all pages with a column gap that you are nor having not here not in the login page? V.g:
Screenshot from 2022-09-15 23-36-38
Screenshot from 2022-09-15 23-35-38

fizcamposancos: No, sorry, I'll fix it

Damrod: I'm sorry, I made a mess with your pr. I wanted to delete the screenshots I posted, and instead I removed the css file. Should be fixed now (I forced pushed your commit from before I created my stupid commit)

webapp/src/pages/RegisterPage/RegisterPage.css
Damrod: I'm not sure. It was pseudo code, not really code. I will try to find a solution for this tomorrow.

webapp/src/pages/RegisterPage/RegisterPage.tsx
```diff
@@ -0,0 +1,161 @@
...
+          setStatus({ type: 'error', message: `${error.response.statusText}` });
+        }
+      } else if (error instanceof Error) {
+        setStatus({ type: 'error', message: `${error.message}` });
+      }
+    }
+  }
+  const handleOnSubmit = (e: React.FormEvent<HTMLFormElement>): void => {
+    e.preventDefault();
+    register();
```
Damrod, line 70:
```
Promise returned from register is ignored 
```
couldn't we attach a `.catch((e)=>console.error(e));` to `register()` ?

webapp/src/pages/RegisterPage/RegisterPage.css
fizcamposancos: Don't worry. But I don't know why it's happening. That gap is in the root.css, isn't it? Shouldn't it be there by deafult?

webapp/src/pages/RegisterPage/RegisterPage.tsx
fizcamposancos: Yes, I promise I thought I did it, maybe I didn't push. I add it

webapp/src/pages/RegisterPage/RegisterPage.css
fizcamposancos: Done

webapp/src/pages/RegisterPage/RegisterPage.tsx
fizcamposancos: I've finally used `useReducer` instead. [https://reactjs.org/docs/hooks-reference.html#usereducer](https://reactjs.org/docs/hooks-reference.html#usereducer)

Damrod:
Now below the title there is a margin + a gap which upon being visualized look exactly the same. Either make the gap between the rows larger and remove the margin, or make the gap between rows none, and let the margin stay the way it is.

Damrod:
what about we attempt to use the grid template columns to set the width of the different components that go from some column to a different one? Cf. [here](https://github.com/42AGV/ft_transcendence/pull/188#issuecomment-1249930590)

webapp/src/pages/RegisterPage/RegisterPage.tsx
Damrod: To be honest I think this is not the _by the book_ ideal use case for useReducer (a reducer becomes a real help, and almost a necessity, when we have to handle complex state logic. Here we don't have complex state logic: we're only updating the dto).

My idea was something like what you can find here `webapp/src/shared/hooks/UseDrag.tsx` L21.

But it works, so I'll just approve it.

fizcamposancos:
> Now below the title there is a margin + a gap which upon being visualized look exactly the same. Either make the gap between the rows larger and remove the margin, or make the gap between rows none, and let the margin stay the way it is.

I removed the margin and modified the gap. Please, take a look at it.

vdedios:
Nice! but I think we should review feedback messages. Right now it seems we have the same "invalid email or passwords" message for all these situations:

- If passwords are different
- If same specific mandatory field is missing

fizcamposancos:
> Nice! but I think we should review feedback messages. Right now it seems that we have the same "invalid email or passwords" message for all these situations:
> 
>     * If passwords are different
> 
>     * If same specific mandatory field is missing

I did some extra input validation, but manually. If you prefer to use something like useForm, please let me know and I will change it. Right now I think it cover most of the possible situations

webapp/src/pages/RegisterPage/RegisterPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+          message: 'Password can not be empty',
+        });
+      } else if (
+        userRegisterUserDto.password !==
+        userRegisterUserDto.confirmationPassword
+      ) {
+        setStatus({
+          type: 'error',
+          message: 'Passwords are different',
+        });
```
gbudau, line 79: suggestion: We could extract this part into a function and validate the form values on the client side before making the API call.

Something like this:
```tsx
async function register() {
  if (!hasValidFormValues()) {
    return;
  }
  try {
    ...
  } catch (error) {
    ...
  }
}
```

gbudau (approved):
Nicely done, thank you! LGTM

webapp/src/pages/RegisterPage/RegisterPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+    password: '',
+    confirmationPassword: '',
+  };
+  const [userRegisterUserDto, updateUserRegisterUserDto] = useReducer(
+    (userRegisterUserDto: RegisterUserDto, updateUserRegisterUserDto: any) => ({
+      ...userRegisterUserDto,
+      ...updateUserRegisterUserDto,
+    }),
+    initialuserRegisterUserDto,
+  );
```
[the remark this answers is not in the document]
Damrod: I already mentioned this [here](https://github.com/42AGV/ft_transcendence/pull/180#discussion_r973507804)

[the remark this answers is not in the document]
Damrod: An [article](https://blog.thoughtspile.tech/2021/10/11/usestate-object-vs-multiple/) on the topic of using multiple state variables vs single state object.

> 
useState calls are more expensive than object allocations, so for 3+ items useState(object) wins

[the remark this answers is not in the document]
gbudau: @Damrod, thanks! It was a good read.

webapp/src/pages/RegisterPage/RegisterPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+        type: 'error',
+        message: 'Passwords are different',
+      });
+      return false;
+    }
+    return true;
+  }
+  const navigate = useNavigate();
+  async function register() {
+    if (hasValidFormValues()) {
```
gbudau, line 76:
```suggestion
    if (!hasValidFormValues()) {
```

fizcamposancos: Sorry, thanks!

PR #190, opened 2022-09-16
Create chat bubble component

### LINKED ISSUE #186: ⚛️ Create Chat Bubble component

fizcamposancos:
Chat Bubble based on the Figma design, and the new proposal with name
Closes #186


Damrod:
I think the name of the author should be in a distinct style.
Screenshot from 2022-09-17 03-02-14

webapp/src/shared/components/ChatBubble/ChatBubble.css
```diff
@@ -0,0 +1,31 @@
...
+  border-radius: 8px;
+  padding: 0.625rem;
+}
+
+.chat-bubble-text > p {
+  margin: 0;
+}
+
+.chat-bubble-text-self {
+  background-color: #595959;
```
Damrod, line 24: I would give names to these colors (another variable with yet another name, in .root)

fizcamposancos: > I think the name of the author should be in a distinct style, or even outside of the bubble. Screenshot from 2022-09-17 03-02-14 like this: chat_bubble_component_author
> 
> But maybe it's too bothersome. What do you people think?
> 
> EDIT: Mostly because otherwise there is no way to validate that a user actually sent a message. They may impersonate other users.

I followed the groups in whatsapp, but without changing the color of the name. Whatsapp assigns a color to every member name of the group. And I don't mind changing the style or putting the name outside the bubble.

But I can not understand how someone can impersonate other users if the name it put by the chat, not by the user.

fizcamposancos: I've change it using the color names we already have, `--color-dark` and `--color-light-variant`. The second one is not exactly the same gray but for me it's ok. Please have a look.

fizcamposancos:
> I think the name of the author should be in a distinct style, or even outside of the bubble. Screenshot from 2022-09-17 03-02-14 like this: chat_bubble_component_author
> 
> But maybe it's too bothersome. What do you people think?
> 
> EDIT: Mostly because otherwise there is no way to validate that a user actually sent a message. They may impersonate other users.

I followed the groups in whatsapp, but without changing the color of the name. Whatsapp assigns a color to every member name of the group. And I don't mind changing the style or putting the name outside the bubble.

But I can not understand how someone can impersonate other users if the name it put by the chat, not by the user.

vdedios:
I agree with damrod, I think its simpler to just display the title above the message. We could mimic whatsapp but that would imply calculating colors depending on how many users we have

webapp/src/shared/components/ChatBubble/ChatBubble.css
vdedios: I think we should stick to the color of the [palette](https://www.figma.com/file/0es84Oh1roPphFOd2GYysn/Library) we have, so agree with using already defined vars.

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
```diff
@@ -0,0 +1,44 @@
+import { AvatarProps, SmallAvatar } from '../Avatar/Avatar';
+import Text, { TextColor, TextVariant, TextWeight } from '../Text/Text';
+import './ChatBubble.css';
+
+export enum ChatBubbleVariant {
+  SELF = 'self',
+  OTHER = 'other',
```
vdedios, line 7: Nice!!

vdedios (approved):
Despite the issue with user name, LGTM!

webapp/src/shared/components/ChatBubble/ChatBubble.css
vdedios: I'd say to use purple color for "SELF" messages. It's vivid and extends "funky" style of the app. It makes good contrast. I've been looking in some chat apps and they same to share the same principle of using one of the most representative colors for self messages:
Screen Shot 2022-09-18 at 7 16 48 PM

vdedios:
I'd make two more very simple changes:
1. Align the bubble name with the bubble itself but not the container:

```css
.chat-bubble-text-author {
  margin-left: 0.625rem;
}
```

2. Remove the two dots (`:`) in the name since it's self descriptive that the name represents what comes bellow it.

Both of them would look like:
image

gbudau (approved):
Thank you! LGTM :rocket:

PR #196, opened 2022-10-04
Chat POC

### LINKED ISSUE #195: develop MVP for a chat like feature
We decided we would try separately to develop a chat MVP to get acquainted with websockets and other technologies.

gbudau:
Proof of concept for a chat room

Done:
WebSocket authentication with sessions
Chat room `/chatroom/:chatRoomId` routes

To do:
Persist the data into the database

close #195


webapp/src/pages/ChatRoom/Messages.tsx
```diff
@@ -0,0 +1,93 @@
+import { useEffect, useState } from 'react';
+import { ChatBubble, ChatBubbleVariant } from '../../shared/components';
+import { User } from '../../shared/generated';
+import { useAuth } from '../../shared/hooks/UseAuth';
+import socket from '../../shared/socket';
+import { AVATAR_EP_URL, WILDCARD_AVATAR_URL } from '../../shared/urls';
+
+type MessageType = {
```
Damrod, line 8: this could come from the generated files, isn't it?

webapp/src/pages/ChatRoom/Messages.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { user: me } = useAuth();
+
+  useEffect(() => {
+    const messagesListener = (messages: MessageType[]) => {
+      setMessages(messages);
+    };
+
+    const messageListener = (message: MessageType) => {
+      setMessages((prevMessages) => {
+        const newMessages = [...prevMessages, message];
```
Damrod, line 29: I understand this is somehow a draft, because we are not persisting the messages ye4t, but I assume when we do, this whole logic to display older messages will become simpler?

webapp/src/pages/ChatRoom/Messages.tsx
```diff
@@ -0,0 +1,93 @@
...
+              ? ChatBubbleVariant.SELF
+              : ChatBubbleVariant.OTHER
+          }
+          name={message.user.username}
+          text={message.content}
+          avatar={{
+            url: message.user.avatarId
+              ? `${AVATAR_EP_URL}/${message.user.avatarId}`
+              : WILDCARD_AVATAR_URL,
+            status: connectedUsers.has(message.user.id) ? 'online' : 'offline',
```
Damrod, line 85: are we interested here in displaying whether or not a particular user is generally connected to our website, as in logged in, or whether he/she's currently browsing a particular chat room?

webapp/src/index.tsx
```diff
@@ -28,6 +28,7 @@ import {
 } from './shared/urls';
 import { AuthProvider } from './shared/context/auth-context';
 import RequireAuth from './shared/components/RequireAuth/RequireAuth';
+import ChatRoom from './pages/ChatRoom/ChatRoom';
```
Damrod, line 31: in line 16, better, I think. We should add it to pages

transcendence-app/src/setup-app.ts
```diff
@@ -0,0 +1,37 @@
+import { INestApplication, ValidationPipe } from '@nestjs/common';
+import * as session from 'express-session';
+import * as ConnectMemcached from 'connect-memcached';
+import * as passport from 'passport';
+import { ConfigService } from '@nestjs/config';
+import { EnvironmentVariables } from './config/env.validation';
+import { WsSessionAdapter } from './chat/ws-session.adapter';
+
+export const setupApp = (app: INestApplication) => {
```
Damrod, line 9: I don't get much how/why can we now afford to not initialize the app from `app.module.ts`

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -0,0 +1,121 @@
...
+@UseGuards(WsAuthenticatedGuard)
+@UseInterceptors(ClassSerializerInterceptor)
+export class ChatGateway
+  implements OnGatewayInit, OnGatewayConnection, OnGatewayDisconnect
+{
+  @WebSocketServer() server!: Server;
+
+  logger = new Logger(ChatGateway.name);
+  messages = new Set<Message>();
+  connectedUsers = new Map<string, User>();
```
Damrod, line 40: I don't understand why do we need this to have both the userId, and the User (that already contains the user id)

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  async handleDisconnect(client: Socket) {
+    const user = client.request.user;
+    if (user) {
+      const matchingSockets = await this.server.in(user.id).allSockets();
+      const isDisconnectedAll = matchingSockets.size === 0;
+      if (isDisconnectedAll) {
+        this.connectedUsers.delete(user.id);
```
Damrod, line 78: I guess it's because it makes this simpler?

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
```diff
@@ -23,7 +23,9 @@ export default function ChatBubble({
   return (
     <div className={`chat-bubble chat-bubble-${variant}`}>
       {avatar && <SmallAvatar {...avatar} />}
-      <div className={'chat-bubble-text-container'}>
+      <div
+        className={`chat-bubble-text-container chat-bubble-text-container-${variant}`}
+      >
         <div className={'chat-bubble-text-author'}>
```
Damrod, line 29: I think we should add variants to this chat bubble such as 'is-consecutive' so that we can maybe not show the author nickname when there are a series of bubbles that are all from the same author, or not show the avatar if it's not the last one in a sequence of such bubbles.

Damrod:
I left a series of comments that don't really need to be addressed here and now: it's just a stream of conciousness of ideas I want to bring up, and don't want to forget. This seemed like a good place to do so, since we're at POC level.

webapp/src/pages/ChatRoom/Messages.tsx
gbudau: It could be if we create at least one endpoint that returns this type of data and document it using Swagger.

webapp/src/pages/ChatRoom/Messages.tsx
```diff
@@ -0,0 +1,93 @@
...
+  content: string;
+  createdAt: number;
+};
+
+function Messages() {
+  const [messages, setMessages] = useState<MessageType[]>([]);
+  const [connectedUsers, setConnectedUsers] = useState<Map<string, User>>(
+    new Map<string, User>(),
+  );
+  const { user: me } = useAuth();
```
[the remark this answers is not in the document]
gbudau: Here we use only the user ID to either display the SELF or OTHER `ChatBubble` variant, and using the info from the `AuthContext` saves us an extra fetch call.

webapp/src/index.tsx
gbudau: Thanks! Done.

transcendence-app/src/setup-app.ts
gbudau: We need to use the same session middleware for the HTTP and the WebSocket server (the `app.useWebSocketAdapter(new WsSessionAdapter(app, sessionMiddleware));` line).

It's the same thing as before, but we set up the middlewares in this function and call it from the main instead of setting them up at the module level.

transcendence-app/src/chat/chat.gateway.ts
gbudau: Good point, thanks! I've changed this to a `Set` of user IDs.

webapp/src/pages/ChatRoom/Messages.tsx
gbudau: Here we use an [updater function](https://beta.reactjs.org/apis/react/useState#examples-updater) to skip adding the `messages` as a dependency of the `useEffect` hook.

When we add persistence, this will stay mostly the same because we want to display the previous messages and add the new message (if we figure out another way, we can change it, of course).

Damrod: What about users that join the chat later on? Or users that join, leave, and then rejoin? Shouldn't they be able to see all chats since the beginning? Wouldn't this have to persist in the backend?

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
gbudau: Thanks! I like this :+1:

I've made this change. Please let me know what we think or if there is any way to improve it.

webapp/src/pages/ChatRoom/Messages.tsx
Damrod: All of them, yes.

webapp/src/pages/ChatRoom/MessageInput.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const submitForm = (e: React.FormEvent) => {
+    e.preventDefault();
+    socket.emit('message', value);
+    setValue('');
+  };
+
+  return (
+    <form className="message-input-form" onSubmit={submitForm}>
+      <Input
+        variant={InputVariant.DARK}
```
Damrod, line 17: What do we think about Making this input component have the text wrap and have the height of the input field increase at each new line appear? I mean something like on desktop facebook.
Something like [this](https://stackoverflow.com/questions/69562816/wrapping-and-height-increasing-input-react-css) suggests

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
Damrod: I will look at it tomorrow on my workstation (I'm on my phone more). Thank you!

webapp/src/pages/ChatRoom/MessageInput.tsx
fizcamposancos: I like it, with maximun of "x" lines, and then scroll. We can try it

gbudau: I like it too :+1:

Does this mean we should also add a button to submit the message since pressing <kbd>Enter</kbd> will add a new line?

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
Damrod: What about this
```
diff --git a/webapp/src/pages/ChatRoom/Messages.tsx b/webapp/src/pages/ChatRoom/Messages.tsx
index e478db1..c6e6c47 100644
--- a/webapp/src/pages/ChatRoom/Messages.tsx
+++ b/webapp/src/pages/ChatRoom/Messages.tsx
@@ -45,6 +45,10 @@ function Messages() {
         const isConsecutive =
           index !== messages.length - 1 &&
           messages[index].user.id === messages[index + 1].user.id;
+        const isFirst =
+          index === 0 ||
+          messages[index].user.id !== messages[index - 1].user.id;
+        console.log(isFirst);
         return (
           <li key={message.id} className="messages-list-item">
             <ChatBubble
@@ -61,6 +65,7 @@ function Messages() {
                   : WILDCARD_AVATAR_URL,
               }}
               isConsecutive={isConsecutive}
+              isFirst={isFirst}
             />
           </li>
         );
diff --git a/webapp/src/shared/components/ChatBubble/ChatBubble.tsx b/webapp/src/shared/components/ChatBubble/ChatBubble.tsx
index 407807c..d0b01a3 100644
--- a/webapp/src/shared/components/ChatBubble/ChatBubble.tsx
+++ b/webapp/src/shared/components/ChatBubble/ChatBubble.tsx
@@ -13,6 +13,7 @@ type ChatBubbleProps = {
   text: string;
   name?: string;
   isConsecutive?: boolean;
+  isFirst?: boolean;
 };
 
 export default function ChatBubble({
@@ -21,6 +22,7 @@ export default function ChatBubble({
   text,
   name,
   isConsecutive = false,
+  isFirst = true,
 }: ChatBubbleProps) {
   return (
     <div className={`chat-bubble chat-bubble-${variant}`}>
@@ -31,7 +33,7 @@ export default function ChatBubble({
         className={`chat-bubble-text-container chat-bubble-text-container-${variant}`}
       >
         <div className={'chat-bubble-text-author'}>
-          {name && !isConsecutive && (
+          {name && isFirst && (
             <Text
               variant={TextVariant.CAPTION}
               color={TextColor.LIGHT}
```

?

gbudau: LGTM :+1:

Damrod:
I think this alignment is also kind of weird, but maybe it's just my opinion
chat_bubble_component_alignment

gbudau:
> I think this alignment is also kind of weird, but maybe it's just my opinion

Good catch! I think it's fixed now.

webapp/src/pages/ChatRoom/MessageInput.tsx
gbudau: I added this change. Please take a look, and if you have any suggestions, please let me know.

Damrod: Both in Firefox and in Chrome, I'm seeing that the stuff I paste in the input field is being pasted with format (??)
Screenshot from 2022-10-07 12-24-25
Screenshot from 2022-10-07 12-24-10
Screenshot from 2022-10-07 12-23-42
???

gbudau: Good catch! Thanks!

I'm looking into how we could fix this, and so far, I didn't find any viable solution.

gbudau: I think it's fixed now. I've added the trick from this article -> https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/

fizcamposancos:
I guess this chatroom would be used when we enter a chat from the chats page, so the chatName, chatId or an identifier should be passed as prop to send messages to the users in that chat, right?

gbudau:
Good point :+1:
We could either pass an id as a prop or use URL parameters, something like:
`/chat/:roomId`
or
`/chat/room/:roomId`
`/chat/user/:userId`

webapp/src/pages/ChatRoom/MessageInput.css
```diff
@@ -0,0 +1,72 @@
...
+
+.message-input-grow-wrap::after,
+.message-input-text {
+  /* Identical styling required!! */
+  outline: none;
+  border: none;
+  margin: 0;
+  min-height: 3rem;
+  max-height: 12rem;
+  word-break: break-all;
```
Damrod, line 47: why do we have here `break-all` ?
It breaks lines in the middle of words. I have never seen any text field where this is normal or desired behaviour.

gbudau: I've added this before when we had a `contenteditable` div because very long words broke the layout.
I see it's not the case anymore, I will remove it :+1:

gbudau:
Marking this as ready, do we want to merge it or use it as a reference? The authentication is 100% complete, and the `ChatGateway` and `ChatRoom` need minor changes once we implement database persistence.
I think @fizcamposancos is working on that (schemas, repositories, services, and controllers for the chat and messages)

webapp/src/pages/ChatRoom/MessageInput.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useRef, useState } from 'react';
```
Damrod, line 1: Don't you think maybe this MessageInput could be a component?

gbudau: What do we think about creating a components directory inside the `ChatRoom` directory? Because these two components are specific to this page. Something like:

```
ChatRoom.tsx
ChatRoom.css
components/
    index.ts
    Messages/
        Messages.tsx
        Messages.css
    MessageInput/
        MessageInput.tsx
        MessageInput.css
```

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -0,0 +1,139 @@
...
+      client.disconnect();
+    } else {
+      this.connectedUsers.add(user.id);
+      const sessionId = client.request.session.id;
+      // Join the session ID room to keep track of all the clients linked to this session ID
+      client.join(sessionId);
+      // Join the user ID room to keep track of all the connected clients
+      // (one user could connect from multiple private tabs or browsers with different session IDs)
+      client.join(user.id);
+      const matchingSockets = await this.server.in(user.id).allSockets();
```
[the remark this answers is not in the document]
gbudau: Sorry. What can we change to improve this?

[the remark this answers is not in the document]
Damrod: Well, generally, speaking about a variable, I'd say if it's initialized from an assignment from a literal, or it's a new, it doesn't need type. If it's assigned from object and array destructuring don't need types.
If we need to research to find the function signature we're inferring the type from, then we'd better declare the type explicitly.
That would be my policy, but maybe others have a different opinion

webapp/src/pages/ChatRoom/MessageInput.tsx
Damrod: The message input seems fairly useful to me in different pages, not just here. But we don't have an use case just yet, so maybe it's okay.

Besides that, I don't have a strong opinion on the topic. Maybe the others have opinions.

transcendence-app/src/chat/chat.gateway.ts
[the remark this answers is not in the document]
gbudau: Thanks, I'll keep it in mind. I've added types to this variable :+1:

Damrod (approved):
lgtm!

webapp/src/pages/ChatRoom/components/MessageInput/MessageInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+import './MessageInput.css';
+
+type MessageInputProps = {
+  to: string;
+};
+
+const MessageInput = ({ to }: MessageInputProps) => {
+  const [value, setValue] = useState('');
+  const divRef = useRef<HTMLDivElement>(null);
+
```
fizcamposancos, line 13: Could you explain me why `useRef `and not only `useState`, and what is this `dataset.replicatedValue`?

fizcamposancos (approved):
Great job!!

webapp/src/pages/ChatRoom/components/MessageInput/MessageInput.tsx
gbudau: You are right :+1:. We could also use only the `useState`. I've changed it.

The replicated value is a [data attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*), we use it to change the content of the div using CSS.

This component has a textarea, which cannot expand its height, and a div, which can do it based on its content.

They share the same grid position, and the height expands depending on which element takes more space.

Whenever the textarea input changes, we copy the value to the data replicated value attribute of the div and assign the div content to this data attribute in CSS.

gbudau:
I did some refactoring:

I've split the `ChatModule` into:

`ChatRoomModule` -> handle chat room events.

`EventsModule` -> handles web socket events, like when a socket connects or disconnects.

Updated the chat room route to `/chatroom/:chatRoomId`

Renamed the `Messages` and `MessageInput` to `ChatRoomMessages` and `ChatRoomMessageInput`.

Please let me know what you think or if there is something we can improve. Thanks!

vdedios (approved):
As said, this looks very good to me! I’ll keep my comments and improvements until I get a little bit more context of some things 🚀

gbudau:
Thanks! :heart:

PR #198, opened 2022-10-11
#197/webapp improvements

### LINKED ISSUE #197: Fixes and improvements for MVP 0.1
This task is about reviewing some detected issues as well as making some improvements in our mvp:

- **Improve search**
  - use a single search context so we don't have to pass states and callbacks around and we centralize everything that matters in a search: query and results
  - fix visual "jumps" each time we update the seach query, try to make results appear smoother.
  - make rows and header fully scrollable so we have more space in a smartphone screen.

- **Move dispatch page into a template:**
  - I read about [atoms, molecules, templates and pages](https://blog.kamathrohan.com/atomic-design-methodology-for-building-design-systems-f912cf714f53) and loved the idea! 
  - The idea is very similiar to the dispatch page but, to convert it into a more reusable template so we can easily use it in chats, games etc. Its main responsibility is to draw data but does not care about how it get it, that should be a responsibility of th page instantiating it.
  - usersPage now holds the responsability of handling data and pass it to templates. The same idea can be easily replicated with other types of rows pages like chats, games, groups etc.

- **Improve UX of editAvatar page and component:**
  -  Make editableAvatar work in touch screens
  -  Add feedback messages to know an action was successful
  - Make selected avatar appear and disabling editing while uploading.

vdedios:
closes #197 

:warning:  **(Read the issue so you get the full context of what has been made)** :warning: 

Since many changes where related I decided to include everything in a single PR. Play around with it and if you need help don't hesitate to ask :)


webapp/src/shared/components/Row/Row.tsx
```diff
@@ -113,3 +119,11 @@ export default function Row({
     )) || <div className={`row paragraph-regular`}>{RowChildren}</div>
   );
 }
+
```
Damrod, line 122: ??

vdedios: leftovers :eyes:

webapp/src/shared/components/Input/Input.tsx
```diff
@@ -46,6 +46,7 @@ export default function Input({
           onChange={onChange}
           name={name}
           type={type}
+          autoComplete="off"
```
Damrod, line 49: I saw chrome complaining about the password not having auto complete off, but generally, if not for passwords, I don't think it's a problem to have auto complete on. Why do you disagree?

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.css
```diff
@@ -1,7 +1,8 @@
-.dispatch-page {
```
Damrod, line 1: If you change the name and the folder of the jsx, maybe you should also change the name and folder of the css?

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
```diff
@@ -0,0 +1,82 @@
...
+import SearchForm from '../../Input/SearchForm';
+import RowsList, { RowItem } from '../../RowsList/RowsList';
+import NavigationBar from '../../NavigationBar/NavigationBar';
+import { useData } from '../../../hooks/UseData';
+import Loading from '../../Loading/Loading';
+import { usersApi } from '../../../services/ApiService';
+import { useSearchContext } from '../../../context/SearchContext';
+
+type RowsPageTemplateProps<T> = {
+  dataValidator: (data: T) => boolean;
```
Damrod, line 15: The data validator is there to look at whether the type of the object is T. I don't think it makes sense to have the data validator take T as an argument. It's a runtime check.

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.css
vdedios: I think the name of the folder is already changed. Maybe you meant the name of the base class to `rows-page-template`?

webapp/src/shared/components/Input/Input.tsx
vdedios: Yes, passwords was the main reason... and also, for me, usually is a bit annoying to have the autocomplete display all the crap you've already typed before :sweat_smile:

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { result, fetchMoreResults } = useSearchContext();
+
+  const data = ((array: object): RowItem[] | null => {
+    let rows = [];
+
+    if (!Array.isArray(array)) {
+      return null;
+    }
+    for (let i = 0; i < array.length; i++) {
+      if (!dataValidator(array[i])) {
```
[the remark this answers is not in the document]
vdedios: agree, is sadly less readable to use a for loop than a map but, since we are sanitizing arrays with no limits in length, I think we gain a bit in performance saving an iteration. A web application could run on the latest iphone or on a 8 years old phone, depending on the device that could be expensive

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.css
Damrod: Ah, I didn't notice the name change. No problem then.

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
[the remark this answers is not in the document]
Damrod: I have to say that we added the type validator before we had properly generated Api files. We didn't have proper type checking from the request methods. I think now the generated fetching methods provide runtime type checking. Maybe it's obsolete, and we don't really need this here at all anymore.

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
```diff
@@ -0,0 +1,82 @@
...
+import NavigationBar from '../../NavigationBar/NavigationBar';
+import { useData } from '../../../hooks/UseData';
+import Loading from '../../Loading/Loading';
+import { usersApi } from '../../../services/ApiService';
+import { useSearchContext } from '../../../context/SearchContext';
+
+type RowsPageTemplateProps<T> = {
+  dataValidator: (data: object) => boolean;
+  dataMapper: (data: T) => RowItem;
+};
```
Damrod, line 17: I added the `button ?: JSX.Element` to what we used to call DispatchPage Props because even back then we could already see that we may need a button below the list of rows, maybe not for the users page, but certainly for the chat. Why did you decide to remove it, now that it looks like we may actually need this feature?

webapp/src/shared/components/Row/Row.tsx
```diff
@@ -95,9 +99,9 @@ export default function Row({
         {RowChildren}
       </label>
       <input
+        type="file"
```
Damrod, line 102: why did you move this three lines above? 😯

webapp/src/shared/components/Avatar/EditableAvatar.tsx
```diff
@@ -37,7 +30,11 @@ export default function EditableAvatar({
     )}px`,
   };
   return (
-    <div className="editable-avatar">
+    <div
+      className={`editable-avatar editable-avatar--${
```
Damrod, line 34: what do you disable here? 😯

webapp/src/pages/UsersPage/UsersPage.tsx
```diff
@@ -9,9 +9,12 @@ import {
   User,
   UserControllerGetUsersRequest,
 } from '../../shared/generated';
-import { DispatchPage } from '../../shared/components/index';
+import { RowsTemplate } from '../../shared/components/index';
 import { useCallback } from 'react';
 import { usersApi } from '../../shared/services/ApiService';
+import { SearchContextProvider } from '../../shared/context/SearchContext';
+
+const ENTRIES_LIMIT = 15;
```
Damrod, line 17: why 15?

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -58,18 +67,32 @@ export default function EditAvatarPage() {
           avatarY: FormatNumber(picturePosition.y),
```
Damrod, line 67: Maybe we should do this transformation inside useDrag, passing FormatNumber as a callback as well, or maybe we shouldn't pass reverseTransform as a callback here, and do both transformations here. This is something I was thinking of doing. What do you think?

webapp/src/shared/components/Avatar/EditableAvatar.css
```diff
@@ -20,11 +25,10 @@
 }
 
 .editable-avatar__image {
-  user-drag: none;
   -webkit-user-drag: none;
   margin: 0;
-  width: 16rem;
-  height: 16rem;
+  width: 15rem;
```
Damrod, line 30: This is veeeeery tricky to change. But I understand where is this coming from. I'm not sure what's wrong with my parameters. I've fixed them twice, and I think there is, again, some problem with the scale of the image and the offset of the picturePosition, for each of the different avatar sizes. Maybe someone has some idea.

Damrod:
Besides the comments I have left, and besides the fact that I haven't yet had the opportunity to actually run to see that it's working, I love it. Will take a look tomorrow. Thank you!

webapp/src/shared/hooks/UseIsElementVisible.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+
+export function UseIsElementVisible() {
+  const [isVisible, setIsVisible] = React.useState(false);
+  const observer = React.useRef<IntersectionObserver | null>();
```
gbudau, line 5: praise: Now we can reuse it in more places, good job :+1: 

suggestion: I think the convention for custom hooks is to start with lowercase letters -> `useIsElementVisible`?

gbudau (approved):
Thanks for improving it! LGTM :rocket:

webapp/src/pages/UsersPage/UsersPage.tsx
fizcamposancos: If I'm not wrong we already have this limit in the back-end set to 20, in the [user.pagination.dto.ts](https://github.com/42AGV/ft_transcendence/blob/main/transcendence-app/src/user/dto/user.pagination.dto.ts) file
MAX_USER_ENTRIES_PER_PAGE = 20;
Though, I am moving this constant to /shared/constants.tsx file with the chat limit that I am working on

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -105,15 +134,26 @@ export default function EditAvatarPage() {
...
+        <Text
+          variant={TextVariant.PARAGRAPH}
+          color={
+            status.type === 'success' ? TextColor.ONLINE : TextColor.OFFLINE
+          }
+        >
+          {status.message}
+        </Text>
+      )}
       <Button
```
Damrod, line 157: The button is too wide: this is not how it was supposed to look in figma. Someone broke the button width at some point in early September.

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
fizcamposancos: I agree with @Damrod, or at least I was using this button feature in the chat page.
Or we can add it separately afterwards

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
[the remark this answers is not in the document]
vdedios: Agree, data validation is not needed anymore, removed and simplified a bit this method!

webapp/src/shared/components/templates/RowsPageTemplate/RowsPageTemplate.tsx
vdedios: I'd say to add later it in chat's page task, we will know better the needs we have there!

webapp/src/shared/components/Row/Row.tsx
vdedios: I think this is a leftover of me trying to add support for adding a local directory path to update the image, but it was tricky and unnecessary at the end

webapp/src/pages/UsersPage/UsersPage.tsx
vdedios: It is down the max_limit. This constant is necessary so we can set de offset of our pagination queries (we fetch data ENTRIES_LIMIT by ENTRIES_LIMIT)

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
vdedios: I think just passing the transformed values to `useDrag` would be the best :+1:

webapp/src/shared/components/Avatar/EditableAvatar.css
vdedios: Yes, this was me playing around with sizes, I have restored the values. Thanks for pointing out

webapp/src/shared/components/Row/Row.css
```diff
@@ -17,14 +19,23 @@
   display: none;
 }
 
+.row-icon {
+  grid-column: 1/2;
+}
+
```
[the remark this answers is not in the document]
vdedios: I think this should be `row-avatar`, we don't have any place indication its grid-column

webapp/src/shared/components/Row/Row.css
```diff
@@ -17,14 +19,23 @@
...
 .row_text_wrapper,
 .row_text_wrapper > * {
   margin: 0;
   text-align: left;
+  overflow: hidden;
+  grid-column: 2/3;
```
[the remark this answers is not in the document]
vdedios: good catch! updated to `grid-column: span 1;`

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -82,6 +105,12 @@ export default function EditAvatarPage() {
     setImgData({
       imgFile: event.target.files[0],
       imgName: event.target.files[0].name,
+      imgSrc: URL.createObjectURL(event.target.files[0]),
+    });
+    console.log(imgData);
+    setStatus({
+      type: 'pending',
+      message: '',
```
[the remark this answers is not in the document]
vdedios: yes! again, leftovers :laughing:

webapp/src/shared/context/SearchContext.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// https://reactjs.org/docs/context.html#reactcreatecontext
+const SearchContext = React.createContext<Context<any> | undefined>(undefined);
+
+export const SearchContextProvider = <T,>({
+  children,
+  fetchFn,
+  maxEntries,
+}: SearchContextProps<T>): JSX.Element => {
```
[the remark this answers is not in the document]
vdedios: I won't go with `query.limit`, its really and offset representing where in the data we are and where we should start fetching in the next interval

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
vdedios: Yes not 1:1 like Figma :smile: Since this is no blocking to merge this, we can review it in future tasks/chat iterations

vdedios:
Merging this! Thanks all of you for your comments :heart:

webapp/src/shared/hooks/UseIsElementVisible.tsx
vdedios: done!

PR #201, opened 2022-10-13
# 200 improve routing in webapp

### LINKED ISSUE #200: Improve webapp routing
This issue is about fixing and improving some aspects about routing:

- We fetch the auth user in every single navigation:
  - We already know who that user is because is auth (and present in auth context), so the idea would be reusing it (and updating it's state everytime we update it)
- We have a bit weird routes.
  - For example, for a normal user we would navigate to `/users/239482309482390-234o89234o238`. But for the auth user, `/user`
  - A more unified and simpler way to do this would be simply: `/user/<username>` and `/users` to display all.
- We have inconsistencies in auth user page: if navigated from a user row, it displays it like a normal user. But if navigate it from user profile, it displays it like a `isMe` user.
  - We can fix this using a single route for a user as mentioned in previous point.

vdedios:
closes #200 

:warning: **Read before #200 to get full context about this PR**  :warning: 

(I've worked on top of https://github.com/42AGV/ft_transcendence/pull/198 so this PR is pointing that branch)


webapp/src/index.tsx
```diff
@@ -103,6 +95,14 @@ root.render(
...
             />
           )}
+          <Route
+            path="*"
+            element={
+              <RequireAuth>
+                <Navigate to="/" />
+              </RequireAuth>
+            }
+          />
```
vdedios, line 105: If route does not exist, navigate to default

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
```diff
@@ -40,39 +51,40 @@ export default function EditAvatarPage() {
...
         setStatus({
           type: 'success',
           message: 'Image visible area saved correctly.',
-        }),
-      )
+        });
+        setAuthUser((prevState) => {
+          if (!prevState) return null;
+          return { ...prevState, avatarX, avatarY };
+        });
```
vdedios, line 86: Since we are using persited authUser, we should update its state everytime we update it on backend

webapp/src/shared/hooks/UseNavigation.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const goBack = () => {
+    return () => {
+      navigate(-1);
+    };
+  };
+
+  const goBackTo = (route: string) => {
+    return () => {
+      navigate(-1);
+      navigate(route);
```
vdedios, line 23: Since we are navigating to another route, we have to first pop out of the stack history previous location

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -18,28 +18,23 @@ import {
...
-  );
-  const getUserById = useCallback(
-    () => usersApi.userControllerGetUserById({ userId: param.id! }),
-    [param.id],
-  );
-  const { data: user } = useData(isMe ? getCurrentUser : getUserById);
-  const navigate = useNavigate();
+const UserComponentTemplate = ({
+  user,
+  isAuthUser,
```
Damrod, line 35: As I wrote in the issue, where I think it may go unnoticed:

> We have inconsistencies in auth user page: if navigated from a user row, it displays it like a normal user. But if navigate it from user profile, it displays it like a isMe user.

This is intentional. If it's displayed from inside IsMe , it displays all available information. Otherwise, it displays info as it would be available to others. This exists, for instance, in the 42 intra as well.

I honestly don't think it's a good idea to remove the feature of being able to see the user page of the user from both points of view (self, and from the point of view of others)

Damrod:
Other than not agreeing with the removal of the feature of being able to see the user of the user currently logged in from two points of view, from the comment above, I think your changes are an improvement. Thanks!

webapp/src/index.tsx
gbudau: question: What do we think about displaying a `NotFoundPage`?

webapp/src/pages/EditAvatarPage/EditAvatarPage.tsx
gbudau: praise: I like this change. Thank you!

webapp/src/shared/hooks/UseNavigation.tsx
gbudau: question: When do we need this behavior to go back and to another page?

webapp/src/pages/UserPage/UserPage.tsx
gbudau: I agree with this :+1:
I think it's nice to have both views.

gbudau (changes requested):
Thanks for these improvements. I left a few questions and suggestions.

webapp/src/shared/hooks/UseNavigation.tsx
vdedios: Because the authUser could change its `username` and, since routes are now based on username, the history stack would point to the old username route

webapp/src/index.tsx
vdedios: Nice idea! I've created a simple one :)

webapp/src/pages/UserPage/UserPage.tsx
vdedios: Ok, agree and I've restore the behaviour! I though it was a "side feature" we had due to how our routing was designed rather than a former feature :smile:

transcendence-app/src/user/dto/user.avatar.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { StreamableFile } from '@nestjs/common';
+import { IsNotEmpty, IsUUID, ValidateIf } from 'class-validator';
+
+export class UserAvatarDto {
+  @IsUUID()
+  @ValidateIf((object, value) => value !== null)
```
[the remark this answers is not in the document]
vdedios: you're right, with static checking is enough. Removed!

webapp/src/index.tsx
```diff
@@ -90,7 +82,7 @@ root.render(
             }
           />
           <Route
-            path={EDIT_AVATAR_URL}
+            path={EDIT_AVATAR_URL(':username')}
```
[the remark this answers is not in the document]
vdedios: Yes, this use was a little bit weird, I've moved it as a constant again :ok_hand:

webapp/src/pages/EditUserPage/EditUserPage.tsx
```diff
@@ -7,51 +7,58 @@ import {
...
       <div className="edit-user-loading">
         <Loading />
       </div>
     </div>
   ) : (
     <div className="edit-user-page">
-      <Header icon={IconVariant.ARROW_BACK} onClick={goBack(navigate)}>
+      <Header
+        icon={IconVariant.ARROW_BACK}
+        onClick={goBackTo(`${USER_URL}/${authUser.username}`)}
```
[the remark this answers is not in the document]
vdedios: Oops it seems this works right on Chrome based browsers but not in others such as Firefox :fox_face:

webapp/src/pages/EditUserPage/EditUserPage.tsx
```diff
@@ -7,51 +7,58 @@ import {
...
+  const { username } = useParams();
+  const { isMe, authUser } = useAuth(username);
   const navigate = useNavigate();
+  const { goBackTo } = useNavigation();
+
+  useEffect(() => {
+    if (!isMe) {
+      navigate('/');
+    }
+  }, [isMe, navigate]);
```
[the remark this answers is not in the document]
vdedios: This behaviour is related to the same as above. Firefox browser may have a different history behaviour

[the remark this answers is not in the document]
vdedios: I've been investigating some solutions and the less problematic is the static url as you say. I've tried to update the url based on the new `username` but I think that, despite possible, it does not worth the effort.

I've added a new `myprofile` route. So now we have:
- `/myprofile` for that special views for the `me` user
- `/user/<username>` to view other uses profiles

webapp/src/pages/EditUserPage/EditUserPage.tsx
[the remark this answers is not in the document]
vdedios: (read comment bellow)

webapp/src/shared/hooks/UseNavigation.tsx
vdedios: (despite the explanation we don't have this anymore, see comment bellow)

gbudau (approved):
Thank you for your efforts! Well done!

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -13,33 +13,32 @@ import {
...
-    () => usersApi.userControllerGetCurrentUser(),
-    [],
-  );
-  const getUserById = useCallback(
-    () => usersApi.userControllerGetUserById({ userId: param.id! }),
-    [param.id],
-  );
-  const { data: user } = useData(isMe ? getCurrentUser : getUserById);
-  const navigate = useNavigate();
+type UserComponentTemplateProps = {
```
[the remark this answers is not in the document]
vdedios: done! thanks for your suggestion ;)

PR #245, opened 2022-10-19
Chat page and create chat page

fizcamposancos:
close #231
close #232
close #233
close #234
close #235 
close #236
close #237
close #244
Please, let me know if you want me to split this PR in several PR to each of the tasks closed
Design is not closed, so any suggestion will be wellcome

PR #246, opened 2022-10-20
#208/Create a database table to persist the blocks

### LINKED ISSUE #208: Create a database table to persist the blocks
Users can block other users.

gbudau:
close #208


db/setup/schema.sql
```diff
@@ -1,24 +1,24 @@
 CREATE TABLE
-  IF NOT EXISTS LocalFile(
-    "id" UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+  IF NOT EXISTS LocalFile (
+    "id" UUID DEFAULT gen_random_uuid () PRIMARY KEY,
```
gbudau, line 3: Sorry for these formatting changes. I've installed the [Prettier SQL](https://marketplace.visualstudio.com/items?itemName=inferrinizzard.prettier-sql-vscode) extension to format SQL files, which uses the [sql-formatter](https://github.com/sql-formatter-org/sql-formatter) library.

db/setup/schema.sql
```diff
@@ -27,6 +27,14 @@ CREATE TABLE
   IF NOT EXISTS AuthProvider (
     "providerId" TEXT NOT NULL,
     "provider" ProviderType NOT NULL,
-    "userId" UUID REFERENCES Users(id),
+    "userId" UUID REFERENCES Users (id) ON DELETE CASCADE,
```
Damrod, line 30: what does on delete cascade do?

> ON DELETE CASCADE clause in MySQL is used to automatically remove the matching records from the child table when we delete the rows from the parent table. It is a kind of referential action related to the foreign key.

Which is here the child table? Can you make an example, please?

Damrod (approved):
LGTM!

db/setup/schema.sql
fizcamposancos: If I am not wrong, the child table is this one, the one with `ON DELETE CASCADE`. So, for example, if you delete a user, his auths, blocks, ... (games, friends,... when we have everything ready) would be deleted automatically "on cascade".

fizcamposancos (approved):
LGTM

gbudau:
I merged all stacked PRs into this one, and will merge this into main after we merge #245

db/setup/schema.sql
vdedios: I think is a good idea, I'll install it to!

transcendence-app/src/shared/relationship/relationship.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { RelationshipService } from './relationship.service';
+import { DbModule } from '../db/db.module';
+
+@Module({
+  imports: [DbModule],
+  providers: [RelationshipService],
+  exports: [RelationshipService],
+})
+export class RelationshipModule {}
```
vdedios, line 10: As stated in https://github.com/42AGV/ft_transcendence/pull/248#discussion_r1004798700 I disagree with having yet another module for this. This has no domain logic in our app, it's just a status between users.

I suspect you had this approach due to this being in a new sql table, but I don't see a problem consuming this block repository from users module. The new table is clearly tied to users :crying_cat_face:

If we fear user controller/service etc starts growing we could split controller in some files to split logic.

Damrod: I also don't understand where is this thing coming from. Was it here yesterday when I reviewed it?

gbudau: Ok, I will move this logic into the Users module.

@Damrod, this was the `BlockModule` yesterday. I renamed it to the `RelationshipModule` today to add the friendship relation in the same module.

gbudau: Done :rocket:
I've added this change in #258 because it's the last PR from the stack of PRs related to this feature

Damrod: You may think it's fun, but I've spent one hour tracking down bugs, since I decided to merge your branch into my latest branch, to try and not have merge conflicts.

gbudau: Sorry, @Damrod.

I think squashing the PRs does not help because if we stack PRs and a squash creates a different commit with all the commits into one, the PRs on top lose the reference to the previous commits.
What do we think about changing our workflow to a merge commit even if the git history will become less "clean"?

cc @fizcamposancos @vdedios

Damrod: I agree. No more squashing, as long as we have so many dependencies.

vdedios: I think, rather than a matter of squashing being the problem, I see the problem is indeed working over other's ongoing work when it has not been merged yet.

Not merged code is prone to change and if one decides to start working over other's, on going, work one may be responsible to keep track of updates because the base branch will likely have changes 🤷‍♂️

Not blaming anyone, is the only thing that can be done now, just pointing out the real problem

gbudau: Squash and merge, or rebase and merge, are incompatible with stacked PRs. We have to become git wizards to solve merge conflicts.

Stacking PRs + merge commits are very beneficial and, in my opinion, could work better with the way we work asynchronously.
We can do small PRs, and if one of the previous PRs takes more time to review or merge, we can continue working on other PRs without too many merge conflicts.

I think we are gaining history readability at the cost of velocity.

Damrod: fportela said
> Si te soy sincero tengo claro que es el squash merge o el merge commit. Lo lei en su día pero ya no me acuerdo. Pero en cualquier caso, si sirve para ahorrar problemas, a costa de trazabilidad, por mi no hay fallo 
I wish he had said so in public. Anyhow, I proceed to enable merge commits into master.

gbudau:
Thanks!

PR #248, opened 2022-10-21
#214/Create an endpoint to block a user

### LINKED ISSUE #214: Create an endpoint to block a user
Create a `POST /users/block/{userId}` endpoint to block a user

gbudau:
This PR adds the `POST /api/v1/users/block/{userId}`
When calling this endpoint, the authenticated user will block the user with `userId`

close #214


Damrod:
It would be cool to have this route   "/api/v1/users/block/{userId}" either in the description of the issue or here in the PR for reference.

Damrod (approved):
It works (and by extension, the branches this is trying to merge into)

transcendence-app/src/shared/relationship/relationship.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { RelationshipService } from './relationship.service';
+import { DbModule } from '../db/db.module';
+
+@Module({
+  imports: [DbModule],
+  providers: [RelationshipService],
+  exports: [RelationshipService],
+})
+export class RelationshipModule {}
```
vdedios, line 10: Why do we need a new module for this? Why this can't be a part of a user? What else do we expect to be blocked in the future? What does "relationship" means?

PR #251, opened 2022-10-23
#176 default avatars on registration

### LINKED ISSUE #176: Create placeholder images for users registered with no image
Users may not provide an image when registered. This task is about generating a default one. Right now we use a random-image url which makes loading slow + all users end up sharing the same one

vdedios:
closes #176

### DIFF
--- a/transcendence-app/src/auth/auth.service.ts
+++ b/transcendence-app/src/auth/auth.service.ts
@@ -10,7 +10,7 @@ import { LoginUserDto } from '../user/dto/login-user.dto';
 import { RegisterUserDto } from '../user/dto/register-user.dto';
 import { User } from '../user/user.domain';
 import { UserService } from '../user/user.service';
-import { LocalFileService } from 'src/shared/local-file/local-file.service';
+import { LocalFileService } from '../shared/local-file/local-file.service';
 
 const scrypt = promisify(_scrypt);
 
--- a/transcendence-app/src/auth/auth.service.ts
+++ b/transcendence-app/src/auth/auth.service.ts
@@ -13,6 +13,8 @@ import { RegisterUserDto } from '../user/dto/register-user.dto';
 import { User } from '../user/user.domain';
 import { UserService } from '../user/user.service';
 import { LocalFileService } from 'src/shared/local-file/local-file.service';
+import { Stream } from 'stream';
+import { AVATARS_PATH } from '../user/constants';
 
 const scrypt = promisify(_scrypt);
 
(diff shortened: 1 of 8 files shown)


Damrod:
If you have done this in pair programming, you don't need a review.

gbudau:
> Have you paired up, as it would seem?

We did pair programming :+1

> If you have done this in pair programming, [you don't need a review](https://blog.submain.com/code-review-vs-pair-programming-2/).

I think this is subjective. For example, [another article](https://mergeboard.com/blog/7-pair-programming-code-reviews/) recommends code review even after pair programming.

transcendence-app/types/github-like-avatar-generator.d.ts
```diff
@@ -0,0 +1,20 @@
+type GenerateAvatarParams = {
+  blocks: number;
+  width: number;
+  fromPixels?: Array<string>;
+};
+
+type GenerateAvatarReturnType = {
+  base64: string;
+  pixels: Array<string>;
```
Damrod, line 9: We have been changing Array<string> to string[], every now and then. I was sold on the idea of using this(by one of you two), but I was argued against (by the other one of you two), and now I'm not sold anymore. I don't think at this point Array<string> is our preferred choice across the project. Please. Let's stick to one convention.

transcendence-app/src/shared/local-file/local-file.service.ts
```diff
@@ -76,4 +107,17 @@ export class LocalFileService {
   ): Promise<LocalFile | null> {
     return this.localFileRepository.updateById(id, localFileDto);
   }
+
+  // default values as suggested in lib example
+  async createRandomSVGFile(blocks = 6, width = 100) {
+    const newAvatar = generateAvatar({
```
Damrod, line 113: As I have already mentioned, it's kind of horrible to review a PR where half of the local variable types are relying on somehow obscure type inference.

transcendence-app/src/auth/auth.service.ts
```diff
@@ -62,10 +66,14 @@ export class AuthService {
     const result = salt + '.' + hash.toString('hex');
 
     const { confirmationPassword: _, ...newUser } = user;
-    return this.userService.addUser({
+
+    const avatarDto = await this.localFileService.createRandomSVGFile(12, 512);
+    const userDto = {
       ...newUser,
       avatarId: null,
```
Damrod, line 73: I can see, because I'm running it locally, and I'm looking at the database, that at some point the avatarId gets assigned and it's not null anymore. I see that it's happening in userService.AddAvatarAndUser(). This particular program flow you have chosen has the inconvenient drawback of having to assign null in the beginning. What about we refactor this a bit so that we never have to assign null to the avatarId, and like that we can remove `| null` from the avatarId type?

Damrod: (We will also have to manually hardcode some avatars into the mock users we are seeding the database with, but shouldn't be a big problem)
BTW, I don't think this should block the merging of this PR, but still, think about this and let me know your opinion, please.

transcendence-app/types/github-like-avatar-generator.d.ts
gbudau: Done :+1:

transcendence-app/src/shared/local-file/local-file.service.ts
gbudau: Sorry, done :+1:

transcendence-app/src/auth/auth.service.ts
gbudau: Good point. We plan to make the `avatarId` mandatory in #252

Damrod (approved):
LGTM! Thanks for this feature

Issue #253, opened 2022-10-23
Simplify submit chatroom message pattern


Damrod:
We currently submit a message in a chatroom only after the user presses the button. Like this, we have access to the enter key to submit newlines. On the other hand, this is cumbersome, and unintuitive. We can do better, without giving up on features, if, for instance we accept an enter keypress to submit, and an enter + shift to insert a newline.

PR #254, opened 2022-10-23
submit keypress with enter in chatroom

Damrod:
Check issue #253

### DIFF
--- a/webapp/src/pages/ChatRoom/components/ChatRoomMessageInput/ChatRoomMessageInput.tsx
+++ b/webapp/src/pages/ChatRoom/components/ChatRoomMessageInput/ChatRoomMessageInput.tsx
@@ -1,5 +1,4 @@
-import React, { useState } from 'react';
-import { Button, ButtonVariant } from '../../../../shared/components';
+import React, { useRef, useState } from 'react';
 import socket from '../../../../shared/socket';
 import './ChatRoomMessageInput.css';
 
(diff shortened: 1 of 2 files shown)


fizcamposancos (approved):
Simple and so practical. I love it.

webapp/src/pages/ChatRoom/components/ChatRoomMessageInput/ChatRoomMessageInput.tsx
```diff
@@ -10,18 +9,21 @@ type ChatRoomMessageInputProps = {
...
-    e.preventDefault();
-    socket.emit('chatRoomMessage', { chatRoomId: to, content: value });
-    setValue('');
+  const onEnterPress = (e: React.KeyboardEvent) => {
+    if (e.code === 'Enter' && !e.shiftKey) {
+      e.preventDefault();
+      value !== '' &&
+        socket.emit('chatRoomMessage', { chatRoomId: to, content: value });
+      setValue('');
+    }
```
gbudau, line 18: praise: Nice solution! Thanks!

webapp/src/pages/ChatRoom/ChatRoom.tsx
```diff
@@ -29,7 +29,7 @@ function ChatRoom() {
   return (
     <div className="chat-room">
       <div className="chat-room-header">
-        <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
+        <Header icon={IconVariant.ARROW_BACK} onClick={goBack()}>
```
gbudau, line 32: suggestion (if-minor): To prevent this bug in the future, we could change the `goBack` function to something like this:

```diff
diff --git a/webapp/src/shared/hooks/UseNavigation.tsx b/webapp/src/shared/hooks/UseNavigation.tsx
index 16dcc94..a1bd92a 100644
--- a/webapp/src/shared/hooks/UseNavigation.tsx
+++ b/webapp/src/shared/hooks/UseNavigation.tsx
@@ -1,20 +1,14 @@
 import { NavigateFunction, useNavigate } from 'react-router-dom';
 
-type Callback = () => void;
-
 type UseNavigationReturn = {
   navigate: NavigateFunction;
-  goBack: () => Callback;
+  goBack: VoidFunction;
 };
 
 export const useNavigation = (): UseNavigationReturn => {
   const navigate = useNavigate();
 
-  const goBack = () => {
-    return () => {
-      navigate(-1);
-    };
-  };
+  const goBack = () => navigate(-1);
 
   return { navigate, goBack };
 };
```

gbudau (approved):

webapp/src/pages/ChatRoom/ChatRoom.tsx
vdedios: why not just use the hook we already have?https://github.com/42AGV/ft_transcendence/blob/7befac8cae299e7e84a97eb3573cc4619770b22c/webapp/src/shared/hooks/UseNavigation.tsx#L13

Damrod: We already are using that hook. You mean why don't we use it in its current form?

PR #260, opened 2022-10-25
#239/method to add chatroom member to db

Damrod:
as detailed in #239


transcendence-app/src/chat/chat.domain.ts
```diff
@@ -17,3 +18,15 @@ export class Chat {
     Object.assign(this, entity);
   }
 }
+
+export class ChatMember {
```
gbudau, line 22: praise: I like how you implement this class, looks good :+1:

suggestion: What do we think about moving it to a new file, `chatmember.domain.ts`?

transcendence-app/src/chat/dto/create-chatmember.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { IsNotEmpty, IsString } from 'class-validator';
+
+export class CreateChatMemberDto {
```
gbudau, line 3: thought: We use `ValidationPipe` and DTOs to [auto-validate](https://docs.nestjs.com/techniques/validation#auto-validation) the incoming request. I see that we are using this class internally, and `ValidationPipe` only validates the controller endpoints.

suggestion: Consider removing this class unless we want to add in the future an endpoint like `POST /chat/room/members` where we specify the `chatId` and `userId` in the request body.

gbudau (approved):

transcendence-app/src/chat/dto/create-chatmember.dto.ts
Damrod: There have already been some [comments](https://github.com/42AGV/ft_transcendence/pull/259#discussion_r1005891878) requesting the addition of this feature. But I do not plan to add it now, and we shouldn't pay for what we don't use. I understand what you are saying. I will remove it.

Damrod: done

transcendence-app/src/chat/chat.domain.ts
Damrod: done

Issue #252, opened 2022-10-23
Use seed files to fill database on dev deployment and improve them


vdedios:
Right now we have a mix of SQL queries and knex seed files to fill database with mocked data. That being said, knex seed files is a much better solution. So this task is about:

- Use knex seed files instead of SQL queries to fill db
- Improve user seed files to also add default avatars for users

Also, to improve user seed file:
- Make avatar field mandatory in User table schema
- Remove logic in webapp to provide a default image url since we will always recieve a avatarId (it's going to be mandatory)

PR #261, opened 2022-10-25
 #238/create chatroom members table 

Damrod:
as detailed in #238


db/setup/schema.sql
```diff
@@ -1,24 +1,24 @@
...
-  IF NOT EXISTS Users(
-    "id" UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+  IF NOT EXISTS Users (
+    "id" UUID DEFAULT gen_random_uuid () PRIMARY KEY,
     "username" VARCHAR(20) NOT NULL UNIQUE,
     "email" VARCHAR(50) NOT NULL,
     "fullName" VARCHAR(255) NOT NULL,
     "password" TEXT,
-    "avatarId" UUID REFERENCES LocalFile(id) UNIQUE,
+    "avatarId" UUID UNIQUE REFERENCES LocalFile (id) ON DELETE SET NULL,
```
vdedios, line 18: we plan to make this mandatory (not null) on https://github.com/42AGV/ft_transcendence/issues/252#event-7648782981. Because of having add default avatars (and also removing the fallback url we have on webapp) all users should now always have an avatar

db/setup/schema.sql
```diff
@@ -42,3 +42,15 @@ CREATE TABLE
...
   );
+
+CREATE TABLE
+  IF NOT EXISTS ChatRoomMembers (
+  "chatId" UUID REFERENCES ChatRoom (id) ON DELETE CASCADE,
+  "userId" UUID REFERENCES Users (id) ON DELETE CASCADE,
+  "joinedAt"  TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  "admin" BOOLEAN DEFAULT false,
+  "muted" BOOLEAN DEFAULT false,
+  "banned" BOOLEAN DEFAULT false,
```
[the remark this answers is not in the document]
Damrod: This proposition for a schema for this table has been hanging [here](https://www.figma.com/file/AEIkPtefbuVG785a780S1G/Database-Schema?node-id=1%3A783) for longer than a month. Some people  have voiced their opinion, and wishes for things to be done differently. You could have.

Won't do, because of this, and because of the reasons posted to the slack conversation. Also, will be glad if you don't change it later on.

db/setup/schema.sql
Damrod: Good. Will merge when it's on master. Also, this is actually taken from the block MR. Not particularly attached to these changes.

gbudau (approved):
Thank you! LGTM!

Issue #204, opened 2022-10-19
⚛️ Create an edit chat room details page


gbudau:
The owner can update the chat room name, avatar, or password.
Webapp URL -> `/chatroom/{chatId}/edit`

PR #262, opened 2022-10-26
#204/edit chatroom details page

Damrod:
As described in #204


transcendence-app/src/chat/chatroom/infrastructure/db/chatroom.entity.ts
```diff
@@ -26,15 +27,17 @@ export interface ChatroomData {
...
   password: string | null;
   avatarId: string | null;
   avatarX: number = 0;
   avatarY: number = 0;
   createdAt: Date;
   ownerId: string;
+  @ApiProperty({ type: Boolean })
   @Expose()
-  get public(): boolean {
+  get isPublic(): boolean {
```
Damrod, line 40: public is a keyword, and becomes _public when openapi generator makes the frontend code. isPublic doesn't, hence, to stay consistent, the change.

webapp/src/shared/components/Header/Header.tsx
```diff
@@ -62,8 +87,43 @@ export default function Header({
...
+    if (statusVariant !== 'button') {
+      statusElement = (
+        <div className="header-status">
+          {statusVariant && <Status variant={statusVariant} />}
+        </div>
+      );
+    } else {
+      if (buttonProps !== undefined) {
+        if (!windowIsBig) {
+          style = {
```
Damrod, line 102: This whole thiong is here (and inmainTabButtons as well) to not have the composition look like a d**k such as this
not_two_balls

Damrod:
I can still improve it slightly, but I think it's good if you start to take a look at this, since it's a big one.

transcendence-app/src/chat/chatroom/infrastructure/db/chatroom.entity.ts
gbudau: Good catch!

webapp/src/shared/components/Header/Header.tsx
vdedios: I wouldn't worry too much merging it as is in this PR. We may have to explore other ideas but for this PR I think is good enough

webapp/src/pages/ChatPage/ChatPage.tsx
```diff
@@ -42,23 +42,30 @@ export default function ChatPage() {
     [],
   );
   const navigate = useNavigate();
-
+  const windowIsBig = useMediaQuery(768);
+  const chatButtons = [
+    {
+      buttonSize: windowIsBig ? ButtonSize.LARGE : ButtonSize.SMALL,
```
vdedios, line 48: I think this responsibility should be button's one, I mean, handling its size. It should know wheter it should display one version or another depending on screenSize and could be done in pure CSS.

We could have another button component wich inherits from a standard button which adds this media-query behaviour (`ReactiveButton`, or any other better name 😄 ) This way we wouldn't even need to handle queries from `tsx` code

webapp/src/shared/components/Header/Header.tsx
vdedios: I've been thinking about this, what do you think about moving this exit button inside the edit page? This way we would avoid:
1. Having to inline styles and calculate translations etc
2. Simplifying mobile UI since one button is leaner and we gain more space for other content

So, a normal user will see this exit button and, an owner, admin etc. would see the edit chatroom option and would find the option to exit inside that page.

webapp/src/shared/hooks/UseMediaQuery.tsx
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useState } from 'react';
+
+export const useMediaQuery = (screenResolutionInPx: number) => {
```
vdedios, line 3: This could be avoided with some of the suggestions I made above

vdedios (dismissed):
Very nice work overall 👏  

I've left some comments and requesting changes because I think we should review the usage of inlined styles with some comments I've left.

Thanks!

webapp/src/pages/ChatPage/ChatPage.tsx
Damrod: You can have two buttons, overlapping, and display one or the other (setting the non relevant one to display:none) depending on the media query. On the other hand, ~what is the principle you would quote to explain that it's not~ why do you think it is not acceptable to write inline styles?

webapp/src/shared/components/Header/Header.tsx
Damrod: That page already has two buttons, and they are stacked. You refused to accept smaller buttons. How would you make the design of the page with three buttons look like?

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+  }, [authUser, chatroom]);
+  const leaveChatroom = useCallback(async () => {
+    if (!chatroomId) return;
+    try {
+      if (isOwner && !ownerHasBeenWarned) {
+        warn(
+          'You are the owner of this chatroom. If you leave, the chatroom ' +
+            'will be deleted. If you attempt to leave again, it will happen. ' +
+            'This is the first and only warning.',
+        );
```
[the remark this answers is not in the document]
Damrod: I think this is a good idea, but I didn't think it was the righ time/place to address it. I just wanted to not leave the situation completely unhandled/overlooked, and to use this notification component you made.

webapp/src/shared/components/Header/Header.tsx
```diff
@@ -35,15 +40,35 @@ type NoFigureHeader = HeaderCommon & {
...
+    case 2:
+      return 30;
+    case 3:
+      return 34;
+    case 4:
+      return 38;
+    default:
+    case 5:
+      return 43;
+  }
```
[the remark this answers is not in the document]
Damrod: Well, these are experimental coefficients. I can spend several hours trying to find what's the rule, or make a table that makes the independent variable match a dependent variable that makes the thing look nice. I could name them, but it would be something along the line of "yDisplacementThatMatchesTheCaseOfFiveElementsAndMakesItLooksNice", if I'm being honest, which is a very useless name.
I could also make it look like a table, with types and coordinates, and it would look better, but it's completely indistinguishable conceptually from this implementation, as long as I name the function properly, which it is.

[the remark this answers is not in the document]
Damrod: Anyway, it depends on whether we want vertically displayed arrays of buttons or not.
If we don't need them, we don't need this, we're good.

webapp/src/shared/hooks/UseMediaQuery.tsx
Damrod: Why should we, I wonder, avoid this.

webapp/src/pages/ChatPage/ChatPage.tsx
Damrod: done

webapp/src/shared/hooks/UseMediaQuery.tsx
Damrod: Apparently, we did not need it. I think if I was still trying to align it to the top of the header, I would still not know how to work around it. But as soon as I removed, this wasn't a problem anymore.

webapp/src/shared/components/Header/Header.tsx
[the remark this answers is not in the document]
Damrod: The vertical array of buttons displays properly outside of the header component, and without all this math.
I know you don't like the idea of having more than one button in the chat page, but I'm gonna need more than your personal preference to remove it.

webapp/src/shared/components/Header/Header.tsx
Damrod: I removed the two buttons, but I'm not convinced it's a good idea.
Maybe the link in the header title text brings to the edit chat page, and there is just one red button and it is, like for normal users, just to leave the chat?

Damrod: The latest proposal is an implementation of my last idea from the message above.

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const mapChatMemberToRow = (member: ChatroomMemberWithUser): RowItem => {
+    return {
+      iconVariant: IconVariant.EDIT,
+      avatarProps: {
+        url: `${AVATAR_EP_URL}/${member.avatarId}`,
+        status: 'offline',
+        XCoordinate: member.avatarX,
+        YCoordinate: member.avatarY,
+      },
+      url: `${CHATROOM_URL}/${chatroomId}/member/${member.username}/edit`,
```
gbudau, line 55: thought: Currently, we use the same url and icon for all the members.
suggestion: What do you think about adding a TODO to deal with the case where the user is not an owner or administrator?

webapp/src/shared/components/Button/ReactiveButton.tsx
```diff
@@ -0,0 +1,12 @@
+import './ReactiveButton.css';
+import React from 'react';
+import Button, { ButtonProps, ButtonSize } from './Button';
+
+export default function ReactiveButton(props: ButtonProps) {
```
gbudau, line 5: praise: Very nice :+1:

webapp/src/shared/context/SearchContext.tsx
```diff
@@ -4,13 +4,15 @@ import { useData } from '../hooks/UseData';
 
 type SearchContextProps<T> = {
   children: React.ReactNode;
-  fetchFn: (requestParams: {}) => Promise<T[]>;
+  fetchFn: <RequestType extends Query>(
+    requestParams: RequestType,
+  ) => Promise<T[]>;
```
gbudau, line 9: praise: Thanks for improving the typings and good idea to extend the `Query` type

webapp/src/shared/context/SearchContext.tsx
```diff
@@ -39,7 +41,7 @@ function queryReducerCallback(state: Query, action: QueryReducerAction) {
   if (type === 'SEARCH') {
     return { search: action.value, offset: 0 };
   } else if (type === 'LOAD_MORE') {
-    return { ...state, offset: state.offset + action.limit };
+    return { ...state, offset: state.offset ?? 0 + action.limit };
```
gbudau, line 44:
```suggestion
    return { ...state, offset: (state.offset ?? 0) + action.limit };
```
suggestion: The pagination doesn't work. These changes could fix it.

gbudau (approved):
Thank you for this feature and all the improvements! LGTM :+1:

webapp/src/shared/context/SearchContext.tsx
Damrod: Should be fixed.

webapp/src/shared/context/SearchContext.tsx
Damrod: Thanks!
I've gona ahead and placed in the shared types file.

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
Damrod: done

webapp/src/pages/ComponentsBookPage/ComponentsExamples/HeaderExample.tsx
```diff
@@ -31,5 +34,17 @@ export const HeaderExample = () => (
     <Header avatar={avProps} iconNavigationUrl="/" statusVariant="playing">
       {placeholderTitle}
     </Header>
+    <Header
```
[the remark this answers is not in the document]
Damrod: Thx!

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
```diff
@@ -0,0 +1,159 @@
...
+        icon={IconVariant.ARROW_BACK}
+        onClick={goBack}
+        statusVariant="button"
+        titleNavigationUrl={isOwner ? `${CHATROOM_URL}/${chatroomId}/edit` : ''}
+        buttonProps={{
+          buttonSize: ButtonSize.SMALL,
+          variant: ButtonVariant.WARNING,
+          iconVariant: IconVariant.LOGOUT,
+          onClick: leaveChatroom,
+        }}
```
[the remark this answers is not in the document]
Damrod: done.

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
```diff
@@ -0,0 +1,159 @@
...
+import Loading from '../../shared/components/Loading/Loading';
+import { ENTRIES_LIMIT } from '../../shared/constants';
+import {
+  Query,
+  SearchContextProvider,
+} from '../../shared/context/SearchContext';
+import { ChatControllerGetChatroomMembersRequest } from '../../shared/generated/apis/ChatApi';
+import { useNotificationContext } from '../../shared/context/NotificationContext';
+
+export default function ChatroomDetailsPage() {
```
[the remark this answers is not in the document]
Damrod: Thx!

webapp/src/shared/components/Header/Header.tsx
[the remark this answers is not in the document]
Damrod: I made it horizontal, as [we have voted](https://42born2code.slack.com/archives/C033YQ28QA0/p1668472053117479?thread_ts=1668338365.304939&cid=C033YQ28QA0).

webapp/src/shared/components/Header/Header.tsx
Damrod: Agreed with [this](https://github.com/42AGV/ft_transcendence/pull/262#discussion_r1022814381) and done

PR #273, opened 2022-10-27
#216/add a button on the users profile page to block or unblock other users

### LINKED ISSUE #216: ⚛️  Add a button on the users profile page to block or unblock other users

gbudau:
It works, but the implementation seems a bit convoluted.

Please share any ideas on how we can improve it. Thanks!

close #216


webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -54,6 +107,21 @@ const UserComponentTemplate = ({
...
+    <Button variant={ButtonVariant.SUBMIT} onClick={unblockUser}>
+      Unblock
+    </Button>
+  );
+
+  const blockUserButton = (
+    <Button variant={ButtonVariant.WARNING} onClick={blockUser}>
+      Block
+    </Button>
+  );
```
fizcamposancos, line 120: I haven't run it yet, and I haven't seen it, but wouldn't it have been better to use the toggle switch?

gbudau: The toggle switch it's one option :+1:. I will give it a try.

gbudau: Done

vdedios:
> It works, but the implementation seems a bit convoluted.
> 
> Please share any ideas on how we can improve it. Thanks!

I took serious your words but it finally wen't a bit further than expected :sweat_smile: 

I started trying to improve your implementation but ended up with a very similar approach, but realized something. We are starting to have a lot of use cases (chats owned by admins, users blocked and muted, unreaden  messages and so on) and we should try to avoid business logic in webapp. Frontend should only be used as a presentation layer, to display already modeled data from backend (or as, much, map some of it when needed)

This approach involves a little bit of business logic (that part of getting the block and evaluating if the user is auth... etc) It also implies two requests and we could calculate all that in backend :smiley: This could be a small problem now, but if we keep adding logic in webapp it will get messier in the future.

I've made a draft PR with my suggestions (a patch was too big :sweat_smile: ) Take the ideas you want from it, I will left some comments there also!

Damrod:
I think I agree with vdedios. That said, it's just a binary choice, and I don't think the problems from having some logic in the frontend will bite us very hard in the future. If you are happy with your design choice despite our opinion, feel free to merge.

gbudau:
Thanks for your feedback and ideas! Do we prefer to merge this and then apply the changes in #278?

fizcamposancos (approved):
Thanks, LGTM.

Damrod (approved):
It works, besides the comments, thanks for this feature!

gbudau:
Thanks!

PR #277, opened 2022-10-27
#220/create a database table to persist the chat room messages

### LINKED ISSUE #220: Create a database table to persist the chat room messages

gbudau:
close #220


transcendence-app/src/chat/chat-room-message-with-user.domain.ts
```diff
@@ -0,0 +1,18 @@
+import { User } from '../user/user.domain';
+import { ChatRoomMessageWithUserEntity } from './infrastructure/db/chat-room-message-with-user.entity';
+
+export class ChatRoomMessageWithUser {
+  id: string;
+  user: User;
+  chatRoomId: string;
```
Damrod, line 7: the name of this field is inconsistent with chatroom-message.domain.ts . `chatRoomId` vs `chatId`.

transcendence-app/src/user/infrastructure/db/user.entity.ts
```diff
@@ -10,16 +10,38 @@ export enum userKeys {
   AVATAR_Y = '"avatarY"',
 }
 
+type UserEntityData = {
```
Damrod, line 13: what do you need this type for?

transcendence-app/src/chat/infrastructure/db/chat-room-message.entity.ts
```diff
@@ -0,0 +1,31 @@
+export enum ChatRoomMessageKeys {
+  ID = '"id"',
+  USER_ID = '"userId"',
+  CHAT_ROOM_ID = '"chatRoomId"',
+  CONTENT = '"content"',
+  CREATED_AT = '"createdAt"',
+}
+
+export interface ChatRoomMessageEntityData {
```
Damrod, line 9: Why do you need this model of splitting XXXData from XXXEntity with the same fields, and then building Entity from Data?

Damrod:
I think we may need a bit more code to have context of how does this work.

Damrod (approved):
But I also trust you, so LGTM! Thank you!

transcendence-app/src/chat/chat-room-message-with-user.domain.ts
vdedios: I guess the chatId will never exist (individual chats will have a userId-userId relation) but agree to update it to chatRoomId so it's clearer what it refers to

transcendence-app/src/chat/chat-room-message-with-user.domain.ts
```diff
@@ -0,0 +1,18 @@
+import { User } from '../user/user.domain';
+import { ChatRoomMessageWithUserEntity } from './infrastructure/db/chat-room-message-with-user.entity';
+
+export class ChatRoomMessageWithUser {
```
vdedios, line 4: Don't see a real value of having this. The idea of the domain is to validate the minimal pieces of our problem

We already have two separate domains for User and Chat. We may do this joined models in upper layers

transcendence-app/src/chat/infrastructure/db/chat-room-message.entity.ts
vdedios: Same doubt!

transcendence-app/src/chat/chat-room-message-with-user.domain.ts
gbudau: Good catch. I've fixed it.
`chatRoomId` is a good option too. For this PR, I preferred to rename it to `chatId` to match the names we already have in the `ChatMember` table.

transcendence-app/src/user/infrastructure/db/user.entity.ts
gbudau: We use this type to instantiate the CreateMessageWithUserEntity https://github.com/42AGV/ft_transcendence/blob/c463830a0d1a7e3914c087cb826b77234afbf7eb/transcendence-app/src/chat/infrastructure/db/chat-room-message-with-user.entity.ts#L22

transcendence-app/src/chat/infrastructure/db/chat-room-message.entity.ts
gbudau: When doing `SELECT * ...`, the node-pg library returns an array of JSON-like objects https://node-postgres.com/api/result#resultrows-arrayany

If we want to instantiate an entity from an object of the pg.Result.rows[i], we need to use the constructor.

Currently, we don't use the entities in our application (we use them like objects and don't use the constructors). For example, let's take a look at the user entity https://github.com/42AGV/ft_transcendence/blob/6740ec026f3b9e6e153d32a8fa5d50a8b0181b34/transcendence-app/src/user/infrastructure/db/user.entity.ts#L13-L24

The constructor takes around nine parameters, and if we want to create an `UserEntity` instance, we would have to pass them in order.

This PR adds the `ChatRoomMessageWithUserEntity`,  and for this, we need to use the constructor https://github.com/42AGV/ft_transcendence/blob/c463830a0d1a7e3914c087cb826b77234afbf7eb/transcendence-app/src/chat/infrastructure/db/chat-room-message-with-user.entity.ts#L22

In order to type the constructor of an XXXEntity, we create the XXXData interface.

transcendence-app/src/chat/chat-room-message-with-user.domain.ts
gbudau: On the `/chatroom` page, we will display the messages from a chat room, and each message contains information about the user (username and avatar)

How do you suggest we do these joined models in the upper layers?

gbudau:
Thank you for your feedback!

transcendence-app/src/chat/chat-room-message-with-user.domain.ts
gbudau: I merge this as is to work on the other tasks that require this table. If there is a better way, we can change it later.

transcendence-app/src/chat/infrastructure/db/chat-room-message.entity.ts
vdedios: Ok, so this is more like a data type workaround :+1:

PR #278, opened 2022-10-27
[#216] Calculate blocks on backend (Proposal for gbudau PR )

vdedios:


vdedios:
(tests failing because I've changed some method signatures, didn't want to spend time on it, if you like this I will fix it ASAP)

transcendence-app/src/user/dto/user.dto.ts
```diff
@@ -30,3 +31,15 @@ export class UserDto {
...
+  username!: string;
+  email!: string;
+  fullName!: string;
+  avatarId!: string | null;
+  avatarX!: number;
+  avatarY!: number;
+  id!: string;
+  createdAt!: Date;
+  isBlocked?: boolean | null;
+}
```
vdedios, line 45: We response with a DTO, today we've added the blocked info but tomorrow we could attach other valuable information to this or other DTOs

transcendence-app/src/user/user.controller.ts
```diff
@@ -202,23 +209,6 @@ export class UserController {
...
-  async getBlock(
-    @GetUser() user: User,
-    @Param('userId', ParseUUIDPipe) blockedUserId: string,
-  ): Promise<void> {
-    const block = await this.userService.getBlock(user.id, blockedUserId);
-    if (!block) {
-      throw new NotFoundException();
-    }
-  }
-
```
vdedios, line 221: I've deleted this method because of the approach taken. I didn't see a use case where we could want to get the relation of blocked uuids, that info is tied to users and would require some later logic...

transcendence-app/src/user/user.service.ts
```diff
@@ -107,6 +118,18 @@ export class UserService {
...
+    userId: string,
+  ): Promise<boolean | null> {
+    if (userMe.id === userId) {
+      return null;
+    }
+
+    const block = await this.blockRepository.getBlock(userMe.id, userId);
+    return block?.blockedId === userId ?? false;
+  }
+
```
vdedios, line 132: And we calculate the related block logic here!

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -44,15 +44,8 @@ const UserComponentTemplate = ({
   isLoading,
 }: UserComponentTemplateProps) => {
   const { goBack } = useNavigation();
-  const { logout, authUser } = useAuth();
-  const isAuthUser =
-    authUser !== null && user !== null && authUser.id === user.id;
```
vdedios, line 49: so, we don't need to evaluate this anymore

webapp/src/shared/hooks/UseBlock.tsx
```diff
@@ -1,37 +1,18 @@
...
-                error instanceof ResponseError &&
-                error.response.status === 404
-              ) {
-                setIsBlocked(false);
-              } else {
-                throw error;
-              }
-            }),
-    [isAuthUser, user],
-  );
```
vdedios, line 30: And we don't need an extra request either

transcendence-app/src/user/user.controller.ts
Damrod: Our backend will need to be aware of the opposite situation, which is, when we actually implement the block, the authenticated user's session will need to know, not if they are blocking someone, but if someone blocked them. But obviously this shouldn't be transparent to users. And also this doesn't make your comment not true.
I don't know yet how will be implement this. But we will see.

gbudau (approved):
Thank you for your improvements! I left a few suggestions, but it looks very good to me.

transcendence-app/src/user/dto/user.dto.ts
```diff
@@ -7,7 +8,7 @@ import {
   ValidateIf,
 } from 'class-validator';
 
-export class UserDto {
+export class UserRequestDto {
```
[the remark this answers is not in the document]
vdedios: I like to differentiate between in and out DTOs so we don't mix up them so that's why I was adding the `Request` thing. I will update it to `CreateUserRequestDto` ;)

transcendence-app/src/user/user.controller.ts
vdedios: Agree, we will have to calculate that in backend

transcendence-app/src/user/dto/user.dto.ts
```diff
@@ -30,3 +31,15 @@ export class UserDto {
...
+export class UserDto {
+  username!: string;
+  email!: string;
+  fullName!: string;
+  avatarId!: string | null;
+  avatarX!: number;
+  avatarY!: number;
+  id!: string;
+  createdAt!: Date;
+  isBlocked?: boolean | null;
```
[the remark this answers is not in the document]
vdedios: No, one is for incoming data (request) the other is for output data (response)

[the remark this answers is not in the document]
vdedios: I don't think is a good idea. Let me explain:

The `User` domain represents a user in our problem which all the characteristics that makes a user (name, email, avatar etc.)

On the other hand, `isBlocked` is a status between users and do not represent nothing for the user itself. It's a relation between users.

We include it in this dto because it's **aggregated data** to communicate between layers (backend to frontend in this case) representing the `block` status between the user making the request and the fetched user itself and w

transcendence-app/src/user/user.service.ts
```diff
@@ -46,12 +47,22 @@ export class UserService {
...
+  ): Promise<UserDto | null> {
     const user = await this.userRepository.getByUsername(username);
-    return user ? new User(user) : null;
+
+    if (user && userMe) {
+      return {
+        ...user,
+        isBlocked: await this.isUserBlockedByMe(userMe, user.id),
+      };
+    }
```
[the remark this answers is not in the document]
vdedios: Fixed to return an instance! that being said:
- As explained in above comment I don't think is a good idea to add `isBlocked` to `User` domain
- I didn't add any serialization annotation because the data could come from a `User` or a `UserEntity` which they are already serialized. Not 100% sure about this btw, let me now and we can fix it in a future task

Thanks as always for your comments! :smiley:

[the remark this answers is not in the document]
vdedios: Forgot to push the commit with the change :sweat_smile: . Check -> https://github.com/42AGV/ft_transcendence/pull/288/files

Issue #276, opened 2022-10-27
Remove generated files from sources


Damrod:
We have a bunch of generated code that is a part of our repo, but should not: it's not source code. We should find a way not have to update and upload it, and have instead scripts that generate it both locally and remotely, in order to run our project upon request.

PR #279, opened 2022-10-28
#276/remove generated code

Damrod:
As detailed in #276


Makefile
```diff
@@ -0,0 +1,42 @@
...
+	$(DOCKER_COMPOSE) logs -f transcendence-app
+
+.PHONY: log-wa
+log-wa:
+	$(DOCKER_COMPOSE) logs -f webapp
+
+.PHONY: log-db
+log-db:
+	$(DOCKER_COMPOSE) logs -f db
+
```
Damrod, line 39: Maybe we could add more rules for all of the check and compiles steps in the pipeline.

Makefile
```diff
@@ -0,0 +1,42 @@
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+PROJECT_ROOT := $(dir $(MKFILE_PATH))
+TRANSCENDENCE_DEPS := $(shell find $(PROJECT_ROOT)/transcendence-app/ -type f -name "*.ts" | grep -v "node_module\|dist\|test")
```
Damrod, line 3: We could discuss which pattern brands a file that the generated files will depend on. Maybe more accurately than my first guess

transcendence-app/swagger.nest-cli.json
```diff
@@ -0,0 +1,19 @@
...
+  "entryFile": "swagger-main",
+  "compilerOptions": {
+    "assets": [{"include": "**/*.json", "outDir": "./dist/src"}],
+    "plugins": [
+      {
+        "name": "@nestjs/swagger",
+        "options": {
+          "classValidatorShim": true,
+          "introspectComments": true,
+          "dtoFileNameSuffix": [".dto.ts", ".entity.ts", ".domain.ts"]
```
Damrod, line 14: This pattern possibly

scripts/generate-openapi.sh
```diff
@@ -0,0 +1,17 @@
...
+PROJECT_ROOT="${SCRIPT_DIR}"/..
+SWAGGER_SPEC_FILE="${PROJECT_ROOT}/transcendence-app/swagger-spec.yaml"
+mkdir -p "${PROJECT_ROOT}"/webapp/src/shared/generated
+ME="$(whoami)"
+DOCKER_COMPOSE="$("${PROJECT_ROOT}"/scripts/get-docker-compose.sh)"
+cd "${PROJECT_ROOT}" && touch "${SWAGGER_SPEC_FILE}"
+{ { cd "${SCRIPT_DIR}" || exit ; } &&
+${DOCKER_COMPOSE} run --rm swagger && ${DOCKER_COMPOSE} down -v ; } ||
+  rm "${SWAGGER_SPEC_FILE}"
+docker run --rm -e MY_USER="${ME}" \
```
Damrod, line 16: Maybe we would like this part of the script in a way that it can be ran independently from the generation of the yaml file, to execute manually when hot reload updates the source

scripts/generate-cmd.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -eux
+docker-entrypoint.sh generate -i /local/transcendence-app/swagger-spec.yaml \
+  -g typescript-fetch \
+  -o /local/webapp/src/shared/generated
+useradd -U -ms /bin/bash "${MY_USER}"
```
Damrod, line 6: Maybe this will quit with error if $MY_USER already exists?

Makefile
vdedios: The idea of this env is something like ensuring we already have the deps installed and the project compiled?

Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+.PHONY: log-tr
+log-tr:
+	$(DOCKER_COMPOSE) logs -f transcendence-app
+
+.PHONY: log-wa
+log-wa:
+	$(DOCKER_COMPOSE) logs -f webapp
+
+.PHONY: log-db
```
vdedios, line 36: What do you think about adding some documentation to the README.md on how to build the project with this new makfile?

scripts/generate-cmd.sh
vdedios: what do you think about using `$USER` instead? you'll always have the current user assigned to the env

scripts/get-docker-compose.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -eu
+which 'docker-compose' > /dev/null 2>&1
+NO_HYPHEN=$?
```
vdedios, line 4: what is the idea of this?

scripts/generate-openapi.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -ux
+SCRIPT_DIR="$(
+  cd -- "$(dirname "$0")" >/dev/null 2>&1 || exit
+  pwd -P
+)"
+PROJECT_ROOT="${SCRIPT_DIR}"/..
+SWAGGER_SPEC_FILE="${PROJECT_ROOT}/transcendence-app/swagger-spec.yaml"
+mkdir -p "${PROJECT_ROOT}"/webapp/src/shared/generated
+ME="$(whoami)"
```
vdedios, line 10: what do you think about using $USER instead? you'll always have the current user assigned to the env

scripts/generate-cmd.sh
vdedios: Maybe getting the user running in this container and adding a condition to skip the `chown` if the match?
https://stackoverflow.com/questions/14810684/check-whether-a-user-exists

scripts/generate-openapi.sh
vdedios: I think is ok as is. If we need to reload the autogenerated interfaces `make gen` seems enough

transcendence-app/src/main.ts
```diff
@@ -22,8 +20,6 @@ async function bootstrap() {
     .setVersion('1.0')
     .build();
   const document = SwaggerModule.createDocument(app, options);
-  const yamlString: string = yaml.stringify(document, {});
```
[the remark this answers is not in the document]
vdedios: not sure about how this worked, but agree I would like to keep hot reload

transcendence-app/src/swagger-main.ts
```diff
@@ -0,0 +1,27 @@
+import { VersioningType } from '@nestjs/common';
+import { NestFactory } from '@nestjs/core';
+import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
+import { writeFileSync } from 'fs';
+import { AppModule } from './app.module';
+import * as yaml from 'yaml';
+import { setupApp } from './setup-app';
+
+async function bootstrap() {
```
vdedios, line 9: Why do we need to detach the swagger generation from the compilation of the transcendence app?

I was thinking if we could just keep the swagger generation on build (or development mode) and just add the generated `.yaml` to the `.gitignore` as you've done

vdedios (approved):
Great job :rocket: !

I missed the good old Makefile, I think is very useful and will help us since we now have many workflows.

On the other hand I have doubts about detaching the swagger generation, leaving this for now as a comment to discuss it further

Makefile
vdedios: Yes, I miss one for seeding data :+1:

Makefile
Damrod: No, this is good old makefile prerequisites. If you attempt to run the transcendence-app/swagger-spec.yaml rule to generate the file of the same name, but the system sees the file and does not detect changes in the prerequisites, it will not run it.

Makefile
Damrod: Yes, I definitely have to do that. I agree.

scripts/generate-openapi.sh
Damrod: Yes, but this $MY_USER would be the user that ran the docker run, passed as an env variable from the docker run command, not the user that is running the commands inside the docker container. The idea is: normally the user inside the container is root or sudoer. If it creates files/folders in a volume it shares with the parent system, and then quits execution, the generated files will be owned by root, and you won't be able to update or delete them without sudo, which is inconvenient. This is a well known docker problem, and this code segment is a workaround.

scripts/get-docker-compose.sh
Damrod: There are two names for "docker compose". One is "docker compose" with a space, as a docker subcommand, and the other one, "docker-compose" with a dash. I personally can't run "docker compose", I need the dash "docker-compose", but I know some of our colleagues can't use the dash. I guess I could do an alias, but in my opinion this is more solid and portable. The dash will solve my problem, whereas this will solve everyone's problem.

transcendence-app/src/swagger-main.ts
Damrod: I guess we could dissassemble the root docker-compose and have a dependency on the webapp service delay it's booting up until the swagger-spec.yaml AND the rest of the generated files have been created by the transcendence-app service + maybe some intermediate service.

This solution seemed a bit simpler, and more modular, so this is how we decided to proceed.

scripts/generate-cmd.sh
Damrod: Possibly. Will check.

transcendence-app/src/main.ts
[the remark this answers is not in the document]
Damrod: The removal of these two lines wouldn't remove hot reload: it would only remove the generation of the swagger spec from hot reload. But still.

Makefile
Damrod: done

Makefile
Damrod: done

transcendence-app/src/swagger-main.ts
Damrod: Let's discuss this in a different issue, please.

scripts/generate-openapi.sh
Damrod: I still added functionality to the script to ba able to run either separately.

Damrod:
You can see the updated readme [here](https://github.com/42AGV/ft_transcendence/tree/%23276/remove_generated_code)

Makefile
```diff
@@ -0,0 +1,54 @@
...
+log-db:
+	$(DOCKER_COMPOSE) logs -f db
+
+.PHONY: get-ip
+get-ip:
+	@docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ft_transcendence-webapp-1
+
```
fizcamposancos, line 48: I don't know why, or how can I change it the name, but in my computer ft_transcendence-webapp-1 is ft_transcendence_webapp_1 so this rule doesn't work unless I manually modify the makefile. Any idea how can I fix this?
And I sometimes use the ip of the db to connect to db manager so would be nice to have something like this: 
```
@echo "webapp ip:"
@docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ft_transcendence_webapp_1
@echo "transcendence-app ip:"
@docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ft_transcendence_transcendence-app_1
@echo "db ip:"
@docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ft_transcendence_db_1
```

fizcamposancos (approved):
Awesome work!! I would need some extra explanation about some issues you've been discussing and I left a comment of a problem I found, but maybe it's only me.

Makefile
Damrod: this should be fixed now, and be agnostic to your container name flavor

PR #288, opened 2022-10-29
#216 - Return an instance and not a plain js object

vdedios:
I've made this comment (https://github.com/42AGV/ft_transcendence/pull/278#discussion_r1008697618) but forgot to push the commit before merging the PR #278 :sweat_smile:


Damrod:
please, erase this one https://github.com/42AGV/ft_transcendence/blob/main/transcendence-app/swagger-spec.yaml~da6da06...%20calculate%20user%20bussiness%20logic%20in%20backend as well

Damrod (approved):

PR #289, opened 2022-10-29
removed some bugs from removal of generated code PR

### LINKED ISSUE #276: Remove generated files from sources

Damrod:
Small script changes to fix #276


.gitignore
```diff
@@ -15,7 +15,7 @@ dist
 
 # generated code
 transcendence-app/swagger-spec.yaml
-webapp/src/shared/generated/*
+webapp/src/shared/generated
```
gbudau, line 18: In the subject, there is this line:
```
Everything should run with a single call to docker-compose up –build
```
Currently, we generate first the code for the webapp, then do docker-compose up --build. I'm concerned that this might affect our evaluation.

I added this here because I didn't have time to review the previous PR.

gbudau (approved):
Thank you for this feature. LGTM :rocket:

.gitignore
Damrod: I'm well aware, but I guess I should open a ticket to leave a pending action point.
Two choices come to mind and some other questions about how does this observation impact what we have had so far as well.
One choice is to work like this and just before we close our project, hardcode the generated files the way we used to.
The other is to refactor the way we generate files, to have instead some intermediate service, that does all the the WebApp generating work, mounting both the swagger spec, and the WebApps/shared/generated folders. I think it would be less optimal then the makefile, but it would be in line with the project's requirements. 
Food for thought.

gbudau: Thanks. I prefer the first choice.
Regarding the .env file, there is a Slack message about this https://42born2code.slack.com/archives/CN3D33GRX/p1655809254653079

PR #293, opened 2022-10-30
Unify chatroom and postgres naming

### LINKED ISSUE #268: Update Postgres repositories files naming convention
[As commented here](https://github.com/42AGV/ft_transcendence/pull/245#discussion_r1004580486)

### LINKED ISSUE #292: Fix naming for chatroom
I am working in one-to-one message feature and I am finding a lot of trouble. We still have lots of things named as `chat` instead of `chatroom`. Also, we need a better folder structure for `chat` module because files start to be a little bit mixed up and we'll be adding more and more.

I've proposed the following conventions:

- chat room joined to one word, `chatroom`. This way we'll have unified names between dashed/camelCase naming (`chat-room` and `chatRoom` vs single `chatroom`)
- all code related to `chatroom` in `chat` module, inside `chatroom` folder. And the same will occur for `one-to-one` chat

vdedios:
closes #292
closes #268

Please, read before linked issues.

:warning: This 99% will cause you conflicts, we may merge before all work you have in progress :warning:


gbudau (approved):
Thank you! It looks more consistent. LGTM :rocket:

fizcamposancos (commented):
One question. You've changed `chatRoom` with `chatroom` and `ChatRoom` with `Chatroom`.
But chat room are two words, aren't they? So camel-case would be chatRoom

fizcamposancos (approved):
For me it's ok

Damrod:
> One question. You've changed `chatRoom` with `chatroom` and `ChatRoom` with `Chatroom`. But chat room are two words, aren't they? So camel-case would be chatRoom

I have been researching this, because I thought just like you, but it turns out chatroom as one word is also acceptable.

I wouldn't agree that it's worth changing the convention we have so far for the case (because it's not wrong), and also all spell checkers I have, sensitive to camel case, will complain if we assume that is just one word. But the new convention, in fact, would not be wrong.

Damrod (approved):
Please, merge this so we can unblock development in the backend.

PR #294, opened 2022-10-30
#223/create a get endpoint to get the messages from a chat room

### LINKED ISSUE #221: Create a database method to add a chat room message

### LINKED ISSUE #222: Create a database method to get the messages from a chat room

### LINKED ISSUE #223: Create a GET endpoint to get the messages from a chat room

Details: Create a GET /chat/room/{roomId}/messages endpoint to get the messages from a chat room. _This endpoint must be protected_

gbudau:
close #221
close #222
close #223


transcendence-app/src/chat/chat.controller.ts
```diff
@@ -130,10 +133,44 @@ export class ChatController {
...
+  ): Promise<ChatRoomMessageWithUser[] | null> {
+    const chatroom = await this.chatService.getChatroomById(chatroomId);
+    if (!chatroom) {
+      throw new NotFoundException();
+    }
+    const isChatMember = await this.chatMemberService.getById(
+      chatroomId,
+      user.id,
+    );
+    if (!isChatMember) {
```
Damrod, line 164: maybe this can also be in L 122?

Damrod:
It's a bit hard to test this without first seeding the chatroommessage table. How do you do it?

transcendence-app/src/shared/dtos/pagination-query.dto.ts
```diff
@@ -0,0 +1,19 @@
+import { ApiPropertyOptional } from '@nestjs/swagger';
```
Damrod, line 1: Thank you for this feature!

transcendence-app/src/chat/chat.controller.ts
gbudau: :+1: done

transcendence-app/src/chat/infrastructure/db/chatmember.repository.ts
```diff
@@ -1,7 +1,12 @@
-import { ChatMember, ChatMemberWithUser } from '../../chatmember.domain';
+import { ChatMemberWithUser } from '../../chatmember.domain';
+import { ChatMemberEntity } from './chatmember.entity';
 
 export abstract class IChatMemberRepository {
-  abstract add(chatMember: ChatMember): Promise<ChatMember | null>;
+  abstract add(chatMember: ChatMemberEntity): Promise<ChatMemberEntity | null>;
```
Damrod, line 5: In user, the add function takes a User (from domain) and returns a Promise<User|null>. Why here you change this to use entity instead?

gbudau:
What I did was a bit of manual setup.

1. Create a chatroom.
2. Login in with one user and join the chatroom by making a request to http://localhost:3000/api#/chat/ChatController_createChatRoomMember with the chatroom ID.
3. Login in with another user and join the same chatroom 
4. Insert chatroom messages into the database using the chatroom ID and user IDs.

```sql
INSERT INTO
  ChatRoomMessage ("chatId", "userId", "content")
VALUES
  (
    'your-chatroom-id',
    'your-user-id',
    'message content'
  );
```

transcendence-app/src/chat/infrastructure/db/chatmember.repository.ts
gbudau: We might want to make some changes to our repository methods.
In the current implementation, we return an object with the same properties as the entity/domain. They seem interchangeable.
We could return an entity instance, which we could serialize using `ClassSerializerInterceptor`.

gbudau: I've created a new task for this #301

fizcamposancos (approved):
LGTM!
One general question, since the controller methods are getting bigger, calling the service, doing some logic and then throwing exceptions or the response. What is better, a simpler controller and the logic in the service or a simpler service and the controller just catching error or the response?

gbudau:
I don't know. I prefer to throw the HTTP exceptions from the controllers when I can because it's easier to document the endpoints, and maybe we want to use the services with other protocols, like WebSocket.
But I think the logic should be in the service as much as possible.

gbudau:
Thanks!

transcendence-app/src/chat/infrastructure/db/chatmember.repository.ts
gbudau: I made a POC with repositories that return entities instances and no domains #303

PR #295, opened 2022-10-31
#252/use seed files to fill database on dev deployment and improve them

### LINKED ISSUE #252: Use seed files to fill database on dev deployment and improve them

vdedios:
closes #252


README.md
```diff
@@ -200,7 +200,7 @@ npx knex seed:make seed_name
 ```
 or its make counterpart, from the root directory
 ```shell
-SEED=<your_seed> && make seed
+SEED=<your_seed> && make seed-make
```
Damrod, line 203: we're missing an "export" in front of this, i.e.
```
export SEED=<your_seed> && make seed-make
```

Damrod:
TBH I don't get how is `seed-make` supposed to work. It creates an empty seed file with an appropriate name, but with placeholders, that cannot be ran later with `seed-run`.
We're supposed to write a <database_name>-seed.ts file with rules and directions on how to fill a particular table, isn't it? Why do we need to create a an empty seed file full of place holders? What am I missing?

gbudau:
The seed-make creates a seed file boilerplate, which we can edit as we want https://knexjs.org/guide/migrations.html#seed-files

db/setup/queries.sql
```diff
@@ -1,97 +0,0 @@
...
-  Users ("username", "email", "fullName")
-VALUES
-  (
-    'vde-dios',
-    'vde-dios@transcendence.com',
-    'vde-dios'
-  );
-
-INSERT INTO
-  ChatRoom ("name", "ownerId")
```
Damrod, line 84: Non blocking, but we have to seed all other tables we have. We effectively are not seeding chatroom tables anymore, after deleting this file.

Damrod (approved):
LGTM!

README.md
gbudau: Thanks! Done

db/setup/queries.sql
gbudau: Thanks! I've created a new task for this #300

Damrod:
> The seed-make creates a seed file boilerplate, which we can edit as we want. For example, instead of having only one seed file, we might want to create multiple files, one for the users, and others for the chatrooms, chat members, chat messages, etc.

A boilerplate that it does not customize in any way. In practical use, this is no different from having a xxx_seed.ts with the following content:

<details>
import { Knex } from 'knex';

export async function seed(knex: Knex): Promise<void> {
  // Deletes ALL existing entries
  await knex('chatroom').del();

  // Inserts seed entries
  await knex('chatroom').insert([
    { id: 1, colName: 'rowValue1' },
    { id: 2, colName: 'rowValue2' },
    { id: 3, colName: 'rowValue3' },
  ]);
}
</details>

And copying and pasting it with a different, appropriate name.

Is it like this? I'm asking over and over because I wish to believe that I'm missing something. Otherwise, it feels like a large overkill to run a specialized command when the thing that it's doing is just a `cp <yourfavouriteroutfortemplate>/seedtemplate seeds/chat_file.ts`

db/setup/queries.sql
vdedios: I'd say we may add a task like this one in every user-story which needs stored data from data-base. I'll add one also for the 1:1 chat US I'm working on

vdedios:
> A boilerplate that it does not customize in any way. In practical use, this is no different from having a xxx_seed.ts file with the following content: (...)

Agree, we should use Makefile as a deployment tool

vdedios:
I've missed some leftovers :sweat_smile: https://github.com/42AGV/ft_transcendence/blob/fd8a4b28de823120bd9a91a96c5a39647ffc3dfa/Makefile#L64-L68

Will fixed when merging next PR, sorry!

Damrod:
I don't really mind to have the additional rules in the makefile. It's just a tool for all of us to use however we want, and if it makes anyone's life just a bit simpler to have it, then it should stay there.
I was just  trying to understand the tool. Sorry to have been so nagging.

PR #303, opened 2022-11-02
#301/return entity instances from repositories

### LINKED ISSUE #301: Return entity instances from repositories
In the current implementation, we return an object with the same properties as the entity/domain. They seem interchangeable.
We could return an entity instance, which we could serialize using the `ClassSerializerInterceptor`.

gbudau:
close #301


Damrod:
please, update scripts/get-swagger-spec-dependencies.sh L17

Damrod:
Screenshot from 2022-11-03 11-30-43

Why don't we annotate the default avatarX and avatarY in the chatroom anymore?

gbudau:
> Why don't we annotate the default avatarX and avatarY in the chatroom anymore?

Thanks. I've fixed it.

Damrod:
More differences:

Screenshot from 2022-11-03 12-43-12
Screenshot from 2022-11-03 12-41-58
(JoinedAt debería ser nullable:true)
Screenshot from 2022-11-03 12-41-31

gbudau:
Thanks, good catch. I've updated it.

Damrod (approved):
LGTM, Thanks!

fizcamposancos (approved):
LGTM. I would say I like it more now. And I think this way we take advantage of Nestjs features because, TBH, I didn´t see the point of having a domain and a entity which seem interchangable.
Very nice work.

transcendence-app/src/auth/auth-provider/infrastructure/db/auth-provider.entity.ts
```diff
@@ -6,10 +6,20 @@ export enum AuthProviderKeys {
...
+export class AuthProvider {
+  provider: AuthProviderType;
+  providerId: string;
+  userId: string;
+
+  constructor(authProviderData: AuthProviderData) {
+    this.provider = authProviderData.provider;
+    this.providerId = authProviderData.providerId;
+    this.userId = authProviderData.userId;
+  }
```
vdedios, line 24: Maybe we could avoid the `*Data` interfaces doing this:

```suggestion
export class AuthProvider {
  provider: AuthProviderType;
  providerId: string;
  userId: string;

  constructor(authProvider: AuthProvider) {
    this.provider = authProvider.provider;
    this.providerId = authProvider.providerId;
    this.userId = authProvider.userId;
  }
```
typescript will correctly infere all fields

transcendence-app/src/user/infrastructure/db/user.entity.ts
```diff
@@ -10,30 +13,31 @@ export enum userKeys {
...
-  avatarId: string | null;
+  avatarId: string;
   avatarX: number;
   avatarY: number;
   createdAt: Date;
-};
+}
 
-export class UserEntity {
+export class User {
```
vdedios, line 28: I think is a good idea to keep the `*Entity` denomination, this way it's clear is data coming from the db

transcendence-app/src/user/infrastructure/db/user.entity.ts
```diff
@@ -10,30 +13,31 @@ export enum userKeys {
...
   id: string;
   username: string;
   email: string;
   fullName: string;
+  @Exclude()
```
vdedios, line 33: I'd say all serialization may be done next to network i/o, this is, in the controllers layer as [NestJS documentations says](https://docs.nestjs.com/techniques/serialization). This entity is directly passed to some controllers but, in other cases, wrapped around a dto. In that case we'd have to serialize in two places if need to make some serialization inside de dto.

We may be still mixing a bit our approach. What do you think about unifiying like this?

- Data we **return**:
  - entities are plain data returned from DB, not serialization or logic here, just data.
  - any needed bussiness logic, mapping etc. is done in services.
  - we finally instance a dto (which has all needed serialization)
  So, with this idea in mind, we could re-use the `UserDto` for all methods returning plain users and move all serialization there.
  
- Data we **recieve**:
  - we serialize the incoming dto (we are already doing this, so good :+1: )

transcendence-app/src/user/user.controller.spec.ts
```diff
@@ -1,31 +1,29 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { AvatarFileInterceptor, UserController } from './user.controller';
 import { UserService } from './user.service';
-import { CreateUserRequestDto } from './dto/user.dto';
+import { CreateUserDto } from './dto/create-user.dto';
```
vdedios, line 4: We could have a denomination for incoming/outcoming dtos. I was thinking in something like `*RequestDto`, `*ResponseDto` you've deleted :sweat_smile: Why don't you like it?

webapp/src/index.tsx
```diff
@@ -91,7 +92,7 @@ root.render(
             path={USER_ME_URL}
             element={
               <RequireAuth>
-                <UserPage displayAsAuthUser />
+                <AuthUserPage />
```
vdedios, line 95: I don't know why we didn't do this before :smile:

vdedios (commented):
Nice, thanks for working on this :+1: domain models seems at this point not needed

Most comments I left on specific things are indeed general things that apply to similar uncommented things

transcendence-app/src/user/infrastructure/db/user.entity.ts
vdedios: Discussed by call: could be but a lot of work changing things on frontend, so no worth it

transcendence-app/src/user/infrastructure/db/user.entity.ts
vdedios: Talked by call: we have arrived to an intermediate solution, please see message posted in Slack (in a few minutes)

transcendence-app/src/auth/auth-provider/infrastructure/db/auth-provider.entity.ts
vdedios: Discussed by call: some of these classes (like the ChatroomMemberWithUser) need this data because the fields returned by the db collapse. So, despite not neede here, I'll keep them for consistency.

transcendence-app/src/user/user.controller.spec.ts
vdedios: Discussed by call: this will apply for Response only since, if done for requests also, we'd have many changes in webapp also

gbudau:
Thank you all :heart:

Issue #226, opened 2022-10-19
⚛️ Create a chat room page


Damrod:
A part of the user story: _As a chat room member, I want to be able to send messages and receive messages from other members in real time_

PR #304, opened 2022-11-03
#224/emit a chat room message event from the client to the server to add a message to a chat room

### LINKED ISSUE #217: Create a database method to upsert a member in a chat room
As a user, I want to be able to a leave chat room.

### LINKED ISSUE #224: Emit a chatRoomMessage event from the client to the server to add a message to a chat room

### LINKED ISSUE #225: Emit a chatRoomMessage event from the server to a chat room to send a message to the clients connected to that chat room

gbudau:
This PR connects the chatroom with the database.

How does this works:

Before sending and receiving messages to/from a chatroom, the user has to join the chatroom by doing an HTTP POST request to [`/chat/room/{chatroomId}/members` ](http://localhost:3000/api#/chat/ChatController_createChatroomMember) endpoint.

After that, they can emit `joinChatroom`, and `leaveChatroom` WebSocket events to join or leave chatrooms as clients (they should be chatroom members before emitting these events, these events are for real-time communication between the connected chatroom members, not for joining or leaving a chatroom as a member)

And they can send and receive messages to/from other chatroom members by emitting and subscribing to the `chatroomMessage` event.

close #217
close #224
close #225

After this change, the chatroom page displays the chatroom ID instead of its name and doesn't load the previous messages.
We will address this in #226

Other changes:
After creating a chatroom, the owner becomes a chatroom member automatically. 
When joining a chatroom, we check if the room has a password and compare it with the password supplied by the user.


transcendence-app/src/shared/filters/bad-request-transformation.filter.ts
```diff
@@ -0,0 +1,10 @@
+import { ArgumentsHost, BadRequestException, Catch } from '@nestjs/common';
+import { BaseWsExceptionFilter, WsException } from '@nestjs/websockets';
+
+@Catch(BadRequestException)
+export class BadRequestTransformationFilter extends BaseWsExceptionFilter {
```
gbudau, line 5: We use this to handle exceptions thrown by the validation pipes in the WebSocket context https://stackoverflow.com/a/63396944

transcendence-app/src/shared/types/custom.d.ts
```diff
@@ -3,7 +3,7 @@ import { User } from '../../user/user.domain';
 
 declare module 'http' {
   interface IncomingMessage {
-    user: User | null;
+    user: User;
```
gbudau, line 6: The user can't be `null` because the [passport](https://www.passportjs.org/concepts/authentication/downloads/html/#middleware) set it to the user object and we have the [WsAuthenticatedGuard](https://github.com/42AGV/ft_transcendence/blob/4092458a15338b76fafa1eec0a60ddc4cad6f7e7/transcendence-app/src/shared/guards/ws-authenticated.guard.ts#L8) which checks if the user is authenticated or throws an exception if not.

Damrod:
I see the messages are persisted, and the GET to http://localhost:3000/api/v1/chat/chat/room/{chatRoomId}/messages works, but I see that, in the frontend, upon leaving a room and returning to it, the previous messages are not collected/the page is not refreshed with old data. Before we had those messages if the server was continuously running, (if we restarted the server, the messages were lost, which is, I assume, what we're fixing with the database).

I assume you are aware of this. Where do we plan to return this functionality?

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -9,75 +11,72 @@ import {
...
   @SubscribeMessage('chatroomMessage')
-  handleMessage(
-    @MessageBody() data: { chatroomId: ChatroomId; content: string },
+  async handleMessage(
+    @MessageBody()
+    createChatroomMessageDto: CreateChatroomMessageDto,
     @ConnectedSocket() client: Socket,
   ) {
+    const { chatroomId, content } = createChatroomMessageDto;
     const user = client.request.user;
```
Damrod, line 42: why cant we use the @GetUser injectable for this?

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -9,75 +11,72 @@ import {
...
-        chatroomId: chatroomId,
-      };
-      chatroom.add(message);
-      this.server.to(chatroomId).emit('chatroomMessage', message);
+    const chatroomMember = await this.chatroomMemberService.getById(
+      chatroomId,
+      user.id,
+    );
+    if (!chatroomMember || chatroomMember.muted) {
+      throw new WsException('Forbidden');
```
Damrod, line 48: We have to remember to n ot silently fail. We have to somehow prevent this situation from normally ever happening when navigating the site.
From the scale:
> [...] there must be no unhandled warning or error

Damrod:
You should link this issue as well #217

transcendence-app/src/chat/chat.gateway.ts
gbudau: I think the `GetUser` decorator will not work in the WebSocket context, but we could create another one :+1:.
Thanks, I will give it a try.

gbudau:
> You should link this issue as well #217

Thanks. Done

transcendence-app/src/shared/db/postgres/utils.ts
```diff
@@ -36,6 +36,29 @@ export const makeQuery = async <T>(
...
+  table: table,
+  entity: T,
+) {
+  const { cols, params, values } = entityQueryMapper(entity);
+  const text = `INSERT INTO ${table}(${cols.join(', ')}) VALUES (${params.join(
+    ',',
+  )}) RETURNING *;`;
+  return client.query(text, values);
+}
+export function updateByIdWithClient<T extends Record<string, any>>(
```
Damrod, line 50: Why are these better here than in base postgres repository?

Damrod (approved):
I need to review better the queries, but so far, LGTM. Thanks for this feature!

transcendence-app/src/chat/chat.gateway.ts
gbudau: Updated. Thanks for the suggestion.

transcendence-app/src/chat/chat.gateway.ts
gbudau: Update: I've tried to create a decorator for this and could not get it working in the WebSocket context :pensive:

This is what I tried:

<details>

```typescript
import { createParamDecorator, ExecutionContext } from '@nestjs/common';

export const UserWs = createParamDecorator(
  (property: string, ctx: ExecutionContext) => {
    const client = ctx.switchToWs().getClient();
    const user = client.request.user;

    return property ? user?.[property] : user;
  },
);
```

</details>

transcendence-app/src/shared/db/postgres/utils.ts
gbudau: I think it's better here because the T type could be any entity, while the base repository is always the same entity.

We could add another type to the base repository <T, U>, but I think it would make it less readable.

vdedios (commented):
Great work!! I've left some doubts and things I think may be improved

transcendence-app/src/shared/db/postgres/utils.ts
vdedios: Check comment above, still think this is not a util

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -9,75 +11,72 @@ import {
...
-      client.emit('chatroomMessages', [...messages]);
+    const message = await this.chatService.addChatroomMessage({
+      chatroomId,
+      userId: user.id,
+      content,
+    });
+    if (message) {
+      this.server.to(chatroomId).emit('chatroomMessage', { ...message, user });
+    } else {
+      throw new WsException('Service Unavailable');
```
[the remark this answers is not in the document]
gbudau: We could add some notifications about this in webapp :+1:

transcendence-app/src/chat/chat.service.ts
```diff
@@ -106,4 +116,33 @@ export class ChatService {
...
+  }
+
+  async addChatroomMember(
+    chatroomId: string,
+    userId: string,
+    joinChatroomDto: JoinChatroomDto,
+  ): Promise<ChatroomMember | null> {
+    const foundChatroom = await this.getChatroomById(chatroomId);
+    if (!foundChatroom) {
+      throw new NotFoundException('Chatroom not found');
```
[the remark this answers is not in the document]
gbudau: We use this method in the http://localhost:3000/api#/chat/ChatController_createChatroomMember endpoint, which is HTTP.
The client could check the status and the body of the response.

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -9,75 +11,72 @@ import {
...
-    @ConnectedSocket() client: Socket,
-  ) {
-    const messages = this.chatrooms.get(chatroomId);
-    if (messages) {
-      client.emit('chatroomMessages', [...messages]);
+    const message = await this.chatService.addChatroomMessage({
+      chatroomId,
+      userId: user.id,
+      content,
+    });
```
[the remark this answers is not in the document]
gbudau: Thanks!

transcendence-app/src/shared/db/postgres/utils.ts
```diff
@@ -36,6 +36,29 @@ export const makeQuery = async <T>(
...
+export function insertWithClient<T extends Record<string, any>>(
+  client: PoolClient,
+  table: table,
+  entity: T,
+) {
+  const { cols, params, values } = entityQueryMapper(entity);
+  const text = `INSERT INTO ${table}(${cols.join(', ')}) VALUES (${params.join(
+    ',',
+  )}) RETURNING *;`;
+  return client.query(text, values);
```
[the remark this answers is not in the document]
gbudau: Thanks for your suggestions :+1: 
I've opened a new task and PR to address them, #309 and #310

transcendence-app/src/shared/db/postgres/utils.ts
gbudau: Thanks, done in #310

PR #307, opened 2022-11-05
#306 add messaging button to user page

### LINKED ISSUE #306: Add chat button to user profile
Add button to user profile in order to enable one-to-one messaging.

Block relation must be handled:
- If I block user, I cannot access chat (subject says, if I block a user I cannot see his/her messages anymore)
- If a user blocks me, I cannot access chat
- If I don't have any block relations (for example when visiting my own profile) I don't have either block or messaging features available

vdedios:
closes #306 

:warning: read before #306 !


transcendence-app/src/chat/chat.controller.ts
```diff
@@ -190,7 +190,7 @@ export class ChatController {
     type: [ChatMessage],
   })
   @ApiBadRequestResponse({ description: 'Bad Request' })
-  @ApiUnprocessableEntityResponse({ description: 'Unprocessable entity' })
+  @ApiServiceUnavailableResponse({ description: 'Service unavailable' })
```
vdedios, line 193: As @gbudau suggested in previous PR ;)

transcendence-app/src/user/dto/user.response.dto.ts
```diff
@@ -1,10 +1,14 @@
 import { User } from '../infrastructure/db/user.entity';
 
+export class UserBlockRelation {
+  isUserBlockedByMe!: boolean;
+  amIBlockedByUser!: boolean;
+}
 export class UserResponseDto extends User {
-  isBlocked: boolean | null;
+  blockRelation: UserBlockRelation | null;
```
vdedios, line 8: since if a user has blocked me is also meaningful, we should get this state in both directions:
a) the user is blocked by me
b) the user has blocked me

transcendence-app/src/user/dto/user.response.dto.ts
```diff
@@ -1,10 +1,14 @@
 import { User } from '../infrastructure/db/user.entity';
 
+export class UserBlockRelation {
+  isUserBlockedByMe!: boolean;
+  amIBlockedByUser!: boolean;
+}
```
vdedios, line 6: since if a user has blocked me is also meaningful, we should get this state in both directions:
a) the user is blocked by me
b) the user has blocked me

webapp/src/shared/hooks/UseBlock.tsx
```diff
@@ -1,24 +1,69 @@
...
+      ...state,
+      isUserBlocked: false,
+    };
+  }
+  throw new Error('Unknown block reducer action');
+};
+
 export function useBlock(user: UserResponseDto | null) {
-  const [isBlocked, setIsBlocked] = useState<boolean | null>(null);
+  const [blockRelation, dispatch] = useReducer(blockReducerCallback, null);
```
vdedios, line 47: Since we start to have a more complex way to handle this state, better to use a semantic dispatch

webapp/src/shared/hooks/UseBlock.tsx
```diff
@@ -1,24 +1,69 @@
...
+    return {
+      ...state,
+      isUserBlocked: true,
+    };
+  } else if (type === 'UNBLOCK_USER') {
+    return {
+      ...state,
+      isUserBlocked: false,
+    };
+  }
```
vdedios, line 42: This may be a bit cumbersome, but good enough for me. The idea is treat `null` as a special case. Only valid to add data if `null` when filling all the state, otherwise it means we don't have a valid blockRelation

transcendence-app/src/chat/chat.controller.ts
gbudau: Thanks

gbudau (approved):

fizcamposancos (commented):
I am not sure if we need the complete block relation and all the logic in the modified useBlock.
We only need to know there is an entry with user 1 and user 2 in the block table (in any order), to enable/disable messages, and who is the blocker to set the toggle switch accordingly.
So I think this could be much simpler.
I am going to think about a little bit more.
Anyway, now it works perfectly. Very good job.

fizcamposancos (approved):
After thinking about it, I think this PR could be simplidied to strictly achieve the subject requirements but right now we have a solution more complete, functional and that can have other benefits in the future.
Well done

vdedios:
> After thinking about it, I think this PR could be simplidied to strictly achieve the subject requirements but right now we have a solution more complete, functional and that can have other benefits in the future.
> Well done

Yes, this may be a bit more from what subject asks for, but I though having a mutual blocking feature was nice and not that much extra work. Thanks!

PR #302, opened 2022-11-02
#282 / chat one-to-one backend

vdedios:
close #282
close #283
close #284

PR #308, opened 2022-11-06
#203/create endpoint and db method to update chatroom

### LINKED ISSUE #202: Create a database method to update the details of a chat room
The owner can update the chat room name, avatar, or password.

### LINKED ISSUE #203: Create an endpoint to update a chat room
Create a `PATCH /chat/room/{roomId}` endpoint to update a chat room
The owner can update the chat room name, avatar, or password.
We must encrypt the password before storing it in the database.
This endpoint must be protected.

fizcamposancos:
close #203 
close #202


transcendence-app/src/chat/chat.controller.ts
```diff
@@ -184,6 +186,23 @@ export class ChatController {
...
+    @Param('chatroomId', ParseUUIDPipe) chatroomId: string,
+    @Body() updateChatroomDto: UpdateChatroomDto,
+  ): Promise<Chatroom> {
+    const updatedChatroom = await this.chatService.updateChatroom(
+      chatroomId,
+      updateChatroomDto,
+    );
+    if (!updatedChatroom) {
+      throw new UnprocessableEntityException();
+    }
```
vdedios, line 202: I think this should better a `ServiceUnavailableException`. Check the [conversation](https://github.com/42AGV/ft_transcendence/pull/302#discussion_r1012830877) I had with @gbudau

transcendence-app/src/chat/chatroom/dto/update-chatroom.dto.ts
```diff
@@ -1,18 +1,22 @@
...
   @IsNotEmpty()
-  @Transform(({ value }: TransformFnParams) => value?.trim())
   @IsOptional()
+  @Transform(({ value }: TransformFnParams) => value?.trim())
   name?: string;
 
-  @IsInt()
+  @IsString()
   @IsOptional()
-  avatarX?: number;
```
vdedios, line 13: We are not going to be able to edit chatroom avatar?

transcendence-app/src/chat/chatroom/infrastructure/db/chatroom.repository.ts
```diff
@@ -9,7 +9,7 @@ export abstract class IChatroomRepository {
   abstract deleteByChatroomName(name: string): Promise<Chatroom | null>;
   abstract updateById(
     name: string,
-    updateChatDto: UpdateChatroomDto,
+    updateChatroomDto: UpdateChatroomDto,
```
vdedios, line 12: Thanks :)

transcendence-app/src/chat/chat.controller.ts
gbudau: thought: In this case, we could go with `UnprocessableEntityException` or `ConflictException` because the name is unique in the [Chatroom table](https://github.com/42AGV/ft_transcendence/blob/21d562e00022cc6436fd5bd47618c95f788be3c5/db/setup/schema.sql#L45).

And if we make the name non-unique, `ServiceUnavailable` then looks more appropriate.

transcendence-app/src/chat/chat.controller.ts
```diff
@@ -184,6 +186,23 @@ export class ChatController {
     return messages;
   }
 
+  @Patch('room/:chatroomId')
+  @ApiBadRequestResponse({ description: 'Bad Request' })
+  @ApiUnprocessableEntityResponse({ description: 'Unprocessable Entity' })
+  async updateChatroom(
```
gbudau, line 192: praise: I like how you implement this method :+1:

transcendence-app/src/chat/chat.service.ts
```diff
@@ -106,4 +107,34 @@ export class ChatService {
       },
     );
   }
+
+  async updateChatroom(
```
gbudau, line 111: praise: Looks very good :+1:

suggestion: We could check if the user is the chatroom owner before updating it.

gbudau (approved):

transcendence-app/src/chat/chat.service.ts
fizcamposancos: Like a double check, access to the edit details page only if admin or owner, and endpoint available only if admin or owner?

fizcamposancos: I'll create an issue about the web admin role

transcendence-app/src/chat/chat.controller.ts
```diff
@@ -187,6 +189,31 @@ export class ChatController {
...
+    @GetUser() userMe: User,
+    @Param('chatroomId', ParseUUIDPipe) chatroomId: string,
+    @Body() updateChatroomDto: UpdateChatroomDto,
+  ): Promise<Chatroom> {
+    const chatroom = await this.getChatroomById(chatroomId);
+    if (!chatroom) {
+      throw new ServiceUnavailableException();
+    }
```
gbudau, line 203: question: What do you think about using the `NotFoundException` here? Because the `getById` method also returns null when there is no chatroom with that ID. `ServiceUnavailable` is also technically correct because the database server could be offline or malfunctioning.

suggestion: Please add an Api*Response decorator for the Swagger documentation.

transcendence-app/src/chat/chat.controller.ts
```diff
@@ -187,6 +189,31 @@ export class ChatController {
...
+    if (userMe.id !== chatroom.ownerId) {
+      throw new ForbiddenException();
+    }
```
gbudau, line 206: :+1:

gbudau (approved):
Thank you! LGTM :+1:

transcendence-app/src/chat/chat.controller.ts
fizcamposancos: Ok, I'll do it

fizcamposancos: Actually, the getById method thows the NotFound exception. So we don't need it here don't we?

gbudau: You are right :+1:
What do you think about moving the owner check in the `ChatService.addChatroomMember` method after finding the chatroom, and we could remove this part from here?

fizcamposancos: Done

vdedios (approved):

vdedios:
Don't forget to squash!! or at least add the task reference to some commit :)

Damrod (approved):
LGTM!

PR #311, opened 2022-11-07
#226/create a chat room page

### LINKED ISSUE #226: ⚛️ Create a chat room page

gbudau:
close #226

### DIFF
--- a/transcendence-app/src/chat/chat.controller.ts
+++ b/transcendence-app/src/chat/chat.controller.ts
@@ -138,6 +138,27 @@ export class ChatController {
     return chatroomsMembers;
   }
 
+  @Get('room/:chatroomId/members/:userId')
+  @ApiOkResponse({
+    description: 'Get a chatroom member',
+    type: ChatroomMember,
+  })
+  @ApiBadRequestResponse({ description: 'Bad Request' })
+  @ApiNotFoundResponse({ description: 'Not Found' })
+  async getChatroomMember(
+    @Param('chatroomId') chatroomId: string,
+    @Param('userId') userId: string,
+  ) {
+    const chatroomMember = await this.chatroomMemberService.getById(
+      chatroomId,
+      userId,
+    );
+    if (!chatroomMember) {
+      throw new NotFoundException();
+    }
+    return chatroomMember;
+  }
+
   @Get('room/:id')
   @ApiCreatedResponse({
     description: 'Get a chatroom',
--- a/transcendence-app/src/chat/chat.controller.ts
+++ b/transcendence-app/src/chat/chat.controller.ts
@@ -57,7 +57,7 @@ export class ChatController {
   async createChatroom(
     @GetUser() user: User,
     @Body() createChatDto: CreateChatroomDto,
-  ) {
+  ): Promise<Chatroom> {
     const chatroom = await this.chatService.createChatroom(
       user.id,
       createChatDto,
@@ -80,7 +80,7 @@ export class ChatController {
     @GetUser() user: User,
     @Param('chatroomId', ParseUUIDPipe) chatroomId: string,
     @Body() joinChatroomDto: JoinChatroomDto,
-  ) {
+  ): Promise<ChatroomMember> {
     const ret = await this.chatService.addChatroomMember(
       chatroomId,
       user.id,
@@ -148,7 +148,7 @@ export class ChatController {
   async getChatroomMember(
     @Param('chatroomId') chatroomId: string,
     @Param('userId') userId: string,
-  ) {
+  ): Promise<ChatroomMember> {
     const chatroomMember = await this.chatroomMemberService.getById(
       chatroomId,
       userId,
@@ -166,7 +166,9 @@ export class ChatController {
   })
   @ApiBadRequestResponse({ description: 'Bad Request' })
   @ApiNotFoundResponse({ description: 'Not Found' })
-  async getChatroomById(@Param('id', ParseUUIDPipe) chatroomId: string) {
+  async getChatroomById(
+    @Param('id', ParseUUIDPipe) chatroomId: string,
+  ): Promise<Chatroom> {
     const chatroom = await this.chatService.getChatroomById(chatroomId);
     if (!chatroom) {
       throw new NotFoundException();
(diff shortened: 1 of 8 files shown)


transcendence-app/src/chat/chat.controller.ts
```diff
@@ -138,6 +138,27 @@ export class ChatController {
...
+  @ApiOkResponse({
+    description: 'Get a chatroom member',
+    type: ChatroomMember,
+  })
+  @ApiBadRequestResponse({ description: 'Bad Request' })
+  @ApiNotFoundResponse({ description: 'Not Found' })
+  async getChatroomMember(
+    @Param('chatroomId') chatroomId: string,
+    @Param('userId') userId: string,
+  ) {
```
Damrod, line 151: What type are we returning from here?

Damrod: No problem. Just wondering.

vdedios (approved):
Great work :+1: 

I think we start to need those seed files because this is getting harder to test :D

Related to this, it would be great if we could have some mocked users with passwords that never change because it's so anoying having to create a new one each time :sweat:

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
```diff
@@ -45,8 +107,16 @@ function Chatroom() {
       <div className="chatroom-message-input">
         <ChatroomMessageInput to={chatroomId} />
       </div>
+      <div className="chatroom-status">
```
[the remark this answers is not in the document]
vdedios: I'm working on[ this poc](https://github.com/42AGV/ft_transcendence/issues/314), if we all like it we can use it instead :stuck_out_tongue_winking_eye:

Damrod:
Can we have the chatroom title be the name of the chatroom and not just the uuid, please?

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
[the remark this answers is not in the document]
vdedios: As said -> https://github.com/42AGV/ft_transcendence/pull/318

webapp/src/pages/ChatroomPage/components/ChatroomMessages/ChatroomMessages.tsx
```diff
@@ -1,46 +1,63 @@
...
 
     socket.on('chatroomMessage', messageListener);
-    socket.on('chatroomMessages', messagesListener);
-    socket.emit('getChatroomMessages', from);
 
     return () => {
       socket.off('chatroomMessage');
-      socket.off('chatroomMessages');
     };
   }, [from]);
```
fizcamposancos, line 63: A general question about `useEffect` What's better, a single `useEffect` with all the dependencies or separate ones, like you have here?

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
```diff
@@ -1,42 +1,104 @@
...
+  if (isChatroomLoading || isChatroomMemberLoading) {
+    return (
+      <div className="chatroom">
+        <div className="chatroom-loading">
+          <Loading />
+        </div>
+      </div>
+    );
+  }
+
```
fizcamposancos, line 88: This way we will always show the loading spinner, even for an instant. I don't now if it is a good practice, but what do you think about fooling the eye a little bit loading the immediate components of the page and the ones with a query a bit later.
I mean, for example, for the chatroom name, removing this `Loading` component and in line 101 something like this `{isChatroomLoading ? '' : chatroom!.name} `. But I insist, I don't know if this make sense or is a complete nuisance.

fizcamposancos (approved):
Very nice job.

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
```diff
@@ -1,42 +1,104 @@
...
+  const { data: chatroom, isLoading: isChatroomLoading } = useData(getChatroom);
+  const getChatroomMember = useCallback(
+    () =>
+      chatApi.chatControllerGetChatroomMember({
+        chatroomId,
+        userId: authUser.id,
+      }),
+    [chatroomId, authUser.id],
+  );
+  const { data: chatroomMember, isLoading: isChatroomMemberLoading } =
```
Damrod, line 57: Non blocking, but, what do you think about the idea of keeping a local copy of all chatmembers with userId === authuser, and update them when we also update the db, the way we do with useAuth? Here we wouldn't need to call the ChatAPI.

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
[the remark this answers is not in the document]
fizcamposancos: How can I test this?

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
gbudau: I see your point, and I'm not sure which is better for this page.
If it takes a long time to fetch the chatroom info because the user has slow internet or the server is under heavy load, the user will see an empty chatroom, and if they try to send messages, it won't work.
On the other side, they will see a loading spinner.

Also, code-wise will be pretty hard to understand what and when it renders because we have to add some conditionals.

We could display a skeleton screen on pages that don't have much user interactivity, like viewing a user profile page, chatroom details page, etc.

What do you think?

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
gbudau: Not a bad idea. I can think of some edge cases. For example, the user could browse from multiple browsers. If they join/leave from one browser, we store that info locally, and other browsers won't be aware until they refresh the page. Or an owner/admin could block them, and even if they are members, they can't access the chatroom.
I don't know how we could sync this info.

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
vdedios: Agree with @gbudau , the alternative would be using skeletons for those components not loaded yet but I think that would be overkill with the time we have.

Managing skeletons is not easy (you have to predict content size so user don't see unexpected jumps, we would have to rework some pages to make the more modular etc.)

I'd say a good enough solution is what we have, a simple loader for the whole screen. That being said, thanks for bringing the point, I think it's an interesting one :+1:

gbudau:
Thank you for your feedback and suggestions!

Issue #312, opened 2022-11-08
⚛️ Small button


Damrod:
While developing some pages with a lot of content, we have noticed we may lighten up the UI by creating a small button, as per figma draft. This issue addresses this.

PR #313, opened 2022-11-08
fix build

Damrod:
As explained in #312 and groomed on slack huddle

### DIFF
--- a/webapp/src/shared/components/Button/Button.tsx
+++ b/webapp/src/shared/components/Button/Button.tsx
@@ -1,5 +1,6 @@
 import Icon, { IconVariant, IconSize } from '../Icon/Icon';
 import './Button.css';
+import React from 'react';
 
 export enum ButtonVariant {
   SUBMIT = 'submit',
@@ -7,14 +8,23 @@ export enum ButtonVariant {
   ALTERNATIVE = 'submit-alternative',
 }
 
-type ButtonProps = {
+type CommonButtonProps = {
   variant: ButtonVariant;
   iconVariant?: IconVariant;
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
   disabled?: boolean;
   form?: string;
+};
+
+type LargeButtonProps = CommonButtonProps & {
   children: string;
 };
+
+type SmallButtonProps = CommonButtonProps & {
+  children?: never;
+};
+
+type ButtonProps = SmallButtonProps | LargeButtonProps;
 
 export default function Button({
   variant,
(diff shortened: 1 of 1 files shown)


webapp/src/pages/LoginPage/LoginPage.css
```diff
@@ -41,9 +40,9 @@
 }
 
 .login-form > .input,
-.button,
-.final-text,
-.login-container__separator {
+.login .button,
+.login .final-text,
+.login .login-container__separator {
```
vdedios, line 45: It seems that adding this `.login` makes button be wider than text in some cases:

https://user-images.githubusercontent.com/57688116/200768230-204156ce-cbe2-4744-a208-635d73e88bd1.mp4

And expected would be button to be width aligned with the text (main version):

https://user-images.githubusercontent.com/57688116/200768723-99038a23-316f-4d50-ba71-f30416b3ab16.mp4

webapp/src/shared/components/Button/Button.css
```diff
@@ -12,6 +10,16 @@
...
 }
 
+.button-large {
+  width: 15rem;
+  border-radius: 0.5rem;
+}
+
+.button-small {
+  width: 3rem;
+  border-radius: 50%;
```
vdedios, line 20: Not sure if yours, but button width in edit profile page seems to be broken:

image

vdedios (commented):
A few aesthetic comments, thanks for working on this :heart:

webapp/src/pages/ComponentsBookPage/ComponentsExamples/ButtonsExample.tsx
```diff
@@ -32,5 +32,22 @@ export const ButtonsExample = () => (
...
+        disabled
+      />
+    </BookSubsection>
+    <BookSubsection title="Small submit type with icon">
+      <Button
+        variant={ButtonVariant.SUBMIT}
+        iconVariant={IconVariant.ARROW_FORWARD}
+        onClick={buttonLink}
+      />
+    </BookSubsection>
```
vdedios, line 51: Thanks :)

gbudau: :+1:

webapp/src/shared/components/Button/Button.tsx
```diff
@@ -1,21 +1,31 @@
...
+
+type LargeButtonProps = CommonButtonProps & {
   children: string;
 };
 
+type SmallButtonProps = CommonButtonProps & {
+  children?: never;
+};
+
+type ButtonProps = SmallButtonProps | LargeButtonProps;
```
gbudau, line 27: praise: I like this :+1:

webapp/src/pages/LoginPage/LoginPage.css
Damrod: Okay, I see, I broke something else. Either way, that was conceptually wrong, because we were changing the style of the global button from the login page style sheet.
I was looking at where were we setting the width of the button to 15 in the components book page, and it was... in the login page css!

webapp/src/shared/components/Button/Button.css
Damrod: Yes, you broke the edit avatar page button back in september, and now in order to fix the broken edit avatar button, I broke every other button. Should be fixed now.

webapp/src/pages/ComponentsBookPage/ComponentsExamples/ButtonsExample.tsx
Damrod: you're welcome

webapp/src/shared/components/Button/Button.css
```diff
@@ -12,6 +10,16 @@
   justify-content: center;
 }
 
+.button-large {
+  width: 15rem;
```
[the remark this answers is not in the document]
Damrod: done

PR #318, opened 2022-11-09
#314 / create basic snackbar component

### LINKED ISSUE #314: ⚛️ Snackbar component
Create a proof of concept for a snackbar component to be used further as a notification component

vdedios:
closes #314


webapp/src/shared/components/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const onAnimationEnd = () => {
+    if (!visible) {
+      setShouldRender(false);
+    }
+  };
+
+  return (
+    <>
+      {shouldRender && (
```
vdedios, line 42: React makes kind of tricky to add animations on unMount state, because it directly removes elements of the DOM when they don't have to be rendered anymore. So we need an intermediate condition which will be set to false once the animation ends. Got some inspiration from -> https://czaplinski.io/blog/super-easy-animation-with-react-hooks/

webapp/src/shared/components/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,62 @@
...
+    if (visible) {
+      setShouldRender(true);
+    }
+  }, [visible]);
+
+  const onAnimationEnd = () => {
+    if (!visible) {
+      setShouldRender(false);
+    }
+  };
```
vdedios, line 38: As said, when animation ends we finally set render state to false

webapp/src/shared/components/Snackbar/Snackbar.css
```diff
@@ -0,0 +1,57 @@
...
+  position: fixed;
+  text-decoration: none;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-direction: row;
+  border: none;
+  border-radius: 0.75rem;
+  height: 3rem;
+  width: 15rem;
```
Damrod, line 11: why forced to 15 wide? We may need to display slightly longer texts, don't you think?

With long texts, it doesn't really work properly.
Screenshot from 2022-11-10 01-35-02

vdedios: Update it to grow with content, thanks :+1:

gbudau (approved):
Thank you for this feature! I'm looking forward to trying it out once we complete #319.

fizcamposancos (approved):
I love it

vdedios:
Thanks! second part in https://github.com/42AGV/ft_transcendence/pull/324

PR #321, opened 2022-11-09
Fix production build

gbudau:
Fix docker-compose.prod.yml environment variables
Add webapp.conf nginx config for the webapp
Fix schema.sql syntax
Update fill-db.sh script to take into account the PostgreSQL user Update transcendence-app .dockerignore to ignore .env* files Update transcendence-app main.ts to not save the Swagger documentation on production environment
Update transcendence-app tsconfig.build.json to exclude knex and types files Update webapp package.json move react-scripts and @types/react-dom to dependencies

### DIFF
--- a/db/fill-db.sh
+++ b/db/fill-db.sh
@@ -5,10 +5,10 @@
 SETUP_DIR=/var/lib/postgresql/setup
 TABLES=schema.sql
 
-echo "$0: running $TABLES"; psql -d "$POSTGRES_DB" -f "$SETUP_DIR/$TABLES";
+echo "$0: running $TABLES"; psql -d "$POSTGRES_DB" -U "$POSTGRES_USER" -f "$SETUP_DIR/$TABLES";
 
 for f in $SETUP_DIR/*.sql; do
     if [ "$f" != "$SETUP_DIR/$TABLES" ] ; then
-        echo "$0: running $f"; psql -d "$POSTGRES_DB" -f "$f";
+        echo "$0: running $f"; psql -d "$POSTGRES_DB" -U "$POSTGRES_USER" -f "$f";
     fi
 done
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -10,14 +10,20 @@ services:
       target: production
       args:
         - REACT_APP_NAME=webapp-prod
+    volumes:
+      - ./nginx/webapp.conf:/etc/nginx/conf.d/default.conf:ro
 
   transcendence-app:
     build:
       context: ./transcendence-app
       target: production
     environment:
-      - POSTGRES_USER={POSTGRES_USER:?required}
-      - POSTGRES_PASSWORD={POSTGRES_PASSWORD:?required}
+      - POSTGRES_USER=${POSTGRES_USER:?required}
+      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD:?required}
+      - FORTYTWO_APP_ID=${FORTYTWO_APP_ID:?required}
+      - FORTYTWO_APP_SECRET=${FORTYTWO_APP_SECRET:?required}
+      - SESSION_SECRET=${SESSION_SECRET:?required}
+      - MEMCACHED_SECRET=${MEMCACHED_SECRET:?required}
 
   db:
     environment:
--- a/nginx/webapp.conf
+++ b/nginx/webapp.conf
@@ -1,0 +1,16 @@
+server {
+    listen       80;
+    listen  [::]:80;
+    server_name  localhost;
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri /index.html;
+    }
+
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+}
--- a/transcendence-app/src/main.ts
+++ b/transcendence-app/src/main.ts
@@ -6,6 +6,8 @@ import { writeFileSync } from 'fs';
 import { AppModule } from './app.module';
 import * as yaml from 'yaml';
 import { setupApp } from './setup-app';
+import { ConfigService } from '@nestjs/config';
+import { EnvironmentVariables } from './config/env.validation';
 
 async function bootstrap() {
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
@@ -22,8 +24,13 @@ async function bootstrap() {
     .setVersion('1.0')
     .build();
   const document = SwaggerModule.createDocument(app, options);
-  const yamlString: string = yaml.stringify(document, {});
-  writeFileSync('./swagger-spec.yaml', yamlString);
+  const configService =
+    app.get<ConfigService<EnvironmentVariables>>(ConfigService);
+  const nodeEnv = configService.get<string>('NODE_ENV');
+  if (nodeEnv && nodeEnv !== 'production') {
+    const yamlString: string = yaml.stringify(document, {});
+    writeFileSync('./swagger-spec.yaml', yamlString);
+  }
   SwaggerModule.setup('api', app, document);
   await app.listen(3000);
 }
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -3,9 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/react-dom": "^18.0.6",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-router-dom": "^6.3.0",
+    "react-scripts": "^5.0.1",
     "socket.io-client": "^4.5.2",
     "web-vitals": "^2.1.4"
   },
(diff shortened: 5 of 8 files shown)


nginx/webapp.conf
```diff
@@ -0,0 +1,16 @@
+server {
+    listen       80;
+    listen  [::]:80;
+    server_name  localhost;
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri /index.html;
```
gbudau, line 9: We need to add this ruleto make it work with react-router https://stackoverflow.com/a/43954597

docker-compose.prod.yml
```diff
@@ -10,14 +10,20 @@ services:
...
       target: production
     environment:
-      - POSTGRES_USER={POSTGRES_USER:?required}
-      - POSTGRES_PASSWORD={POSTGRES_PASSWORD:?required}
+      - POSTGRES_USER=${POSTGRES_USER:?required}
+      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD:?required}
+      - FORTYTWO_APP_ID=${FORTYTWO_APP_ID:?required}
+      - FORTYTWO_APP_SECRET=${FORTYTWO_APP_SECRET:?required}
+      - SESSION_SECRET=${SESSION_SECRET:?required}
+      - MEMCACHED_SECRET=${MEMCACHED_SECRET:?required}
```
gbudau, line 26: I've added these variables here, and we can omit the `transcendence-app/.env-production` file.

Or we could add a volume and mount the `.env-production` file inside the container because, in the development, we have a volume for the transcendence-app files.

transcendence-app/src/main.ts
```diff
@@ -22,8 +24,13 @@ async function bootstrap() {
...
   const document = SwaggerModule.createDocument(app, options);
-  const yamlString: string = yaml.stringify(document, {});
-  writeFileSync('./swagger-spec.yaml', yamlString);
+  const configService =
+    app.get<ConfigService<EnvironmentVariables>>(ConfigService);
+  const nodeEnv = configService.get<string>('NODE_ENV');
+  if (nodeEnv && nodeEnv !== 'production') {
+    const yamlString: string = yaml.stringify(document, {});
+    writeFileSync('./swagger-spec.yaml', yamlString);
+  }
```
gbudau, line 33: There was some permission error while creating this file in production environment.

If we want to keep it, we could change the permissions for the /transcendence-app/dist directory.

webapp/package.json
```diff
@@ -3,9 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/react-dom": "^18.0.6",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-router-dom": "^6.3.0",
+    "react-scripts": "^5.0.1",
```
gbudau, line 10: Create react app needs these two packages to build

db/fill-db.sh
```diff
@@ -5,10 +5,10 @@
 SETUP_DIR=/var/lib/postgresql/setup
 TABLES=schema.sql
 
-echo "$0: running $TABLES"; psql -d "$POSTGRES_DB" -f "$SETUP_DIR/$TABLES";
+echo "$0: running $TABLES"; psql -d "$POSTGRES_DB" -U "$POSTGRES_USER" -f "$SETUP_DIR/$TABLES";
```
gbudau, line 8: Use the user from the `POSTGRES_USER` environment variable, by default it was using the `postgres` user.

transcendence-app/src/main.ts
gbudau: I've removed this and fixed the directory permission.

.github/workflows/production-run_check.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Cache Docker layers
+        uses: actions/cache@v3
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
```
gbudau, line 23: From https://docs.docker.com/build/ci/github-actions/examples/#github-cache

.github/workflows/production-run_check.yml
```diff
@@ -0,0 +1,73 @@
...
+          target: production
+          push: false
+          load: true
+          tags: ft_transcendence/webapp-prod:latest
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+      - name: Run
+        run: |
+          docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d
+          docker-compose -f docker-compose.yml -f docker-compose.prod.yml down
```
gbudau, line 59: Here we use `docker-compose up` without the `--build` flag to use the images from the cache.

.github/workflows/production-run_check.yml
```diff
@@ -0,0 +1,74 @@
...
+        env:
+          POSTGRES_USER: ${{ secrets.POSTGRES_USER }}
+          POSTGRES_PASSWORD: ${{ secrets.POSTGRES_PASSWORD }}
+          FORTYTWO_APP_ID: ${{ secrets.FORTYTWO_APP_ID }}
+          FORTYTWO_APP_SECRET: ${{ secrets.FORTYTWO_APP_SECRET }}
+          FORTYTWO_APP_CALLBACK_URL: ${{ secrets.FORTYTWO_APP_CALLBACK_URL }}
+          SESSION_SECRET: ${{ secrets.SESSION_SECRET }}
+          MEMCACHED_SECRET: ${{ secrets.MEMCACHED_SECRET }}
```
vdedios, line 67: Since we are only going to use this as a build test, why not using just

README.md
```diff
@@ -236,7 +236,93 @@ TODO: Add notes about how to use the system.
...
+
+Generate the generated files, that are not a part of the repository:
+
+```shell
+make gen
+```
+or
+```shell
+./scripts/generate-openapi.sh [options]
+```
```
vdedios, line 309: I'd just keep the make instructions in the readme. They are handier and nobody is going to use the scripts if you have a good makefile. Also, we want readme to be concise

nginx/webapp.conf
vdedios: Could you elaborate why we need this additional config?

README.md
```diff
@@ -236,7 +236,93 @@ TODO: Add notes about how to use the system.
...
+
+
+Change the current working directory to `ft_transcendence`
+
+```shell
+cd ft_transcendence
+```
+
+Either create a `.env` file in the root directory containing the following environment variables (replace the FORTYTWO_APP_ID and FORTYTWO_APP_SECRET with the ones from the OAuth application you create in the previous step):
```
vdedios, line 275: If we are going to use `.env` from now, shoudn't be `.env.sample` and `.env.test` deleted?

vdedios (approved):
Nice! thanks for working on this :)

.github/workflows/production-run_check.yml
gbudau: I think this comment stops mid-sentence. Can you clarify, please?

nginx/webapp.conf
fizcamposancos: Are you already testing the production environment? This seems like something I would find or even thought about at this moment. I am really impressed

README.md
```diff
@@ -236,7 +196,73 @@ TODO: Add notes about how to use the system.
...
+POSTGRES_PASSWORD=your-postgres-password
+SESSION_SECRET=your-session-secret-uuid
+MEMCACHED_SECRET=your-memcached-secret-uuid
+```
+
+Or export them from the shell like:
+
+```shell
+export FORTYTWO_APP_ID=your-42-app-id
+export FORTYTWO_APP_SECRET=your-42-app-secret
```
Damrod, line 251: Isn't this impossible to properly test as long as our cursus is frozen?

.github/workflows/production-run_check.yml
Damrod: what about we add some make rule to stop the production server, analogous to `make prod`? Something like this `docker-compose -f docker-compose.yml -f docker-compose.prod.yml down` maybe?

README.md
```diff
@@ -236,7 +196,73 @@ TODO: Add notes about how to use the system.
 
 ## 🚀 Deployment <a name = "deployment"></a>
 
-TODO: Add additional notes about how to deploy this on a live system.
+### Prerequisites
```
Damrod, line 199: Maybe here it would be good to explain that this whole section is about running the app in production mode, as opposed to the first hald that is running the app in development mode.

Also, I'm having trouble with the production mode because I have by default an apache2 server running on port 80, so maybe it would be good to say that we should shut down any such services before.

For me it was `sudo service apache2 stop`

Damrod:
I'm having this error when browsing the landing page and running the app in production mode
Screenshot from 2022-11-17 20-18-22
Was it always there even in development mode but I only noticed it now? Or is a production mode problem?

Damrod (approved):
LGTM!
Thank you for this feature!

README.md
gbudau: Updated. Thanks for the suggestions.

.github/workflows/production-run_check.yml
gbudau: Done :+1:

gbudau:
> I'm having this error when browsing the landing page and running the app in production mode
Was it always there even in development mode but I only noticed it now? Or is a production mode problem?

If I'm not mistaken, this log comes from the `useData` hook when there is a request error. Should we remove it?

gbudau:
Thanks for your suggestions!

PR #322, opened 2022-11-09
#276/reorder make clean rule

Damrod:
When running make clean, first we remove the generated code, and then stop the running deployment of our app. If a browser is actively running the app, it will complain that the generated files are missing, before it goes down. Removing the files AFTER stopping the deployment would make a more graceful shutdown.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ gen:
 
 .PHONY: clean
 clean:
+	$(DOCKER_COMPOSE) down -v
 	rm -rf $(PROJECT_ROOT)/webapp/src/shared/generated || true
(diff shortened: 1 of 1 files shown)


vdedios (approved):
Not directly related to this PR but it would be nice to have a `test` rule so we can pass the full test suite. What do you think?

gbudau (approved):
Thanks!

fizcamposancos (approved):
LGTM

Damrod:
> Not directly related to this PR but it would be nice to have a `test` rule so we can pass the full test suite. What do you think?

why not

PR #329, opened 2022-11-13
#315/Create a page to join a chat room

### LINKED ISSUE #241: ⚛️  Add a button to join a chat room.

### LINKED ISSUE #242: ⚛️  Add a form to enter the password in case of a password-protected chat room

### LINKED ISSUE #315: ⚛️  Create a page to join a chat room
Subtasks:

- [x] #242 
- [x] #241

gbudau:
close #241
close #242
close #315


Damrod:
If the user is not a member of the chatroom, and is redirected to the join page, I don't think we should display the notification 'Not a member of the chatroom. Forbidden' anymore. But we do, even after your changes.

Damrod:
What do we plan to do with areas 1 and 2?
And Why no public/private in area 3?
no_privacy_status
bad_area_desktop
bad_area_mobile

Damrod:
Some gap in 4? 
nogap
..
Are we even seing the same designs?
Screenshot from 2022-11-13 23-42-21

transcendence-app/src/chat/chat.service.ts
```diff
@@ -165,12 +165,7 @@ export class ChatService {
     if (!foundChatroom) {
       throw new NotFoundException('Chatroom not found');
     }
-    const { password, confirmationPassword } = joinChatroomDto;
-    if (password !== confirmationPassword) {
-      throw new BadRequestException(
-        'Password and Confirmation Password must match',
-      );
-    }
+    const { password } = joinChatroomDto;
```
Damrod, line 168: Good catch!

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+  return (
+    <div className="join-chatroom">
+      <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
+        join chatroom
+      </Header>
+      <div className="join-chatroom-avatar">
+        <LargeAvatar
+          url={
+            chatroom.avatarId
+              ? `${CHATROOM_EP_URL}/${chatroomId}/avatars/${chatroom.avatarId}`
```
Damrod, line 105: maybe leave a TODO to issue 317?

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <div className="join-chatroom">
+      <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
+        join chatroom
+      </Header>
+      <div className="join-chatroom-avatar">
+        <LargeAvatar
+          url={
+            chatroom.avatarId
```
Damrod, line 104: Also, are we sure it's reasonable to make different endpoints for avatars for chat and users?
In `webapp/src/pages/ChatPage/ChatPage.tsx` line 28, we are presuming it's `url: `${AVATAR_EP_URL}/${chatroom.avatarId}`,` instead

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+    event.preventDefault();
+    try {
+      await chatApi.chatControllerCreateChatroomMember({
+        chatroomId: chatroomId,
+        joinChatroomDto: {
+          password: isProtectedChatroom ? password : undefined,
+        },
+      });
+      notify('You joined a chatroom');
+      navigate(`${CHATROOM_URL}/${chatroomId}`, { replace: true });
```
Damrod, line 65: What does `{ replace: true }` do?

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+    } catch (error: unknown) {
+      if (error instanceof ResponseError) {
+        const responseBody = await error.response.json();
```
Damrod, line 68: Nice!

Damrod (approved):
besides the comments, LGTM! Thanks!

gbudau:
> If the user is not a member of the chatroom, and is redirected to the join page, I don't think we should display the notification 'Not a member of the chatroom. Forbidden' anymore. But we do, even after your changes.

I've tried the `hide()` method from the `useNotificationContext()`, and it still shows for a few moments.

@vdedios, any idea how we could do it?

Or maybe we could change the warning to something else, like "_You are not a chatroom member. Please join the chatroom first_" or something like that.

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
gbudau: Thanks. Done.

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
gbudau: Good catch! I've changed the `url` of the avatar in the `ChatPage`.

I think we need a different endpoint for the chatroom avatars because the current endpoint is for the user avatars `/api/v1/users/avatars/{avatarId}`, or we could create a new controller for the avatars with an endpoint like `/api/v1/avatar/{avatarId}`

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
gbudau: A call to `navigate` pushes a new entry into the history stack so the user can click the back button to get back to the page. If we pass `{ replace: true }`, we replace the current entry with the new one.
For example, if the user clicks the "join" button, we replace the join chatroom screen with the chatroom screen. Then when they click the back button, they don't see the join page again.

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
gbudau: Thanks!

gbudau:
@Damrod
Updated. Thanks for the suggestions.

Damrod:
We can do better than this:
[unmatched_styles.webm](https://user-images.githubusercontent.com/2774420/201767830-861822ad-9888-4a89-83ee-ec6c79f6275e.webm)
But probably the sensible way to do it would be a template.
Don't consider this to be blocking. Do merge, please.

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
Damrod: I think the new controller with the `/api/v1/avatar/{avatarId}` endpoint would be the way to go. But it's not for this PR, obviouly.

vdedios (approved):
Nice! 👍 Please, review my comment about the notification!

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
```diff
@@ -0,0 +1,141 @@
...
+        if (responseBody.message) {
+          warn(responseBody.message);
+        } else {
+          warn(error.response.statusText);
+        }
```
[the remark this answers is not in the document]
gbudau: I've added a boolean to check if the chatroom exists and if the user is a chatroom member. It should work now.

gbudau:
Thanks!

Issue #207, opened 2022-10-19
⚛️  Create an edit chat room member page


gbudau:
Webapp URL -> `/chatroom/{roomId}/{username}/edit`

PR #331, opened 2022-11-15
#206/Create an endpoint to update a chat room member

### LINKED ISSUE #206: Create an endpoint to update a chat room member
Create a `PATCH /chat/room/{roomId}/members/{memberId}` endpoint to update a chat room member
This endpoint must be protected. Only the chat room owners or administrators have access

gbudau:
close #206


Damrod:
As of now, it's possible to have owner true and admin false, but owner always implies at least admin. Do you think we could simplify some logic if we enforced this relationship in the database?

gbudau:
> As of now, it's possible to have, for the same chatroom member, owner true and admin false, but owner always implies at least admin rights and privileges. Do you think we could simplify some logic if we enforced this relationship in the database?

Good point. We check if the user is the chatroom owner and ignore other properties, so maybe not?

transcendence-app/src/chat/chat.service.ts
```diff
@@ -180,4 +184,46 @@ export class ChatService {
...
+    }
+    if (authChatroomMember.owner) {
+      return this.chatroomMemberService.updateById(
+        chatroomId,
+        userId,
+        updateChatroomMemberDto,
+      );
+    }
+    if (
+      !authChatroomMember.admin ||
```
vdedios, line 217: I'm not very sure where are we going to designate the admins yet 😅 We don't have any screen defined where they are going to exist, right?

vdedios (commented):
nice! 🚀

transcendence-app/src/chat/chat.service.ts
gbudau: We have a design proposal made by @Damrod, [here](https://www.figma.com/file/EFntd84Sl8NuHJQBrkqBVS/aollero-%2Fchat-v0.2?node-id=751%3A2184&t=Iirxid9OyqGNW8y0-0)
And a task for it, #207

Damrod: Maybe you are speaking about different things?
Chatroom admins vs site wide admins?
Maybe not?

gbudau: Perhaps. I was referring to chatroom admins :sweat_smile:

transcendence-app/src/chat/chat.service.ts
```diff
@@ -175,4 +179,46 @@ export class ChatService {
...
+      chatroomMember.admin ||
+      updateChatroomMemberDto.admin !== undefined
+    ) {
+      throw new ForbiddenException();
+    }
```
Damrod, line 217: A banned admin cannot update itself, and also can still update non owners, even if the admin is banned. Does this make sense?

Damrod (approved):
LGTM!

Still concerned about my comments, but to test all cases easily I think the UI is in order.

transcendence-app/src/chat/chat.service.ts
gbudau: Good catch! What do we think is a better option to solve this?

1. Add a check constraint on the `ChatMember` table to disallow a chatroom member that is admin and banned/muted. For example:

```sql
    CHECK (
      NOT (
        (
          admin IS true
          AND banned IS true
        )
        OR (
          admin IS true
          AND muted IS true
        )
      )
    )
```
2. Add a check to this condition.

gbudau:
Thanks!

transcendence-app/src/chat/chat.service.ts
Damrod: I don't think this is a good solution:

> When owner of a channel, user can kick, ban, mute other users and administrators When administrator of a channel, user can kick, ban mute other users, but not owners

gbudau: I see. With this solution (table constraint check), the owner has to remove the admin rights before banning/muting an admin.

Then it is better to add some condition to check that the authenticated user is an admin and it's not banned or muted.

gbudau: BTW. If the admin is not banned but muted, should we allow them to ban/mute other users?

gbudau: Updated. I've added the conditions to this `if` statement.

Issue #337, opened 2022-11-19
Create Avatar Page template


Damrod:
We have several pages that follow the same UI design template:
For desktop: header on top, A big avatar on the left, some content on the right, and a button on the bottom.
For mobile: header on top, A big avatar, some content, and a button on the bottom.

For all of these pages, we have slightly different style, which leads to slight missalignments, and different responsive behaviour across them.

This is an opportunity to create a component/templatePage that all of these pages can use to create their own customized version, leading to a more uniform style.

As starting point, a task to create a template is proposed:
- [x] generic "AvatarPage" template
- [x]  add optional secondary button to avatar page

As of now, the pages following with more or less success (or could follow to greater design wholeness) this template are:
- [x] edit user page
- [ ] user page
- [x] myprofile page
- [ ] join chat page
- [x] create chat page
- [ ] edit chatroom details
- [x] ... edit chatmember page

These pages could be updated to use the template from the previous step.

Damrod:
#272 is related

Damrod:
For traceability I quote myself from [here](https://github.com/42AGV/ft_transcendence/pull/345#issue-1458505975)
> Between current main and commit https://github.com/42AGV/ft_transcendence/commit/c3bd9372da6b661cd5121fb526f3f098f91213d6 , some commits were lost (namely at least this https://github.com/42AGV/ft_transcendence/commit/1208122abc909b4810cbdade0174ff3f0122f28e and this https://github.com/42AGV/ft_transcendence/commit/a1216f46ce0aa0cd33c95ceae1e228c2f219461a ).
This makes it so that the pages from the ancestry of the edit user page, have a slightly different styling than the pages made from the ancestry of the user page.
This PR is the first of three, where it provides the necessary functionality to the avatar page template to fit the edit chat page as well, as a series of fitting edit chat page and edit user page into the avatar page template.

PR #346, opened 2022-11-21
#337 / fit edit user page in avatar template page

Damrod:
as explained in #337 

> Between current main and commit https://github.com/42AGV/ft_transcendence/commit/c3bd9372da6b661cd5121fb526f3f098f91213d6 , some commits were lost (namely at least this https://github.com/42AGV/ft_transcendence/commit/1208122abc909b4810cbdade0174ff3f0122f28e and this https://github.com/42AGV/ft_transcendence/commit/a1216f46ce0aa0cd33c95ceae1e228c2f219461a ).
> This makes it so that the pages from the ancestry of the edit user page, have a slightly different styling than the pages made from the ancestry of the user page.

This PR is the second of three, where it fits edit user page into the avatar page template.


webapp/src/pages/EditUserPage/EditUserPage.tsx
```diff
@@ -1,45 +1,31 @@
...
-        edit profile
-      </Header>
-      <div className="edit-user-avatar">
-        <LargeAvatar
-          url={`${AVATAR_EP_URL}/${authUser.avatarId}`}
-          editUrl={EDIT_AVATAR_URL}
-          XCoordinate={authUser.avatarX}
-          YCoordinate={authUser.avatarY}
+  return (
+    <AvatarPageTemplate
```
gbudau, line 12: praise: Good idea for adding this template :+1:
It looks like we are using it in lots of places.

gbudau (approved):
Thank you for these improvements! LGTM :tada:

PR #347, opened 2022-11-21
#300/create a seed file to seed the chatrooms

### LINKED ISSUE #300: Create a seed file to seed the chatrooms

### LINKED ISSUE #349: Update sublevel of chatroom members row to show the chat-status
The sublevel of a user row in the chatroom member page should show the status of that user: owner, admin, muted, banned or blank

gbudau:
This PR adds database seed for users with passwords (the usernames are lowercase alphabetical letters, and the password is the same)

For example:
```
username: a
password: a
```

It also adds seeds for chatrooms and chatroom messages.

To test this, please do a `make re` to recreate the database and run the seed files.

close #300
close #349


transcendence-app/seeds/001_seed_users.ts
```diff
@@ -0,0 +1,69 @@
...
+    username,
+    email: `${username}@${username}.com`,
+    fullName: username,
+    avatarId: avatarId,
+    password: hashedPassword,
+  };
+};
+
+export async function seed(knex: Knex): Promise<void> {
+  // Deletes ALL existing entries and local files
```
gbudau, line 38: If we delete the users, we also delete the chatrooms because of the `ON DELETE CASCADE` on the `ownerId` column.

So we can also delete all the avatar files here.

fizcamposancos (approved):
Looks and works good. 
Just a general comment, beyond the scope of this PR, but related. You've identified the chat owner between parenthesis in the chat name. All right, but right now in a manually created chatroom there is no way to know the owner. According to our Figma design the chat status of an user is shown in the sublevel of the row. I will create an issue to remember updating it.
And if the owner enters his chat profile, the admin switch is off, even thought the owner is always an administrator. I think this should be always on for the owner. What do you think?

gbudau:
> All right, but right now in a manually created chatroom there is no way to know the owner. According to our Figma design the chat status of an user is shown in the sublevel of the row. I will create an issue to remember updating it.
> And if the owner enters his chat profile, the admin switch is off, even thought the owner is always an administrator. I think this should be always on for the owner. What do you think?

Good points. I will try to address them in this PR.

transcendence-app/src/chat/chatroom/chatroom-member/infrastructure/db/postgres/chatroom-member.postgres.repository.ts
```diff
@@ -137,8 +137,8 @@ export class ChatroomMemberPostgresRepository
 
     const orderBy =
       sort === BooleanString.True
-        ? ChatroomMemberKeys.JOINED_AT
-        : userKeys.USERNAME;
+        ? userKeys.USERNAME
```
Damrod: Didn't you say something like:

> we could add a second column to O`RDER BY`, for instance:
> `ORDER BY createdAt/joinedAt, userId`
> in this case, if createdAt/joinedAt isn't unique, it orders by userId

transcendence-app/seeds/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  const localFileService = new LocalFileService(
+    configService,
+    localFileRepository,
+  );
+  const avatarDto = await localFileService.createRandomSVGFile(12, 512);
+
+  return { id: uuidv4(), ...avatarDto };
+};
+
+export const defaultUsernames = 'abcdefghijklmnopqrstuvwxyz'.split('');
```
Damrod, line 27: Sometimes, I think it's useful to have usernames with overlapping sequences of characters, so that we can see how the table of (chatmembers in this case) dynamically builds depending on the search query string.

Something like "aa" "ab" "bb" "bc" "cb", for instance. But this is non blocking, You can merge and if someone really wants this feature, they can make the change.

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
```diff
@@ -45,6 +45,21 @@ export default function ChatroomDetailsPage() {
   const { authUser } = useAuth();
   const isOwner: boolean = authUser?.id === chatroom?.ownerId;
   const mapChatMemberToRow = (member: ChatroomMemberWithUser): RowItem => {
+    const memberDetails = () => {
```
Damrod, line 48: Thanks for this feature!

Damrod (approved):
Looks good to me! 
Thanks!

webapp/src/pages/ChatroomDetailsPage/ChatroomDetailsPage.tsx
gbudau: Good catch! Fixed.

transcendence-app/seeds/utils/index.ts
gbudau: Good idea! Done

gbudau:
Thank you for your suggestions!

PR #348, opened 2022-11-21
#337/fit edit chat details in avatar page template

Damrod:


Damrod:
Here I had a lot to adjust, because whenever I fixed something, something else was breaking. This is the end result.

Damrod:
There is some problem with the placeholder avatars (WILDCARD_AVATAR_URL), but they are transitory anyway.

webapp/src/shared/components/templates/AvatarPageTemplate/AvatarPageTemplate.tsx
```diff
@@ -23,6 +28,20 @@ type AvatarPageTemplateProps = {
...
+
+type AvatarPageCaptionLikeTemplateProps = AvatarPageTemplateCommonProps & {
+  variant: AvatarPageVariants;
+  captionLikeElement: JSX.Element;
+};
+
+type AvatarPageTemplateProps =
+  | AvatarPageRegularTemplateProps
+  | AvatarPageCaptionLikeTemplateProps;
+
```
gbudau, line 44: suggestion: What do you think about adding the prop `captionLikeElement?: JSX.Element;` to the `AvatarPageTemplateProps`. This way, we could remove the variant and do something like this:

```tsx
          <div className="avatar-page-caption-like-wrapper">
            {captionLikeElement && captionLikeElement}
          </div>
```

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
```diff
@@ -78,64 +73,54 @@ function JoinChatroom({ chatroomId }: JoinChatroomProps) {
...
   return (
-    <div className="join-chatroom">
-      <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
-        {`channel/${chatroom.name}`}
-      </Header>
-      <div className="join-chatroom-avatar">
-        <LargeAvatar
-          url={
+    <div className="join-chatroom-page">
+      <AvatarPageTemplate
```
gbudau, line 78: praise: Thanks for adding the template to these pages.

gbudau (approved):
Thank you for these improvements. LGTM :rocket:

webapp/src/shared/components/templates/AvatarPageTemplate/AvatarPageTemplate.tsx
Damrod: done. Thx!

webapp/src/pages/JoinChatroomPage/JoinChatroomPage.tsx
Damrod: Thank you!

PR #351, opened 2022-11-22
#341/Fix transcendence-app/dist permissions

### LINKED ISSUE #341: Fix transcendence-app/dist permissions
In current main branch, `transcendence-app/dist` folder contents are created with `root:root` owner and group. This prevents transcendence app from normally updating the contents of the folder upon modification, and forces a `chown $USER:$USER -R transcendence-app/dist` as a workaround.
Investigate what is the root cause of the problem, and fix it.

gbudau:
To test these changes, please create a `.env` file in the root directory containing the following environment variables:

```
FORTYTWO_APP_ID=your-42-app-id
FORTYTWO_APP_SECRET=your-42-app-secret
SESSION_SECRET=your-session-secret-uuid
MEMCACHED_SECRET=your-memcached-secret-uuid
```

close #341


docker-compose.override.yml
```diff
@@ -22,8 +22,9 @@ services:
       context: ./transcendence-app
       target: development
     volumes:
-      - ./transcendence-app:/transcendence-app
-      - /transcendence-app/node_modules
+      - ./transcendence-app/src:/transcendence-app/src:ro
+      - ./transcendence-app/test:/transcendence-app/test:ro
+      - ./transcendence-app/seeds:/transcendence-app/seeds:ro
```
gbudau, line 27: We mount only these subdirectories inside the container because if we mount the entire directory, Docker creates the `dist` and `node_modules` directories on the host during docker image build.

If we change any files outside these subdirectories, we rebuild the image using `make`.

docker-compose.yml
```diff
@@ -33,6 +33,10 @@ services:
       - POSTGRES_HOST=db
       - POSTGRES_PORT=5432
       - POSTGRES_DB=ft_transcendence
+      - FORTYTWO_APP_ID=${FORTYTWO_APP_ID:?required}
+      - FORTYTWO_APP_SECRET=${FORTYTWO_APP_SECRET:?required}
+      - SESSION_SECRET=${SESSION_SECRET:?required}
+      - MEMCACHED_SECRET=${MEMCACHED_SECRET:?required}
```
gbudau, line 39: I moved these environment variables here because usually, we run the project with Docker Compose, and we can define them with a `.env` file.

When we run the `transcendence-app` in standalone mode, for example, for tests or in Github CI, we use the `/transcendence-app/.env.{environment}` files

scripts/docker-compose.yml
```diff
@@ -7,8 +7,7 @@ services:
       target: swagger
     volumes:
       - transcendence-app-data:${TRANSCENDENCE_APP_DATA:-/var/lib/transcendence-app}
-      - ../transcendence-app:/transcendence-app
-      - /transcendence-app/node_modules
+      - ../transcendence-app/swagger-spec:/transcendence-app/swagger-spec
```
gbudau, line 10: The same thing as above, we create a subdirectory for the Swagger spec file.

docker-compose.override.yml
Damrod: What do you think about letting docker create the detectors and then doing something like we do in `/scripts/generate-cmd.sh` lines 6 and 7 (passing MY_USER as an env var from the calling process)?

gbudau: Do you mean the `dist` and `node_modules` directories? With this change, Docker creates these directories only inside the image, and if we don't mount the entire transcendence-app directory, it doesn't create them on the host.

Damrod:
This is not working for me at least. We may look at it this evening if you can.

transcendence-app/src/swagger-main.ts
```diff
@@ -21,7 +21,7 @@ async function bootstrap() {
     .build();
   const document = SwaggerModule.createDocument(app, options);
   const yamlString: string = yaml.stringify(document, {});
-  writeFileSync('./swagger-spec.yaml', yamlString);
+  writeFileSync('./swagger-spec/swagger-spec.yaml', yamlString);
   app.close();
 }
 bootstrap();
```
Damrod, line 27: what does the .keep file do?

transcendence-app/src/main.ts
```diff
@@ -22,8 +20,6 @@ async function bootstrap() {
     .setVersion('1.0')
     .build();
   const document = SwaggerModule.createDocument(app, options);
-  const yamlString: string = yaml.stringify(document, {});
```
Damrod, line 25: We don't want anymore the spec file to regenerate when we're running the server in hot reload?

docker-compose.yml
Damrod: I had an .env file from when we ran tests for the app in production mode, I didn't change it, and it worked. Should I erase all of the fields that you don't mention here?  Wouldn't that break the prod mode on my side?

Damrod (approved):
As mentioned.

Thanks for this bugfix!

Damrod:
Please, when you merge this one, do so without squashing. There are many breaking changes, and I have merged it into my branch.

docker-compose.yml
gbudau: I think the `FORTYTWO_APP_CALLBACK_URL` for the development environment is `http://localhost:3000/api/v1/auth/login` and for production is `http://localhost/api/v1/auth/login` (different ports and possibly different domain/IP). The rest can be the same.

transcendence-app/src/swagger-main.ts
gbudau: We create the swagger spec files in this directory, and we add this empty file to add this directory to git.
Otherwise, we should create it in the swagger e2e tests or when generating the spec files.

transcendence-app/src/main.ts
gbudau: Yes, now we generate the spec file with the swagger main to separate the concerns.

gbudau:
> This is not working for me at least. We may look at it this evening if you can.

Sorry. I forgot to update some conditions in the get-swagger-spec-dependencies.sh, script, and add a volume for /transcendence-app/src directory in the swagger compose file.

gbudau:
Thanks!

PR #354, opened 2022-11-23
#317/Add generated avatar to new chatrooms

### LINKED ISSUE #270: Redesign Avatar implementation
As they are also going to be used by chat and, it could may make sense to create a module

### LINKED ISSUE #317: Add generated avatar to new chatrooms
sic

gbudau:
close #270
close #317

### DIFF
--- a/transcendence-app/src/chat/chat.service.ts
+++ b/transcendence-app/src/chat/chat.service.ts
@@ -25,6 +25,9 @@ import { ChatroomMemberService } from './chatroom/chatroom-member/chatroom-membe
 import { User } from '../user/infrastructure/db/user.entity';
 import { UpdateChatroomMemberDto } from './chatroom/chatroom-member/dto/update-chatroom-member.dto';
 import { IChatroomMemberRepository } from './chatroom/chatroom-member/infrastructure/db/chatroom-member.repository';
+import { LocalFileService } from '../shared/local-file/local-file.service';
+import { LocalFileDto } from '../shared/local-file/local-file.dto';
+import { ChatroomDto } from './chatroom/dto/chatroom.dto';
 
 @Injectable()
 export class ChatService {
@@ -34,6 +37,7 @@ export class ChatService {
     private chatroomMessageRepository: IChatroomMessageRepository,
     private chatroomMemberService: ChatroomMemberService,
     private chatroomMemberRepository: IChatroomMemberRepository,
+    private localFileService: LocalFileService,
   ) {}
 
   async retrieveChatrooms({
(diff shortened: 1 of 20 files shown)


webapp/src/shared/components/templates/AvatarPageTemplate/AvatarPageTemplate.tsx
```diff
@@ -17,7 +17,7 @@ type AvatarPageTemplateProps = {
   isNotFound: boolean;
   headerStatusVariant?: StatusVariant;
   title: string;
-  avatarProps: LargeAvatarProps;
+  avatarProps?: LargeAvatarProps;
```
gbudau, line 20: I removed the avatar from the `CreateChatroomPage` and made this prop optional.

webapp/src/pages/CreateChatroomPage/CreateChatroomPage.tsx
```diff
@@ -77,19 +72,10 @@ export default function CreateChatroomPage() {
     register().catch((e) => console.error(e));
   };
 
-  const chatroom: Partial<Chatroom> = {};
   return (
     <AvatarPageTemplate
       isLoading={false}
       title="create chatroom"
-      avatarProps={{
```
gbudau, line 85: I removed this, but we could use an editable avatar. The user uploads an image, and when they submit the form, we check if there is an avatar or generate it in the backend.

If we like the idea, we could create a ticket for this. It's nice to have, but not mandatory.

transcendence-app/src/chat/chat.service.ts
```diff
@@ -67,20 +71,27 @@ export class ChatService {
...
+        createdAt: new Date(Date.now()),
+        avatarX: 0,
+        avatarY: 0,
+        ...chatroom,
+      },
+    );
   }
 
   async createChatroom(ownerId: string, chatroom: CreateChatroomDto) {
     const { confirmationPassword: _, ...newChat } = chatroom;
```
Damrod, line 92: If you don't mind, could you refer to me to some source where I can get some information on how does this _ (underscore) syntax work? Thanks

Damrod (approved):
LGTM, thank you!

transcendence-app/src/chat/chat.service.ts
gbudau: If I'm not mistaken, this is a [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), and we assign the confirmation password to the `_` variable to signal we don't use it.

gbudau:
Thank you!

PR #358, opened 2022-11-24
#352/Implement user blocking in the chatroom messages

### LINKED ISSUE #352: Implement user blocking in the chatroom messages
If a user blocks another user, they won't see their messages anymore

gbudau:
This PR adds a new endpoint to get a list of the users blocked by the authenticated user.

When they open a chatroom page, they fetch a list of blocked users and filter the messages.

When the authenticated user blocks or unblocks a user, the server sends a block or unblock event to which the client listens and updates the list of blocked users.

This way, the messages are filtered even if a user blocks another user from another tab, browser, or device.

To do:

- [ ] We could create a context and custom hook for the block list and use it in more places

close #352

### DIFF
--- a/transcendence-app/src/user/infrastructure/db/postgres/block.postgres.repository.ts
+++ b/transcendence-app/src/user/infrastructure/db/postgres/block.postgres.repository.ts
@@ -5,6 +5,7 @@ import { PostgresPool } from '../../../../shared/db/postgres/postgresConnection.
 import { makeQuery } from '../../../../shared/db/postgres/utils';
 import { Block, BlockKeys } from '../block.entity';
 import { IBlockRepository } from '../block.repository';
+import { User, UserData, userKeys } from '../user.entity';
 
 @Injectable()
 export class BlockPostgresRepository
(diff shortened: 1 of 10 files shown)


transcendence-app/src/user/user.controller.ts
```diff
@@ -168,19 +168,30 @@ export class UserController {
     return avatar;
   }
 
+  @Get('block/list')
```
Damrod, line 171: We're not using this endpoint at the moment, are we?

What do you think we will need it for?

transcendence-app/src/user/infrastructure/db/postgres/block.postgres.repository.ts
```diff
@@ -42,4 +43,15 @@ export class BlockPostgresRepository
     });
     return blockData && blockData.length ? new this.ctor(blockData[0]) : null;
   }
+
+  async getBlocks(blockerId: string): Promise<User[] | null> {
+    const usersData = await makeQuery<UserData>(this.pool, {
```
Damrod, line 48: Why do we need to instantiate this with `UserData` instead of just plain `User`?

Damrod (approved):
I feel it's a bit suspicious if we implement the filtering of the messages of the blocked users only in the front end, but reading carefully the scale, that's what it says. The blocking user won't see the blocked users' messages. `See` is the keyword, so I a have to agree that this works. Also, I can imagine doing something else might be extremelly hard. So on my account this is good.

fizcamposancos (approved):
Nice working feature!!

vdedios (approved):
Nice 👍

transcendence-app/src/user/infrastructure/db/postgres/block.postgres.repository.ts
gbudau: We do this for readability. We type the query return to `UserData` because the result it's an object, then instantiate it to the `User` class for serialization.

gbudau:
Thank you!

Issue #353, opened 2022-11-23
⚛️ Create a subscribed chat page


gbudau:
This page lists all the chatrooms where the authenticated user is a member.

The URL is `/chat`, and will replace the current chat page.

The current chat page URL, which list all the chatrooms, becomes  `/chat/discover`

Damrod:
I think in the scope of this ticket we should have, in the `/chat` URL, not only the chatrooms where the auth user is a member, but also all one to one chats of where there is at least one message. Do we agree? For the one to one part, we will have to wait.

PR #359, opened 2022-11-25
#353/created discover chatroom page

Damrod:
As described in #353

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -20,6 +20,7 @@ import {
   ChatroomDetailsPage,
   JoinChatroomPage,
   EditChatroomMemberPage,
+  DiscoverChatPage,
 } from './pages';
 import { BrowserRouter, Route, Routes } from 'react-router-dom';
 import {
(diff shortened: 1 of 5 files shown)


Damrod:
Maybe we could like to filter from the discover page the chatrooms where the user is already a member to have two disjoint sets of chatrooms?

fizcamposancos:
> Maybe we would like to filter from the discover page the chatrooms where the user is already a member to have two disjoint sets of chatrooms?

I  would leave it like this, discover page showing everything, just in case you thought you were not a member and start searching in the discover section.
Another issue would be how we deal with one-to-one chats. The ones you are already chatting with in the chat page. And in the discover page? Only chatrooms or also all the users (as possible one-to-one chats)?

fizcamposancos:
Question: in the chat page we are ordering by chatname, each set of chatroom loaded, and in the discover page? It looks strange

fizcamposancos (approved):
Beside the comments, nice job!!

vdedios (approved):
Good 👌 👌

gbudau (approved):

gbudau:
BTW, awesome detail with the Discover/Chat button arrow direction :heart:

Issue #330, opened 2022-11-14
In user page, refactor relationship between one to one message button and block toggle


Damrod:
As of now, we have a button to send messages to other users that is available to the authenticated user in any other user's user page, depending on:

1. Whether the authenticated user is blocking the user.
2. Whether the authenticated user is being blocked by the particular user.

This basically makes the user be able to be completely aware of blocks he's the recipient of.

Please, rethink. Possible solutions may include:
1. completely removing the blocking user from the blocked user's user list OR
2. silently not delivering the sent messages from the blocked to the blocker without actually removing access to the one to one chat page OR
3. not do anything, if we all agree this is an acceptable violation of the blocking user's privacy, or if we just decide it's too much bother to change it, since the project is never going to go into production in the real world.

Damrod:
Also, according to the subject:
```
Users must be able to block other user and therefore they will not see their messages
anymore
```
Does this include the messages the blocked user will send in chatrooms where blocker and blockee are both members? Decide.

gbudau:
I vote for 2:
- silently not delivering the sent messages from the blocked to the blocker without actually removing access to the one-to-one chat page

> Does this include the messages the blocked user will send in chatrooms where blocker and blockee are both members?

I think it's only for the one-to-one chats (also because it's harder to implement for the chatrooms)

Damrod:
> Thanks, good point.
> 
> I vote for 2:
> 
> * silently not delivering the sent messages from the blocked to the blocker without actually removing access to the one-to-one chat page
> 
> > Does this include the messages the blocked user will send in chatrooms where blocker and blockee are both members?
> 
> I think it's only for the one-to-one chats (also because it's harder to implement for the chatrooms)

I'm afraid we can't skip this. If you carefully read the scale (that I won't quote here), you can see this is a must. We need blocked users have their messages hidden from the blocking user even in chatrooms.

PR #362, opened 2022-11-27
#286 create ws event to enable one-to-one messages

### LINKED ISSUE #286: Create pertinent ws events to enable real time messaging

vdedios:
closes #286

### DIFF
--- a/transcendence-app/src/chat/chat.gateway.ts
+++ b/transcendence-app/src/chat/chat.gateway.ts
@@ -19,6 +19,8 @@ import { WsAuthenticatedGuard } from '../shared/guards/ws-authenticated.guard';
 import { ChatService } from './chat.service';
 import { ChatroomMemberService } from './chatroom/chatroom-member/chatroom-member.service';
 import { CreateChatroomMessageDto } from './chatroom/chatroom-message/dto/create-chatroom-message.dto';
+import { CreateChatMessageDto } from './chat/dto/create-chat-message.dto';
+import { UserService } from '../user/user.service';
 
 @WebSocketGateway({ path: '/api/v1/socket.io' })
 @UseGuards(WsAuthenticatedGuard)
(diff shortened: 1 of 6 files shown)


transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -30,10 +32,45 @@ export class ChatGateway {
...
+    }
+
+    const message = await this.chatService.saveOneToOneChatMessage({
+      senderId,
+      recipientId,
+      content,
+    });
+
+    if (message) {
+      this.server.to(recipientId).emit('chatMessage', { ...message });
```
gbudau, line 64: suggestion: If I'm not mistaken, we also need to emit the message to the sender.

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -30,10 +32,45 @@ export class ChatGateway {
...
+    const senderId = client.request.user.id;
+
+    const block = await this.userService.getBlockRelation(
+      recipientId,
+      senderId,
+    );
+
+    if (block?.isUserBlockedByMe || block?.amIBlockedByUser) {
+      throw new WsException('There is an existing blockage. Forbidden.');
+    }
```
gbudau, line 55: question: We have an ongoing [task](https://github.com/42AGV/ft_transcendence/issues/330) about whether the blocked user should know who blocks them. What is your opinion on this matter?

transcendence-app/src/chat/chat.gateway.ts
vdedios: You mean if we the user has multiple tabs/apps wathever open right? I'll add it 👍

transcendence-app/src/chat/chat.gateway.ts
vdedios: I won't waste time improving this. Although it would be really nice the subject only mentions about not being able to see blocked user messages and the simplest way is just to prevent accessing the chat conversation like is now

gbudau: Ok, then we could also close [#330](https://github.com/42AGV/ft_transcendence/issues/330)

gbudau (approved):

transcendence-app/src/chat/chat.gateway.ts
Damrod: why do we need to spread the message if we're not modifiying it at all?

Damrod (approved):
LGTM. 👌 Go ahead!

fizcamposancos (approved):
LGTM!

transcendence-app/src/chat/chat.gateway.ts
vdedios: > suggestion: If I'm not mistaken, we also need to emit the message to the sender.

If you don't mind I will update this in next PR, I've made other backend changes

> why do we need to spread the message if we're not modifiying it at all?

Because we only draw the messages we receive by ws. I tought at first that we directly draw, in webapp, a message once we send it. But we send it, save it, and receive it. I think this is good so we only see persisted messages

transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -30,10 +32,45 @@ export class ChatGateway {
...
+
+    const message = await this.chatService.saveOneToOneChatMessage({
+      senderId,
+      recipientId,
+      content,
+    });
+
+    if (message) {
+      this.server.to(recipientId).emit('chatMessage', { ...message });
+    } else {
```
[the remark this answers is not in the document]
vdedios: Yes, but from the user perspective it couldn't be sent because no user will receive it unless is saved.

Maybe as an internal log for us could be very useful to log the db error so we knew what happened.

PR #376, opened 2022-11-29
#287 one to one chat

### LINKED ISSUE #287: Create one-to-one chat page

vdedios:
closes #287

This PR covers:
- move owned chats and discovery into `/chats` so we have `chat/<userId>` available for one-to-one messaging.
- since chatroom messaging structure was almost identical to one-to-one, move all that code into a template, make some tweaks to make it generic and reuse it in chatroom and chat messaging pages
- create messaging page (with some fixes in DB and chat gateway that had to be done for it to work)


Damrod:
It doesn't render:
Screenshot from 2022-11-29 22-49-53

webapp/src/shared/components/templates/ChatMessagingTemplate/ChatMessagingTemplate.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <div className="chat-template">
+      <div className="chat-template-loading">
+        <Loading />
+      </div>
+    </div>
+  );
+}
+
+export default function ChatMessagingTemplate({
```
gbudau, line 38: praise: Thank you for adding these templates. Very nice :+1:

Damrod: Is it working for you?

gbudau:
@Damrod, try to generate the webapp code (`make gen`) because there are new entities.

webapp/src/shared/components/templates/ChatMessagingTemplate/ChatMessagingTemplate.tsx
gbudau: I had the same error, and generating the webapp code fixed it.

Damrod: I was in the wrong branch (vdedios/one-to-one-chat-page rather than vdedios/one-to-one-chat)

Damrod:
If we enter a url such as `http://localhost:3000/chat/{NonExistingUser}` we stay in loading page forever. I think we should redirect to not found page.

webapp/src/shared/components/templates/ChatMessagingTemplate/ChatMessagingTemplate.tsx
vdedios: Yes sorry, I started working on one branch and switched to other to easily solve conflicts

vdedios:
> If we enter a url such as http://localhost:3000/chat/{NonExistingUser} we stay in loading page forever. I think we should redirect to not found page.

That should be fixed now. Thanks!

Damrod (approved):
Lgtm! Thanks!

vdedios:
> If I start a one to one chat with someone, from this user profile clicking the "send message button, then this chat should be visible in the general chatpage, shoudn't it?
> I can not see it. The chat list is empty

Yes you are right, that part is missing. But is another task. I will create a ticket and work on it, thanks for pointing out!

PR #383, opened 2022-12-01
#373/Create-database-schema-migrations

### LINKED ISSUE #373: Create database schema migrations
Migrations allow us to define sets of schema changes. So upgrading a database is a breeze.

gbudau:
close #373


Damrod:
I think this may need a readme update

Makefile
```diff
@@ -83,4 +85,12 @@ get-ip:
 
 .PHONY: seed
 seed:
-	$(DOCKER_COMPOSE) exec -it transcendence-app npx knex seed:run
+	$(DOCKER_COMPOSE) exec transcendence-app npx knex seed:run
```
gbudau, line 88: I removed the `-it` flags because it doesn't work on campus `docker-compose` v1.

transcendence-app/migrations/20221201150119_add_localfile_table.ts
```diff
@@ -0,0 +1,21 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
```
vdedios, line 3: Aside from filling up our tables on startup, could you elaborate what other benefits we gain? (new for me and wish to learn :p )

gbudau: Currently, the database initialization scripts are only run if we start the container with an empty data directory.
And when we want to edit the database schema, we delete the volumes and restart the containers, which run the initialization scripts.

Or, if we want preserve the data in our database, we could make schema changes manually without deleting the volumes.

With migrations, we can update the database schema automatically without nuking the database or doing migrations manually.

If we want to add a new table, update the column or a table, or make any schema change, we can create a migration that has `up` and `down` functions, and we can run the following CLI commands:
```shell
knex migrate:latest # run all the migrations from the current version until the latest one
knex migrate:rollback # rollback to the previous version
knex migrate:up # run one migration, from the current version to the next version
knex migrate:down # rollback one migration from the current version to the previous version
```

Damrod: As soon as you do a down, you nuke the table, at least in the current implementation. I would expect a system to map items from a previous table scheme to a new table scheme, that would run for each of the items in one of the tables, for each of the tables.
How do you do that? What am I missing?

vdedios: I have another question, this means with migrations we would be able to deploy new versions without affecting already existing data in the DB? (even if we make an schema upgrade)

gbudau: > As soon as you do a down, you nuke the table, at least in the current implementation. I would expect something like a system to map items from a previous table scheme to a new table scheme, that would run for each of the items in one of the tables, for each of the tables, that saves the old data in memory while it does the swap, and then removes the old invalid data.
> How do you do that? What am I missing?

I don't understand this question. Could you clarify, please?
The current migrations create tables. The opposite of creating a table is deleting the table (drop table).

gbudau: > I have another question, this means with migrations we would be able to deploy new versions without affecting already existing data in the DB? (even if we make an schema upgrade)

Yes, we would be able to do that.

Makefile
```diff
@@ -39,13 +41,13 @@ gen:
...
 clean:
-	$(DOCKER_COMPOSE) down -v
+	make down
 	rm -rf $(PROJECT_ROOT)/webapp/src/shared/generated || true
 	rm $(PROJECT_ROOT)/transcendence-app/swagger-spec/swagger-spec.yaml || true
 
 .PHONY: prod-clean
 prod-clean:
-	$(DOCKER_COMPOSE) -f docker-compose.yml -f docker-compose.prod.yml down -v
+	make prod-down
```
vdedios, line 50: Not very sure if I am mistaken but, with this you'll finally execute:
`	$(DOCKER_COMPOSE) -f docker-compose.yml -f docker-compose.prod.yml down`
which will remove everything (including volumes where our DB data is stored)

So everytime we re-deploy our app all data will be deleted, right? (Despite of the knex migrations)

docker-compose.yml
```diff
@@ -42,17 +42,9 @@ services:
...
     environment:
       - POSTGRES_DB=ft_transcendence
 
   memcached:
     image: memcached:1.6.15-alpine3.16
     restart: always
-
-volumes:
-  transcendence-db:
-  transcendence-app-data:
```
vdedios, line 58: Why did you differentiate between dev and prod volumes?

scripts/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       target: swagger
     volumes:
       - transcendence-app-data:${TRANSCENDENCE_APP_DATA:-/var/lib/transcendence-app}
-      - ../transcendence-app/src:/transcendence-app/src
+      - ../transcendence-app/src:/transcendence-app/src:ro
```
vdedios, line 10: What does this `ro` mean?

vdedios (approved):
Aside from comments, very nice work! Thanks for working on this 🚀

docker-compose.yml
gbudau: Knex saves the migration filenames and executes them once. Unfortunately, it also saves the extension. In development environment, we use typescript extensions, and in production javascript.

If we want to test both dev and prod environment on the same machine, it throws the following error:

```
Error: The migration directory is corrupt, the following files are missing: migration.[ts, js]*
```

* `.ts` or `.js` depending on which environment we run first

scripts/docker-compose.yml
gbudau: It means [read-only](https://docs.docker.com/compose/compose-file/#short-syntax-4) access.

More info here https://docs.docker.com/storage/volumes/#use-a-read-only-volume.

gbudau:
Thank you!

PR #384, opened 2022-12-01
#364/Create a context for the authenticated user blocklist

### LINKED ISSUE #364: ⚛️ Create a React context and a custom hook for the authenticated user blocklist
We could use the authenticated user blocklist in more than one place in our webapp.

gbudau:
close #364

### DIFF
--- a/webapp/src/shared/components/templates/ChatMessagingTemplate/components/ChatMessages/ChatMessages.tsx
+++ b/webapp/src/shared/components/templates/ChatMessagingTemplate/components/ChatMessages/ChatMessages.tsx
@@ -5,6 +5,7 @@ import { AVATAR_EP_URL } from '../../../../../urls';
 import { ENTRIES_LIMIT } from '../../../../../constants';
 import { useIsElementVisible } from '../../../../../hooks/UseIsElementVisible';
 import { useAuth } from '../../../../../hooks/UseAuth';
+import { useBlocklist } from '../../../../../hooks/UseBlocklist';
 
 import './ChatMessages.css';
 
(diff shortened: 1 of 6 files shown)


webapp/src/shared/components/templates/ChatMessagingTemplate/components/ChatMessages/ChatMessages.tsx
```diff
@@ -40,7 +40,7 @@ export default function ChatMessages({
   return (
     <ul className="chat-template-messages-list">
       {messages
-        .filter((message) => !blocks.has(message.userId))
+        .filter((message) => !userBlocks(message.userId))
```
vdedios, line 43: Nice!!

webapp/src/shared/context/UserBlocklistContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }),
+    [userBlocks],
+  );
+
+  return (
+    <UserBlocklistContext.Provider value={contextValue}>
+      {children}
+    </UserBlocklistContext.Provider>
+  );
+};
```
vdedios, line 82: Why do you need to memoize the callback if you are already wrapping it in a `useCallback` ? I'll just do:

```suggestion
  return (
    <UserBlocklistContext.Provider value={userBlocks}>
      {children}
    </UserBlocklistContext.Provider>
  );
};
```

vdedios (approved):
Nice feature, thanks!! 😃

webapp/src/shared/context/UserBlocklistContext.tsx
gbudau: Update. Thanks for the suggestion.

gbudau:
Thank you!

PR #385, opened 2022-12-02
#355/create an endpoint to edit the avatar of a chatroom

### LINKED ISSUE #355: Create an endpoint to edit the avatar of a chatroom
Create a `PUT/chat/room/{roomId}/avatar` endpoint to update a chatroom avatar
This endpoint must be protected
Only the owner can update the avatar

gbudau:
close #355

### DIFF
--- a/transcendence-app/src/shared/avatar/dto/avatar.response.dto.ts
+++ b/transcendence-app/src/shared/avatar/dto/avatar.response.dto.ts
@@ -1,7 +1,7 @@
 import { StreamableFile } from '@nestjs/common';
 import { IsNotEmpty, IsUUID } from 'class-validator';
 
-export class UserAvatarResponseDto {
+export class AvatarResponseDto {
   @IsUUID()
   avatarId!: string;
 
--- a/transcendence-app/src/shared/avatar/interceptors/avatar.file.interceptor.ts
+++ b/transcendence-app/src/shared/avatar/interceptors/avatar.file.interceptor.ts
@@ -1,0 +1,27 @@
+import { UnprocessableEntityException } from '@nestjs/common';
+import {
+  AVATARS_PATH,
+  AVATAR_MAX_SIZE,
+  AVATAR_MIMETYPE_WHITELIST,
+} from '../constants';
+import LocalFileInterceptor from '../local-file/local-file.interceptor';
+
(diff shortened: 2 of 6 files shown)


transcendence-app/src/shared/avatar/avatar.service.ts
```diff
@@ -40,7 +44,17 @@ export class AvatarService {
       fileTypeResult && AVATAR_MIMETYPE_WHITELIST.includes(fileTypeResult.mime);
     if (!isValid) {
       this.localFileService.deleteFileData(path);
+      const allowedTypes = AVATAR_MIMETYPE_WHITELIST.join(', ');
+      throw new UnprocessableEntityException(
+        `Validation failed (allowed types are ${allowedTypes})`,
+      );
```
vdedios, line 50: Nice but isn't this evaluated in the interceptor?

vdedios (approved):
👍

transcendence-app/src/chat/chat.controller.ts
```diff
@@ -380,4 +389,40 @@ export class ChatController {
...
+  @ApiConsumes('multipart/form-data')
+  @ApiFile('file')
+  @ApiProduces('image/jpeg')
+  @ApiOkResponse({
+    description: 'Update a chatroom avatar',
+    type: AvatarResponseDto,
+  })
+  @ApiNotFoundResponse({ description: 'Not Found' })
+  @ApiUnprocessableEntityResponse({ description: 'Unprocessable Entity' })
+  @ApiPayloadTooLargeResponse({ description: 'Payload Too Large' })
```
fizcamposancos, line 403: Is this decorator used automatically with a certain file size? Where is defined? Nestjs, Db, server?

transcendence-app/src/shared/avatar/avatar.service.ts
gbudau: The interceptor checks the mime type, and this function uses the `file-type` library to check the binary format.

For example, someone could rename a `.pdf` file to `.jpeg`, and the mime type is `image/jpeg`. In this case, the mime type, according to the binary format, is `application/pdf`

gbudau:
Thank you!

PR #397, opened 2022-12-12
add roles postgres repository

Damrod:
Signed-off-by: Alvaro Ollero <aollero@gmail.com>

### DIFF
--- a/transcendence-app/src/user/infrastructure/db/postgres/user-to-role.postgres.repository.ts
+++ b/transcendence-app/src/user/infrastructure/db/postgres/user-to-role.postgres.repository.ts
@@ -1,0 +1,49 @@
+import { UserToRole, userToRoleKeys } from '../user-to-role.entity';
+import { UserWithRoles, UserWithRolesData } from '../user-with-role.entity';
+import { Role } from '../../../../shared/enums/roles.enum';
+import { Injectable } from '@nestjs/common';
+import { BasePostgresRepository } from '../../../../shared/db/postgres/postgres.repository';
+import { IUserToRoleRepository } from '../user-to-role.repository';
+import { PostgresPool } from '../../../../shared/db/postgres/postgresConnection.provider';
+import { table } from '../../../../shared/db/models';
+import { makeQuery } from '../../../../shared/db/postgres/utils';
+
+@Injectable()
+export class UserToRolePostgresRepository
+  extends BasePostgresRepository<UserToRole>
+  implements IUserToRoleRepository
+{
+  constructor(protected pool: PostgresPool) {
+    super(pool, table.USERTOROLE, UserToRole);
+  }
+
+  async addUserToRole(userToRole: UserToRole): Promise<UserToRole | null> {
+    return this.add(userToRole);
+  }
+
+  async getUserWithRoles(id: string): Promise<UserWithRoles | null> {
+    const userData = await makeQuery<UserWithRolesData>(this.pool, {
+      text: `SELECT u.*,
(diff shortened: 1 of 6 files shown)


transcendence-app/src/user/infrastructure/db/postgres/user-to-role.postgres.repository.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  async addUserToRole(userToRole: UserToRole): Promise<UserToRole | null> {
+    return this.add(userToRole);
+  }
+
+  async getUserWithRoles(id: string): Promise<UserWithRoles | null> {
+    const userData = await makeQuery<UserWithRolesData>(this.pool, {
+      text: `SELECT u.*,
+                    array_agg(ur.${userToRoleKeys.ROLE}) FILTER (WHERE ur.${userToRoleKeys.ROLE} IS NOT NULL) as roles
```
gbudau, line 27:
```suggestion
                    coalesce( array_agg(ur.${userToRoleKeys.ROLE}) FILTER (WHERE ur.${userToRoleKeys.ROLE} IS NOT NULL), '{}' ) as roles
```

praise: I like how you implement this method and query

question: What do you think about removing the `none` role and using the [coalesce](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL) function to return an empty array when the user doesn't have a role?

transcendence-app/src/user/user.service.ts
```diff
@@ -250,4 +249,80 @@ export class UserService {
...
+    const userToRole = await this.userToRoleRepository.addUserToRole(user);
+    if (!userToRole) {
+      throw new UnprocessableEntityException();
+    }
+    return userToRole;
+  }
+
+  async addUserToRoleFromUsername(
+    username: string,
+    role: string,
```
gbudau, line 287:
```suggestion
    role: Role,
```

suggestion: If I'm not mistaken, we could use the `Role` type here and in the `deleteUserToRoleFromUsername()` method

transcendence-app/src/user/user.service.ts
```diff
@@ -250,4 +249,80 @@ export class UserService {
     }
     return friends;
   }
+
+  async getUserWithRolesFromUsername(
```
gbudau, line 253: praise: Thank you for adding username methods

transcendence-app/src/user/user.service.ts
```diff
@@ -250,4 +249,80 @@ export class UserService {
...
+    username: string,
+  ): Promise<UserWithRoles | null> {
+    const user = await this.userRepository.getByUsername(username);
+    if (!user) {
+      throw new UnprocessableEntityException();
```
gbudau, line 258: suggestion: For the get and delete methods, we could throw a `NotFoundException()` when there is no user with that `id`/`username`

gbudau (approved):
Thank you for this feature and improvements! LGTM :rocket:

transcendence-app/src/user/user.service.ts
Damrod: done

transcendence-app/src/user/user.service.ts
Damrod: I'm not sure about this (I'd prefer the calling method to not have to be aware of the Role type), but I don't mind much anyway, so it's done. I'll keep changing everything in the next PR.

transcendence-app/src/user/infrastructure/db/postgres/user-to-role.postgres.repository.ts
Damrod: done

Issue #393, opened 2022-12-08
Add status and friends relation


fizcamposancos:
Status of users (online, offline, playing) should only be visible to user friends

PR #402, opened 2022-12-16
#393 Add status and friend relation

fizcamposancos:


webapp/src/shared/context/UserFriendContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+type FriendId = string;
+
+export interface UserFriendContextType {
+  userFriends: (userId: string) => boolean;
+}
+
+export const UserFriendContext = createContext<UserFriendContextType>(null!);
+
+export const UserFriendProvider = ({ children }: { children: ReactNode }) => {
```
gbudau, line 19: praise: I like how you implemented the context :+1:

webapp/src/pages/UserPage/UserPage.css
```diff
@@ -2,6 +2,14 @@
   display: contents;
 }
 
+.user-page-loading {
```
gbudau, line 5: question: There is a loading class in the `AvatarPageTemplate`. Maybe we could remove this one?

gbudau (approved):

webapp/src/pages/UserPage/UserPage.css
fizcamposancos: You are right, this is a leftover of some tests I did before.

transcendence-app/src/user/user.controller.ts
```diff
@@ -242,7 +242,7 @@ export class UserController {
   unfollowUser(
     @GetUser() user: User,
     @Param('userId', ParseUUIDPipe) followedUserId: string,
-  ): Promise<Friend> {
+  ): Promise<void> {
```
Damrod, line 245: Why do you remove the return object of these two methods? I see that across the rest of our codebase, we do return the object when we delete, or add, and it makes sense, because we enable the app that consumes the API (in this case webapp), to check that the API is doing what it expects to do.

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -11,12 +11,13 @@ import {
 } from '../../shared/components';
 import { AVATAR_EP_URL, CHAT_URL } from '../../shared/urls';
 import { useData } from '../../shared/hooks/UseData';
-import React, { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
```
Damrod, line 14: why do you remove React from the import list?
When you remove it:
```
'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.ts(2686)`
```

Damrod (approved):
besides the comments, lgtm! Thanks!

transcendence-app/src/user/user.controller.ts
gbudau: In this case, we could check the response status -> 206 (No Content) when the request is successful and 422 (Unprocessable Entity) or 404 (Not Found) when there was an error.

Similar to [Github API](https://docs.github.com/en/rest/users/followers?apiVersion=2022-11-28#follow-a-user)

webapp/src/pages/UserPage/UserPage.tsx
gbudau: Is this an error from the IDE? According to [this](https://stackoverflow.com/questions/64656055/react-refers-to-a-umd-global-but-the-current-file-is-a-module) StackOverflow thread, importing React is no longer required since version 17.

Damrod: It's a symbol we don't use here, so it's totally fine. If I do, for instance, `React.UseEffect(...` then my IDE does complain. I guess my IDE is wrong.

Issue #405, opened 2022-12-19
Implement a policies guard in the authorization module


Damrod:
As suggested [here](https://docs.nestjs.com/security/authorization#advanced-implementing-a-policiesguard) we can investigate if this policies guard offers value for our goals, and possibly implement it

PR #406, opened 2022-12-21
#405/add policies guard

Damrod:
as described in #405


gbudau (approved):
I like how you make use of decorators and pipes.

Issue #404, opened 2022-12-19
Add endpoints to grant admin and banned roles to users


Damrod:
In the context of the global admin piece of work, we need a way for the admin to be able to grant and remove rights to other users. This ticket tackles this.

PR #407, opened 2022-12-22
#404/create post and delete user roles eps

Damrod:
as described in #404


transcendence-app/src/auth/auth.controller.ts
```diff
@@ -93,4 +103,32 @@ export class AuthController {
   loginLocalUser(@Req() req: Request, @Body() user: LoginUserDto): User {
     return req.user;
   }
+
+  @Post('local/role')
```
gbudau, line 107: praise: I like how you implemented these endpoints

question: What do you think about moving the role routes to the authorization controller?

`local/role` -> `authorization/role`

transcendence-app/src/authorization/infrastructure/db/user-to-role.entity.ts
```diff
@@ -11,11 +12,19 @@ export interface UserToRoleData {
...
-  id: string;
-  role: Role;
+  @IsString()
+  @IsUUID()
+  @IsNotEmpty()
+  id!: string;
+
+  @IsEnum(Role)
+  @IsNotEmpty()
+  role!: Role;
```
gbudau, line 22: suggestion: We could remove the `@IsNotEmpty` and `@IsString` decorators

question: What do you think about creating a `UserToRoleDto`?

gbudau (approved):

transcendence-app/src/auth/auth.controller.ts
Damrod: Yes, why not.

transcendence-app/src/authorization/infrastructure/db/user-to-role.entity.ts
Damrod: I don't like duplicated code, and UserToRole Dto would be the third object with that shape, and the second object that implements the usertoroledata interface. But I guess it's more readable if it exists, and so I have made the change you requested.

PR #408, opened 2022-12-22
#404/fix instantiation of the subject in the global guard

Damrod:
There is a bug, that this PR attempts to fix, where the global policies guard is not actually instantiating the usertoroleDto (Because the userToRoleDto doesn't have a constructor) class to pass it to the `ability.can(...,usertoroledto)`. This fixes it by using the user-to-role.entity instead for the casl logic, separating it from the data transfer (that still uses the dto).


gbudau (approved):
Thank you! LGTM

Issue #409, opened 2022-12-23
Retrieve user with roles endpoint


Damrod:
As part of the effort to have an authorization system, we need a way to access this from the front end.

PR #410, opened 2022-12-23
#409/get user with roles endpoint

Damrod:
as described in #409


Damrod:
what if there is a username that is named 'role'. Well, good question. Proceeding to fix it.

gbudau (approved):

PR #413, opened 2022-12-24
#400 / pong poc

### LINKED ISSUE #400: Develop an MVP for a pong game feature
We decided to try to develop al pong game MVP to get acquainted with WebSockets, Canvas API and other technologies.

vdedios:
~⚠️ working version with minimal bugs but still to be refactored since it's a bit spaghetti~
Code already refactored

- working with arrow keys and touch screen by sliding the paddle

https://user-images.githubusercontent.com/57688116/209722818-011d7130-48a1-4fdb-bd51-16b8a8787eab.mov

closes #400


webapp/src/shared/components/Game/Game.tsx
```diff
@@ -0,0 +1,348 @@
...
+  height: number;
+};
+
+type GameBall = GameElement & {
+  vx: number;
+  vy: number;
+  radius: number;
+};
+
+type Coord = {
```
Damrod, line 39: we already have this type here webapp/src/shared/types.ts (Position, but you can rename it if you want).

Damrod:
The slider stays inside the "court" when using the keyboard, but can go out of the court when using the touchscreen. Is this intentional?

webapp/src/shared/components/Game/Game.tsx
vdedios: You are right, but I prefer to have this additional type since it represent a different thing. Position is used for the position in the screen and Coord is for the physics of the game inside the canvas (the game "court"), thanks!

vdedios:
> The slider stays inside the "court" when using the keyboard, but can go out of the court when using the touchscreen. Is this intentional?

Is not intentional, is a thing that I have to fix, the drag part is a bit tricky 😓. Thanks for pointing out!

vdedios:
Aside from the bug @Damrod noticed, and another thing I want to fix (delta time, which is currently being assumed as 60fps and should be calculated with each screen) this POC ready to review

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import { CANVAS_WIDTH, CANVAS_HEIGHT } from './constants';
+import { GameBall, GamePaddle } from './types';
+import { initialBallState, initialPaddleState, isLose, isWin } from './physics';
+import useGameControls from './useGameControls';
+import useGameAnimation from './useGameAnimation';
+
+import './Game.css';
+
+export default function Game() {
```
gbudau, line 11: suggestion: What do you think about renaming this component to `PongTraining`/`GameTraining` and adding a button to navigate on the `PlayPage` because when we implement the real version we will move the game logic in the backend?

webapp/src/shared/components/Game/useGameControls.tsx
```diff
@@ -0,0 +1,66 @@
...
+        };
+      }
+    },
+    [paddleRef],
+  );
+
+  window.addEventListener('keydown', movePaddle, false);
+  window.addEventListener('keyup', stopPaddle, false);
+  window.addEventListener('touchmove', dragPaddle, false);
+  window.addEventListener('contextmenu', (e) => e.preventDefault());
```
gbudau, line 61: issue: This could potentially memory leak if we don't remove the event listeners

suggestion: We could place the `window.addEventListener()` calls in a `useEffect` and call `window.removeEventListener()` on cleanup

Here is an example: https://beta.reactjs.org/reference/react/useEffect#listening-to-a-global-browser-event

gbudau (approved):
Thank you for this feature! Cool game mode, LGTM :rocket:

webapp/src/shared/components/Game/useGameControls.tsx
vdedios: not sure if that would cause a memory leak but a registered listener which we are not going to use so, agree and removed 👍

webapp/src/shared/components/Game/Game.tsx
vdedios: For now I'll keep it as is since we don't have anything else. When we create the real game page we can move this easily it and rename it as well ;)

Damrod:
When playing with the keyboard arrows, you are at comparative disadvantage VS playing with touch: you can slide much faster with the touch than with the arrows. Do we like this? Maybe increasing the paddle movement acceleration when using arrows? Maybe nothing? Not sure.

webapp/src/pages/PlayPage/PlayPage.css
```diff
@@ -0,0 +1,20 @@
...
+  grid-column-gap: var(--column-gap);
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-template-rows: auto 1fr auto;
+  grid-row-gap: calc(var(--row-gap) * 1.5);
+  overflow: scroll;
+}
+
+@media all and (min-width: 768px) {
+  .avatar-page-template .header {
```
Damrod, line 16: why are you changing here the class="avatar-page-template" ?

vdedios:
> When playing with the keyboard arrows, you are at comparative disadvantage VS playing with touch: you can slide much faster with the touch than with the arrows. Do we like this? Maybe increasing the paddle movement acceleration when using arrows? Maybe nothing? Not sure.

Yes, this is due to the drag mode being just a follow up of your finger. This is, the paddle will move as fast as you move your finger (which is the most intuitive I've figured out)

Using the arrows you have a constant speed because you are just pressing a button, no more input than that. But sure, I've increased it a bit since is defined in a constant. We can tweak it as much as we like

webapp/src/pages/PlayPage/PlayPage.css
vdedios: Ouch, leftovers, removed. Thanks!

vdedios:
I've fixed the paddle going out of the court when dragging and I've also updated it to calculate time in each frame instead of assuming a constant 60fps. This way the game should go now as fast in a 60Hz or a 45Hz display

Damrod:
I have seen three or four times an issue with the bouncing of the ball, where it will "stick" to some wall (horizontal or vertical), and stay there for a while. I've managed to record a clip with this effect, I'll post it when I can.

Damrod:
You can see the issue I'm reporting above by the end of the following [youtube short](https://www.youtube.com/shorts/Gc-Sck5u9rM) I've recorded

Damrod:
Actually, I've just seen it again, and when it hits the brick, it's stuck to the back wall (moving slowly along it) and it hits the brick, it doesn't hit the brick only once per traversal. It hits it about five to ten times. It's as if the ball, when hitting any wall at a particular angle, got trapped between the visible wall, and some invisible wall very very close to the visible wall, and parallel to it.
I think I've seen it without a wall, when it hits one of the corners in a very particular angle, in will bounce back at 45 degrees (about 45, but let's say some oblique angle), but it will move at a much slowe pace, as if it was bouncing in two invisible walls that are very close to each other.

Damrod (approved):
I've made some comments that I think are worth at least opening an issue on the topic, but I don't think this should stop you from merging as we are. Thank you!

vdedios:
About all your above comments @Damrod, they should be fixed now. The problem was when I introduced time calculations based on screen rate I was just getting raw times which, sometimes, got very little or huge values and lead to that crazy behaviour.

We needed a constant time for each animation render so, instead of calculating the delta time for each frame, I've changed to calculate the fps of the screen when the game starts and use that constant value for the rest of the game.

Everything should work now 😄 thanks for your comments!

Issue #411, opened 2022-12-23
⚛️ add optional admin tab to navigation bar


Damrod:
If a user is an owner or global moderator, it will need a way to navigate to the administrator pane. This issue tackles that

PR #414, opened 2022-12-24
#411/add admin item for admin users to navbar

Damrod:
I broke the previous [PR](https://github.com/42AGV/ft_transcendence/pull/412), so I'm opening this one again, with exactly the same content it had


gbudau (approved):

PR #418, opened 2022-12-27
#417/allow admins to see chatrooms they havent joined

Damrod:
As described in #417

### DIFF
--- a/transcendence-app/src/authorization/casl-ability.factory.ts
+++ b/transcendence-app/src/authorization/casl-ability.factory.ts
@@ -1,7 +1,7 @@
 import {
   AbilityBuilder,
-  AnyMongoAbility,
   createMongoAbility,
+  ExtractSubjectType,
   InferSubjects,
   MongoAbility,
 } from '@casl/ability';
@@ -25,13 +25,16 @@ export type SubjectCtors =
   | typeof UserWithAuthorizationResponseDto
   | typeof ChatroomMessageWithUser;
 
-export type Subject = InferSubjects<SubjectCtors>;
+export type Subject =
+  | InferSubjects<SubjectCtors>
+  | 'all'
+  | 'readChatroomMessagesList';
 
 type AppAbility = MongoAbility<[Action, Subject]>;
 @Injectable()
 export class CaslAbilityFactory {
   private setGlobalAbilities(
-    { can, cannot, build }: AbilityBuilder<AnyMongoAbility>,
+    { can, cannot, build }: AbilityBuilder<AppAbility>,
     globalUserAuthCtx: UserWithAuthorization,
   ) {
     cannot(Action.Read, UserWithAuthorizationResponseDto);
@@ -48,7 +51,8 @@ export class CaslAbilityFactory {
       role: Role.owner,
     });
     return build({
-      detectSubjectType: (object) => object.constructor,
+      detectSubjectType: (object) =>
+        object.constructor as ExtractSubjectType<Subject>,
     });
   }
 
@@ -104,6 +108,7 @@ export class CaslAbilityFactory {
           can(Action.Read, ChatroomMessageWithUser, {
             chatroomId: user.chatId,
           });
+          can(Action.Read, 'readChatroomMessagesList');
         }
         if (user.crm_owner) {
           can(Action.Update, Chatroom, { id: user.chatId });
--- a/transcendence-app/src/authorization/casl-ability.factory.ts
+++ b/transcendence-app/src/authorization/casl-ability.factory.ts
@@ -13,13 +13,13 @@ import { ChatroomMember } from '../chat/chatroom/chatroom-member/infrastructure/
 import { UpdateChatroomMemberDto } from '../chat/chatroom/chatroom-member/dto/update-chatroom-member.dto';
 import { Chatroom } from '../chat/chatroom/infrastructure/db/chatroom.entity';
 import { Role } from '../shared/enums/role.enum';
-import { UserToRoleDto } from './dto/user-to-role.dto';
+import { UserToRole } from './infrastructure/db/user-to-role.entity';
 
 export type SubjectCtors =
   | typeof ChatroomMember
   | typeof UpdateChatroomMemberDto
   | typeof Chatroom
-  | typeof UserToRoleDto;
+  | typeof UserToRole;
 
 export type Subject = InferSubjects<SubjectCtors>;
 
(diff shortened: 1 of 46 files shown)


webapp/src/pages/ChatroomPage/ChatroomPage.tsx
```diff
@@ -32,9 +32,14 @@ export default function ChatroomPage() {
...
   chatroomId: string;
   authUser: UserWithAuthorizationResponseDto;
+  overridePermissions?: boolean;
 };
 
-function Chatroom({ chatroomId, authUser }: ChatroomProps) {
+function Chatroom({
+  chatroomId,
+  authUser,
+  overridePermissions = false,
```
Damrod, line 41: This is a way to have the two setups: normal, where it works as it should for regular users, even if the user is admin (no magic powers), and elevated privileges so that it can be seen fully when navigating to this page from the admin tab

transcendence-app/src/chat/chat.controller.ts
```diff
@@ -335,7 +335,7 @@ export class ChatController {
     if (!messages) {
       throw new ServiceUnavailableException();
     }
-    return messages;
+    return messages.filter((message) => ability.can(Action.Read, message));
```
Damrod, line 338: this is not strictly necessary in our current setup, but it could be useful in the unlikely case when we wanted to implement the filtering of blocked users messages here. It could be made within the casl framework, I believe,

transcendence-app/src/authorization/casl-ability.factory.ts
```diff
@@ -96,9 +98,17 @@ export class CaslAbilityFactory {
 
       can(Action.Create, Chatroom);
       can(Action.Read, Chatroom);
-      if (user.crm_member && user.crm_owner) {
-        can(Action.Update, Chatroom);
-        can(Action.Delete, Chatroom);
+      if (user.crm_member) {
+        if (!user.crm_banned) {
+          can(Action.Join, Chatroom, { id: user.chatId });
```
Damrod, line 103: this is redundant. If user.crm_member, then {id: user.chatId} always. But I think it's somehow clearer like this. Let me know if you disagree

transcendence-app/src/chat/chat.controller.ts
```diff
@@ -313,19 +315,17 @@ export class ChatController {
...
     const chatroom = await this.chatService.getChatroomById(chatroomId);
     if (!chatroom) {
       throw new NotFoundException();
     }
-    const isChatMember = await this.chatroomMemberService.getById(
-      chatroomId,
-      user.id,
-    );
-    if (!isChatMember || isChatMember.banned) {
+    if (ability.cannot(Action.Join, chatroom)) {
```
gbudau, line 328: thought: I found this confusing because we check if the user can join a chatroom from an endpoint that lists the chatroom messages.

suggestion: What do you think about adding another `Subject` type (`'readChatroomMessagesList'`?) and using it for readability?

For example:

<details>

```diff
diff --git a/transcendence-app/src/authorization/casl-ability.factory.ts b/transcendence-app/src/authorization/casl-ability.factory.ts
index c799a6d..8625be5 100644
--- a/transcendence-app/src/authorization/casl-ability.factory.ts
+++ b/transcendence-app/src/authorization/casl-ability.factory.ts
@@ -1,7 +1,7 @@
 import {
   AbilityBuilder,
-  AnyMongoAbility,
   createMongoAbility,
+  ExtractSubjectType,
   InferSubjects,
   MongoAbility,
 } from '@casl/ability';
@@ -25,13 +25,16 @@ export type SubjectCtors =
   | typeof UserWithAuthorizationResponseDto
   | typeof ChatroomMessageWithUser;
 
-export type Subject = InferSubjects<SubjectCtors>;
+export type Subject =
+  | InferSubjects<SubjectCtors>
+  | 'all'
+  | 'readChatroomMessagesList';
 
 type AppAbility = MongoAbility<[Action, Subject]>;
 @Injectable()
 export class CaslAbilityFactory {
   private setGlobalAbilities(
-    { can, cannot, build }: AbilityBuilder<AnyMongoAbility>,
+    { can, cannot, build }: AbilityBuilder<AppAbility>,
     globalUserAuthCtx: UserWithAuthorization,
   ) {
     cannot(Action.Read, UserWithAuthorizationResponseDto);
@@ -48,7 +51,8 @@ export class CaslAbilityFactory {
       role: Role.owner,
     });
     return build({
-      detectSubjectType: (object) => object.constructor,
+      detectSubjectType: (object) =>
+        object.constructor as ExtractSubjectType<Subject>,
     });
   }
 
@@ -104,6 +108,7 @@ export class CaslAbilityFactory {
           can(Action.Read, ChatroomMessageWithUser, {
             chatroomId: user.chatId,
           });
+          can(Action.Read, 'readChatroomMessagesList');
         }
         if (user.crm_owner) {
           can(Action.Update, Chatroom, { id: user.chatId });
diff --git a/transcendence-app/src/chat/chat.controller.ts b/transcendence-app/src/chat/chat.controller.ts
index 8fffba0..1d16c7f 100644
--- a/transcendence-app/src/chat/chat.controller.ts
+++ b/transcendence-app/src/chat/chat.controller.ts
@@ -325,7 +325,7 @@ export class ChatController {
     if (!chatroom) {
       throw new NotFoundException();
     }
-    if (ability.cannot(Action.Join, chatroom)) {
+    if (ability.cannot(Action.Read, 'readChatroomMessagesList')) {
       throw new ForbiddenException();
     }
     const messages = await this.chatService.getChatroomMessagesWithUser(
diff --git a/transcendence-app/src/chat/chat.gateway.ts b/transcendence-app/src/chat/chat.gateway.ts
index b29c77b..ccbb143 100644
--- a/transcendence-app/src/chat/chat.gateway.ts
+++ b/transcendence-app/src/chat/chat.gateway.ts
@@ -125,6 +125,9 @@ export class ChatGateway {
     @ConnectedSocket() client: Socket,
   ) {
     const cr = await this.chatService.getChatroomById(chatroomId);
+    if (!cr) {
+      throw new WsException('Chatroom not found');
+    }
     const crm = await this.authorizationService.getUserAuthContextForChatroom(
       client.request.user.id,
       chatroomId,
```

</details>

webapp/src/pages/ChatroomPage/ChatroomPage.tsx
gbudau: praise: I like this :+1:

gbudau (approved):

vdedios (approved):
I still have to catch up with this admin PRs but this looks nice 👌

transcendence-app/src/chat/chat.controller.ts
Damrod: I'm unconvinced. Read is an action. The subjects should be separate and orthogonal from the actions. If you embed the action in the Subject, you defeat the purpouse of the whole system, I believe. But I think my concern is mostly the naming. I like how something like your proposal enables us to create custom abstract subjects, and how we have stricter typing now.
Anyway I implemented it the way you asked.

PR #423, opened 2022-12-28
#419/fix bug when username of a 42 user already exists

### LINKED ISSUE #419: Fix bug when username already exists for given 42 oauth user
Right now, when the username doesn't exist, getUserByUsername method in User service returns null, but it also returns null when the friend relation is not relevant, and the second user is not provided. In 42oauth guard, we rely on this method returning not null when the username exists, so the part of the code that relies on this, is currently broken. Namely, trying to log in with a username whose 42 username already exists in our local database.

As proposed fixes: we could decouple the friend relation logic from the user specific data in user service, or just duplicate the user service method, removing the friend relation logic from the second one.

Damrod:
as described in #419
Closes #419


gbudau (approved):

PR #427, opened 2022-12-31
#426 friends page first

### LINKED ISSUE #396: ⚛️ remove friend toggle from userpage where user is self
We can click the friend toggle in the user page where the user is the same as the authenticated user, but it has no effect and it's not persisted. It should be removed from the UI.

### LINKED ISSUE #426: Make friends page first
- Right now we land in the general "users" page. Since we have a friends page, I think is more valuable for the user to land in the list her friends since those are the players she wants to keep an eye on. The users page should be for searching new friends

- ~Also, we have a bug were add friend toggle button is displayed for "me". But toggling it we get an unprocessable entity response and, hence, do nothing.~ (this is duplicated from #396)

vdedios:
As commented in #426 

Closes #426
Closes #396


Damrod:
`DEFAULT_LOGIN_REDIRECT_URL` maybe we would like to change this one as well? Right now, when we land, we arrive still at the regular users (new discover friends page) page, and the user icon of the navbar is inactive. Please, take a look.

webapp/src/shared/urls.ts
```diff
@@ -17,7 +17,7 @@ export const CHATROOM_URL = '/chatroom';
 export const CREATE_CHATROOM_URL = `${CHATROOM_URL}/createchatroom`;
 export const EDIT_CHATROOM_AVATAR_URL = `${CHATROOM_URL}/:chatroomId/edit/avatar`;
 export const COMPONENTS_BOOK_URL = '/components-book';
-export const DEFAULT_LOGIN_REDIRECT_URL = `${USERS_URL}`;
+export const DEFAULT_LOGIN_REDIRECT_URL = `${FRIENDS_URL}`;
```
vdedios, line 20: @Damrod you mean this url? I think is already changed. Or maybe you meant another thing?

Damrod: Yes, you have changed that one. I don't know why, but after login in I'm still redirected to the old users page.

vdedios: Totally right, we were using the user_page url instead of the default_login_redirect one. Now should be working. Thanks!

Damrod:
I think the new default redirect page should have its friend's navbar item colored, the way it is in the discover friends page. Screenshot from 2022-12-31 18-54-25
Screenshot from 2022-12-31 18-54-35

Damrod (approved):
besides the comment, LGTM!

webapp/src/shared/urls.ts
Damrod: Thanks!

vdedios:
> I think the new "find new friends" page (former users page) should have its friend's navbar item colored, the way it is in the friends page.

Fixed, thanks again

Issue #415, opened 2022-12-24
Modify login to forbid login for banned users


Damrod:
An owner/admin should be able to:

1. Destroy chatrooms
2. **Ban users from the website**
3. See all chat channels without joining
4. Edit chatroom members

This issue tackles the second point.

PR #428, opened 2023-01-03
#415/banned users shouldn't be able to use any endpoint

Damrod:
as explaine in #415

### DIFF
--- a/transcendence-app/src/authorization/authorization.service.ts
+++ b/transcendence-app/src/authorization/authorization.service.ts
@@ -103,18 +103,22 @@ export class AuthorizationService {
   async getUserAuthContextForChatroom(
     userId: string,
     chatId: string,
-  ): Promise<ChatroomMemberWithAuthorization> {
-    const g_user = await this.getUserWithAuthorizationFromId(userId);
-    const crm = await this.chatroomMemberRepository.getById(chatId, userId);
-    return new ChatroomMemberWithAuthorization({
-      ...g_user,
-      chatId,
-      crm_member: !!crm,
-      crm_owner: crm?.owner,
-      crm_admin: crm?.admin,
-      crm_banned: crm?.banned,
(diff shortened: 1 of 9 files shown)


transcendence-app/src/chat/chat.gateway.ts
```diff
@@ -132,6 +132,9 @@ export class ChatGateway {
       client.request.user.id,
       chatroomId,
     );
+    if (!crm) {
```
Damrod, line 135: before, there was some code path where we could be throwing an http exception here. Now it should happen, even conceptually, anymore. I'm not sure what would happen if we tried http exceptions in ws tbh.

Damrod:
I think this part about the backend of the ban user feature is mostly ready. We could try and implement a ws event that logs out some user in real time after any admin bans them. I'm looking into it, but it's mostly a plus.

transcendence-app/src/authorization/authorization.service.ts
```diff
@@ -103,18 +103,22 @@ export class AuthorizationService {
...
+        ...g_user,
+        chatId,
+        crm_member: !!crm,
+        crm_owner: crm?.owner,
+        crm_admin: crm?.admin,
+        crm_banned: crm?.banned,
+        cr_muted: crm?.muted,
+      });
+    } catch (e) {
+      return null;
```
vdedios, line 120: what about logging this error?

vdedios:
> I think this part about the backend of the ban user feature is mostly ready. We could try and implement a ws event that logs out some user in real time after any admin bans them. I'm looking into it, but it's mostly a plus.

Although if would be nice, if the user already cannot navigate else where once is banned, I think we are at a very good point right now with this 👍

transcendence-app/src/authorization/authorization.service.ts
Damrod: done

PR #430, opened 2023-01-03
#429 / decouple game state

### LINKED ISSUE #429: Decouple game state from main game loop
Right now we use refs to keep the position of the ball and the user paddle. The idea of this task is to move that state into a global game state. 

With this, when we move the game logic to the server, it will be transparent for the client where is getting the state from (client calculation or server) because the model of the information would be the same and decoupled.

vdedios:
Closes #429 

⭐ Bonus: scaffolding of game module

This PR could be tricky (specially the thing about the state), if you need further explanation I can record a short video. Don't hesitate to ask


transcendence-app/src/game/game.gateway.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Socket, Server } from 'socket.io';
+import { BadRequestTransformationFilter } from '../shared/filters/bad-request-transformation.filter';
+import { WsAuthenticatedGuard } from '../shared/guards/ws-authenticated.guard';
+import { GameInputDto } from './dto/game-input.dto';
+
+@WebSocketGateway({ path: '/api/v1/socket.io' })
+@UseGuards(WsAuthenticatedGuard)
+@UseInterceptors(ClassSerializerInterceptor)
+@UseFilters(BadRequestTransformationFilter)
+export class GameGateway {
```
vdedios, line 25: Don't pay much attention to this new Game module, is just a dummy scaffolding I made to start thinking how is going to be the  ws connection between client and server

webapp/src/pages/PlayPage/PlayPage.tsx
```diff
@@ -1,17 +1,56 @@
...
   return (
-    <div className="play-page">
-      <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
-        Hit the brick!
-      </Header>
-      <Game />
-    </div>
+    <SearchContextProvider fetchFn={mockFetchFn} maxEntries={ENTRIES_LIMIT}>
+      <MainTabTemplate dataMapper={mockedMapper} buttons={buttons} />
+    </SearchContextProvider>
```
vdedios, line 54: I've created this page where we will list all live games so it's easier to start working in parallel around the game mvp

vdedios: The idea is to keep the GameTraining component and evolve the Game component into what should be the real game (one player against the other)

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -1,72 +1,73 @@
...
+  useClientGameEngine,
+} from './hooks';
+import { GameStateContextProvider } from './context';
 
 import './Game.css';
 
-export default function Game() {
+const GAME_SERVER_MESSAGE = 'gameServerMessage';
+
+const Game = () => {
```
vdedios, line 16: This component is an exact copy of the `GameTraining` below one. It only have scaffolding ws connection to start working on it

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -1,72 +1,73 @@
...
+  const [score, setScore] = React.useState<number>(0);
 
-  const { drawBall, drawPaddle, drawBrick, resetGame } = useGameAnimation(
-    ballRef,
-    paddleRef,
-  );
-  useGameControls(paddleRef);
+  const gameLoop = React.useCallback(() => {
+    const canvasContext = canvasRef.current?.getContext('2d');
+    const gameState = runGameFrame(deltaTimeRef.current);
```
vdedios, line 26: This is one of the magics, we just a game frame and get the current state which is currently decoupled from the game. This way, we can get this state from the server or the client, indepently (although it will need further reconciliation as we talked yesterday)

webapp/src/shared/components/Game/state.tsx
```diff
@@ -0,0 +1,142 @@
...
+        ...state,
+        paddle: {
+          ...paddle,
+          x: getPaddleDragX(paddle, dragPrevPos, dragCurrPos),
+        },
+      };
+    default:
+      return state;
+  }
+};
```
vdedios, line 142: ✨  This is the other magic, with this reducer we gather all the actions which can modify the state. This is also benefitial because those will be the command inputs we'll send to the server to run the game

webapp/src/shared/components/Game/hooks/useGameAnimation.tsx
```diff
@@ -5,22 +5,15 @@ import {
...
+const useGameAnimation = () => {
   const deltaTimeRef = React.useRef<number>(0);
 
   const drawBall = React.useCallback(
-    (context: CanvasRenderingContext2D) => {
-      const ball = ballRef.current;
-      const paddle = paddleRef.current;
-
-      ballRef.current = getBallPos(ball, paddle, deltaTimeRef.current);
+    (context: CanvasRenderingContext2D, ball: GameBall, paddle: GamePaddle) => {
```
Damrod, line 16: `paddle: GamePaddle` is unused here and now. Why do you think the callback to draw the Ball needs the paddle? Is it also scaffolding or just a mistake?

webapp/src/shared/components/Game/state.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const coef = Math.ceil(3 * Math.random());
+  const angle = (Math.PI / 2) * coef + Math.PI / 4;
+
+  return {
+    x: BALL_SPEED * Math.cos(angle),
+    y: BALL_SPEED * Math.sin(angle),
+  };
+};
+
+export const initialBallState = (): GameBall => {
```
Damrod, line 57: Where should this eventually go? Here or physics? Now its more duplicated...

Damrod:
To be honest I feel like this whole thing should be in a feature branch.

Right now, functionally, nothing appears to be broken (although I have, again, detected some very weid behaviour that I haven't managed to see twice or to record, and that I'm not sure is not a regression or was a bug that was there already), and I can see the intent, although I get somehow lost in the specifics, and there is, as you well know, a lot of duplicated code that is supposed to be scaffolding on which to build subsequent features.

I'm not sure this kind of empty scaffolding belongs in the main branch.

We have code reviews in order to merge whatever we think adds features and doesn't have regressions. I cannot attest now that there aren't.

Damrod:
That said, I really like where it appears like your feature is going, your workflow, and how are you structuring the code. 

But this thing may break a lot of things, and once it's in main, it will be hard to find where did things break. I don't want to siglehandedly accept the responsibility of this merge. But if you insist and there's nobody else around, I'll still approve, not to block you.

transcendence-app/src/game/game.gateway.ts
```diff
@@ -0,0 +1,36 @@
+import {
```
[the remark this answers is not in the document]
vdedios: Deleted!

webapp/src/shared/components/Game/hooks/useGameAnimation.tsx
vdedios: It was a leftover, removed!

webapp/src/shared/components/Game/state.tsx
vdedios: This should go in the state, it was a leftover, deleted!

vdedios:
> I cannot attest now that there aren't regressions, as I have, again, detected some very weid behaviour that I haven't managed to see twice or to record, and that I'm not sure is not a regression or was a bug that was there already:
> I have started a game in freshly new running instance of our app, and I have seen the ball bouncing around at what I estimate to be ten times the speed it normally has (without having modified the constants at all, btw). Upon reloading, this wasn't happening anymore, but it did happen once. I assume it's due, again, due to some strange effect on the calculation of the frame rate. But I don't know.

I think is not a bug but something related with create-react-app which keeps old state when hot reloading and this conflicts with the new one. Maybe you started the app from another branch and, when switching to this one, noticed the issue?

> But this PR may breaks a lot of things, and once it's in main, if something is broken, it will be hard to find where did things break. I don't want to siglehandedly accept the responsibility of this merge: something obnoxious may be hidden in the heavy incomplete boilerplate.

If this set of PRs breaks something it would be only inside the game feature which is currently under construction, so I'd say is not risky. Hontesly I prefer to integrate these PRs, if you don't mind, to avoid merging a future huge PR and having to fix tons of conflicts.

Also, since this is a big feature, the idea of splitting in several PRs is you can easy follow the changes.

Damrod (approved):
LGTM! Thank you!

PR #432, opened 2023-01-05
#415/admins should update other's roles through websockets

Damrod:
Working model of admin communicating different role assignments throu…gh websockets (to some extent)

Signed-off-by: Alvaro Ollero <aollero@gmail.com>

### DIFF
--- a/transcendence-app/src/socket/socket.gateway.ts
+++ b/transcendence-app/src/socket/socket.gateway.ts
@@ -5,17 +5,28 @@ import {
 } from '@nestjs/common';
 import {
   ConnectedSocket,
+  MessageBody,
   OnGatewayConnection,
   OnGatewayDisconnect,
   OnGatewayInit,
   SubscribeMessage,
   WebSocketGateway,
   WebSocketServer,
+  WsException,
 } from '@nestjs/websockets';
 import { Socket, Server, RemoteSocket } from 'socket.io';
 import { DefaultEventsMap } from 'socket.io/dist/typed-events';
 import { WsAuthenticatedGuard } from '../shared/guards/ws-authenticated.guard';
 import { SocketService } from './socket.service';
+import { AuthorizationService } from '../authorization/authorization.service';
+import { User } from '../user/infrastructure/db/user.entity';
+import { UserWithAuthorizationResponseDto } from '../authorization/dto/user-with-authorization.response.dto';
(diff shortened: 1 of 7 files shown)


webapp/src/shared/context/auth-context.tsx
```diff
@@ -80,6 +98,44 @@ export const AuthProvider = ({ children }: { children: ReactNode }) => {
     [isLoading, authUser, logout],
   );
 
+  useEffect(() => {
```
Damrod, line 101: This is mainly why do we need this PR. If we are admins, we may enable/disable endpoints by assigning roles to a user, but we need a way to update in real time other user's authUser when we do so. This is my solution.

webapp/src/pages/AdminPage/AdminPage.tsx
```diff
@@ -0,0 +1,106 @@
+import {
```
Damrod, line 1: This page is a draft to facilitate testing. The final shape should be nothing like this.

webapp/src/pages/AdminPage/AdminPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+    [userToRoleValues],
+  );
+  const handleOnSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    if (userToRoleValues) {
+      try {
+        const userToModify = await getUserId(userToRoleValues.username);
+        socket.emit('toggleUserWithRoles', {
+          id: userToModify.id,
+          role: UserToRoleDtoRoleEnum.Owner,
```
Damrod, line 44: To test the assignment of different roles, i'm hardcoding different values here in L44.

transcendence-app/src/socket/socket.gateway.ts
```diff
@@ -77,4 +83,19 @@ export class SocketGateway
   handleGetFriends(@ConnectedSocket() client: Socket) {
     this.socketService.getFriends(client);
   }
+
+  @SubscribeMessage('toggleUserWithRoles')
```
Damrod, line 87: Since I'm implementing this via ws, we don't really need the http endpoint to ass and remove usertorole anymore, but if I remove them, I also lose in the frontend the UserToRoleDto type, which I need. My solution has been to just leave those endpoints, as test endpoints, that enable swagger to document the UserToRoleDto class, and openapi generator create the correspoding type in webapp.
If someone has a different solution, I'd be interested in knowing that.

Issue #434, opened 2023-01-09
fix banned owner authorization case


Damrod:
As it is in current master, and admin can ban an owner, and then the owner, although it will be allowed to use most endpoints, and although it will not be logged out inmediately, will not be able to log back in after login out. We can fix this with two possible parallel approaches:
- prohibit to ban an owner
- allow an owner to login (local and 42oauth) even if it's "banned".

PR #435, opened 2023-01-09
#434/fix owner login to be admitted even if banned

Damrod:
As explained in #434


gbudau (approved):

Issue #443, opened 2023-01-18
⚛️ Create common request error handler


Damrod:
We have this pattern very often:
```
  if (error instanceof ResponseError) {
    error.response.json().then((responseBody) => {
      if (responseBody.message) {
        errMessage = responseBody.message;
      } else {
        errMessage = error.response.statusText;
      }
    });
  } else if (error instanceof Error) {
    errMessage = error.message;
  }
  logger(errMessage);
```
we can DRY it

Issue #377, opened 2022-11-29
Change logs console.error with warn(error)


fizcamposancos:
Maybe now that we have the notifications, we could do a warn(e) instead of console.error(e) in the catch block of the returned promises.

_Originally posted by @Damrod in https://github.com/42AGV/ft_transcendence/pull/374#discussion_r1034516091_

PR #444, opened 2023-01-18
#443/make common error handler function

Damrod:
as explained in #443 and #377

### DIFF
--- a/webapp/src/pages/UserPage/UserPage.tsx
+++ b/webapp/src/pages/UserPage/UserPage.tsx
@@ -25,7 +25,7 @@ import {
 import { useNotificationContext } from '../../shared/context/NotificationContext';
 import socket from '../../shared/socket';
 import { WsException } from '../../shared/types';
-import { handleRequestError } from '../../shared/hooks/HandleRequestError';
+import { handleRequestError } from '../../shared/utils';
 
 export default function UserPage() {
   const { warn } = useNotificationContext();
--- a/webapp/src/pages/UserPage/UserPage.tsx
+++ b/webapp/src/pages/UserPage/UserPage.tsx
@@ -25,6 +25,7 @@ import {
 import { useNotificationContext } from '../../shared/context/NotificationContext';
 import socket from '../../shared/socket';
 import { WsException } from '../../shared/types';
+import { handleRequestError } from '../../shared/hooks/HandleRequestError';
 
 export default function UserPage() {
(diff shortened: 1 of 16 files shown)


webapp/src/shared/hooks/HandleRequestError.tsx
```diff
@@ -0,0 +1,21 @@
+import { ResponseError } from '../generated';
+
+export const handleRequestError = (
+  error: unknown,
+  msg: string,
+  logger: (arg0: any) => void = console.error,
+) => {
```
gbudau, line 7: question: I see we place the file in the hooks directory, do we plan to change it into a custom hook in the future?

gbudau (approved):

webapp/src/shared/hooks/HandleRequestError.tsx
gbudau: We could keep it as a function because we don't use any state or effects. What do you think about moving the file into another directory (maybe `/shared/utils` ?) and changing the extension to `.ts`?

fizcamposancos (approved):
I like how the error handler looks like. Nice work!

webapp/src/shared/hooks/HandleRequestError.tsx
Damrod: Now it should be in the right place

webapp/src/shared/utils/HandleRequestError.ts
```diff
@@ -0,0 +1,21 @@
+import { ResponseError } from '../generated';
+
+export const handleRequestError = (
+  error: unknown,
+  msg: string,
+  logger: (arg0: any) => void = console.error,
```
vdedios, line 6: Nice!

webapp/src/pages/UserPage/UserPage.tsx
```diff
@@ -74,7 +75,7 @@ export default function UserPage() {
           await usersApi.userControllerUnfollowUser({ userId: user.id });
         }
       } catch (error) {
-        console.error(error);
+        handleRequestError(error, 'Could not toggle friends', warn);
```
vdedios, line 78: Pitminí:

```suggestion
        handleRequestError(error, 'Could not toggle followship', warn);
```

vdedios (approved):
Nice work, LGTM! DRY! 🦾

PR #445, opened 2023-01-19
#439/Implement 2FA with TOTP frontend

### LINKED ISSUE #165: Implement 2FA with TOTP
- [x] #438 
- [x] #439

### LINKED ISSUE #439: Implement 2FA with TOTP frontend
Subtask of #165

gbudau:
close #439
close #165

### DIFF
--- a/webapp/src/pages/AuthUserPage/AuthUserPage.tsx
+++ b/webapp/src/pages/AuthUserPage/AuthUserPage.tsx
@@ -1,5 +1,6 @@
 import {
   AvatarPageTemplate,
+  Button,
   ButtonVariant,
   IconVariant,
   Row,
@@ -10,10 +11,49 @@ import {
 } from '../../shared/components';
 import './AuthUserPage.css';
 import { useAuth } from '../../shared/hooks/UseAuth';
-import { AVATAR_EP_URL, EDIT_USER_URL } from '../../shared/urls';
+import {
+  AVATAR_EP_URL,
+  EDIT_USER_URL,
+  TWO_FACTOR_AUTH_ENABLE_URL,
+} from '../../shared/urls';
+import { authApi } from '../../shared/services/ApiService';
+import { useNavigate } from 'react-router-dom';
+import { useNotificationContext } from '../../shared/context/NotificationContext';
 
 export default function AuthUserPage() {
-  const { authUser, logout, isLoading: isAuthUserLoading } = useAuth();
+  const {
+    authUser,
+    setAuthUser,
+    logout,
+    isLoading: isAuthUserLoading,
+  } = useAuth();
+  const navigate = useNavigate();
+  const enableTwoFactorAuthButton = (
+    <Button
+      variant={ButtonVariant.SUBMIT}
+      onClick={() => navigate(TWO_FACTOR_AUTH_ENABLE_URL)}
+    >
+      Enable 2FA
+    </Button>
+  );
+  const { warn } = useNotificationContext();
+  const disableTwoFactorHandler = async () => {
+    if (!authUser) {
+      return;
+    }
+    try {
+      await authApi.authControllerDisableTwoFactorAuthentication();
+      setAuthUser({ ...authUser, isTwoFactorAuthenticationEnabled: false });
+    } catch (error) {
+      warn('Service Unavailable');
+    }
+  };
+  const disableTwoFactorAuthButton = (
+    <Button variant={ButtonVariant.WARNING} onClick={disableTwoFactorHandler}>
+      Disable 2FA
+    </Button>
+  );
+
   return (
     <div className="auth-user-page">
       <AvatarPageTemplate
@@ -54,6 +94,9 @@ export default function AuthUserPage() {
             url={EDIT_USER_URL}
             title="Edit profile"
           />
+          {authUser?.isTwoFactorAuthenticationEnabled
+            ? disableTwoFactorAuthButton
+            : enableTwoFactorAuthButton}
         </>
       </AvatarPageTemplate>
     </div>
--- a/webapp/src/pages/AuthUserPage/AuthUserPage.tsx
+++ b/webapp/src/pages/AuthUserPage/AuthUserPage.tsx
@@ -1,6 +1,6 @@
 import {
   AvatarPageTemplate,
-  Button,
+  ButtonProps,
   ButtonVariant,
   IconVariant,
   Row,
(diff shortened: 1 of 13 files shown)


webapp/src/pages/AuthUserPage/AuthUserPage.tsx
```diff
@@ -54,6 +94,9 @@ export default function AuthUserPage() {
             url={EDIT_USER_URL}
             title="Edit profile"
           />
+          {authUser?.isTwoFactorAuthenticationEnabled
```
Damrod, line 97: What do you think about placing this button in the secondaryButton prop of the AvatarPage template instead?

Damrod:
There is some bug, I don't where is it coming from:
When I first enable 2fa with user `a`, it works, and it makes a new 2fa entry in my 2fa app in my cellphone: `ft_transcendece (a@a.com)`.
Then I remove this, I log out, and I try to enable another 2fa with a different user. This time around, the 2fa entry the 2fa app in my cellphone, the name is still `ft_transcendence (a@a.com)`, and no matter how many times I try to input a code with the fresh, newly registered entry, the code is wrong.

???

webapp/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -38,10 +39,14 @@ export default function LoginPage() {
       const authUser = await authApi.authControllerLoginLocalUser({
         loginUserDto: { username, password },
       });
-      const { gOwner, gAdmin, gBanned } =
-        await authApi.authControllerRetrieveAuthUserWithRoles();
-      setAuthUser({ ...authUser, gOwner, gAdmin, gBanned });
-      navigate(DEFAULT_LOGIN_REDIRECT_URL, { replace: true });
+      if (authUser.isTwoFactorAuthenticationEnabled) {
+        navigate(TWO_FACTOR_AUTH_VALIDATE_URL, { replace: true });
```
Damrod, line 43: what happens if an admin is 2fa enabled? Will they not have their authUser set to have admin rights? (I hit the bug I've reported while investigating this)

webapp/src/pages/EnableTwoFactorAuthPage/EnableTwoFactorAuthPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <div className="enable-two-factor-auth-page">
+      <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
+        enable 2fa
+      </Header>
+      <div className="enable-two-factor-auth-page-main">
+        <img
+          className="enable-two-factor-auth-page-qrcode"
+          src={TWO_FACTOR_AUTH_QR_EP_URL}
+          alt="QR Code"
+        />
```
vdedios, line 75: I have a similar issue related to what @Damrod explained. In my case, when disabling and enabling again 2FA I got the same QR as before (when a new one was expected) and I could enable 2FA.

After investigating a bit I think the proble is here, the QR image is being cached by the browser because you are requesting it as an asset. So, when you leave the page and re-enter (after a new 2FA activation, even from another user) the browser returns the old cached image which is tied to the previous session.

vdedios: One solution to fix this could be to attach a query string to the url with the current utc time. This way we would have a different url each time the user requests the QR and the browser would be forced to request the resource again (https://stackoverflow.com/questions/126772/how-to-force-a-web-browser-not-to-cache-images)

The downside for this option is the browser would be caching assets that the user is not going to consume again. But since the user is not going to visit this page frequently, not a big deal.

webapp/src/pages/EnableTwoFactorAuthPage/EnableTwoFactorAuthPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+    }
+    try {
+      await authApi.authControllerEnableTwoFactorAuthentication({
+        twoFactorAuthenticationCodeDto: { code },
+      });
+      setAuthUser({ ...authUser, isTwoFactorAuthenticationEnabled: true });
+      notify('You enabled 2FA');
+      navigate(USER_ME_URL, { replace: true });
+    } catch (error: unknown) {
+      if (error instanceof ResponseError) {
```
Damrod, line 42: if we stack the changes from #444 on top of this one, I'd like to remove this block and the same one from `webapp/src/pages/ValidateTwoFactorAuthPage/ValidateTwoFactorAuthPage.tsx`
Otherwise, if you stack this one on top of that one, please, use the shared function we have there.

webapp/src/pages/EnableTwoFactorAuthPage/EnableTwoFactorAuthPage.tsx
vdedios: But I see that a more elegant solution would be to directly append extra headers to the controller response to indicate the browser we don't want to cache the resource, such as `@Header('Cache-Control', 'no-cache')`

gbudau: Thank you, guys! I added the query string because the header didn't work for me.

vdedios: Thanks! I am investigating this because it neither worked for me. I think it's due to client side navigation. I'll write my conclusions in this thread 😄

webapp/src/pages/AuthUserPage/AuthUserPage.tsx
gbudau: Thanks for the suggestion! Done.

webapp/src/pages/LoginPage/LoginPage.tsx
gbudau: I think we deal with that case in the `ValidateTwoFactorAuthPage` where we set the `authUser` and the permissions.

webapp/src/pages/EnableTwoFactorAuthPage/EnableTwoFactorAuthPage.tsx
Damrod: The current fix is still not working for me, guys. Images below. Different query string, same image:
Screenshot from 2023-01-19 13-05-27
Screenshot from 2023-01-19 13-05-16

vdedios: steps to reproduce?

Damrod: 1 - login with user `a`
2 - click enable 2fa
2bis - open qr code in a different tab
3 - scan qr code from original tab with cellphone app
4 - log out
5 - log in with 'a', check that 2fa works as expected
6 - log out from 'a'
7 - log in with 'b'
8 - try to enable 2fa clicking the button
9 - see that prompted qr code is the same than the one in the side tab i opened in point 2bis

Damrod: I've rebuilt the app and the volumes and now it works.

??

Damrod (approved):
LGTM!
Thank you!

webapp/src/pages/EnableTwoFactorAuthPage/EnableTwoFactorAuthPage.tsx
gbudau: Sure, if you prefer, we could merge #444 before this one (I will try to review it this afternoon) or vice versa.

Damrod: Merge this one. Mine is stacked on top of the admin page one, and thats big, so I assume it's gonna take quite longer, where instead this one is mostly ready.

PR #447, opened 2023-01-20
#446 / move pong engine to library

### LINKED ISSUE #446: Move game logic into a shared library
This is needed since we'll run the same pong engine in the backend to calculate the real game state and in the webapp to simulate and interpolate the game

vdedios:
closes #446 

❗ This may break what you have locally. Follow these steps to run from this branch:
1. `git reset --hard && git clean -dxf`
2. generate a new `.env`
3. run `make` as usual
4. run `npm i` and `npm run build-packages` **from the root of the project** to install the dependencies locally (no need to run `npm i` inside each folder as before because this is now a monorepo, explained below)
5. Maybe reload editor in case any dependencies still marked as missing.

Changes:
- Move all keyboard and drag logic into the state reducer
- Move all the pong engine files to a shared library
- For that library to work I've set up a monorepo. I suggest you to watch this [intro video](https://www.youtube.com/watch?v=KEkRy4q_0oI&t=84s) to get to know how it works and read npm workspaces [documentation](https://docs.npmjs.com/cli/v7/using-npm/workspaces).
- Finally, since the npm setup has changed a bit, fix al docker and build related things


transcendence-app/package.json
```diff
@@ -68,9 +68,7 @@
     "@types/cookie-parser": "^1.4.3",
     "@types/express": "^4.17.14",
     "@types/express-session": "^1.17.5",
-    "@types/jest": "29.2.4",
     "@types/multer": "^1.4.7",
```
Damrod, line 71: Al of these changes are made in order to eventually have `"pong-engine": "^1.0.0",` this here, isn't it? That you will place in a follow up PR?

vdedios: Yes! It's being used in the webapp indeed -> https://github.com/42AGV/ft_transcendence/blob/7d59b0e78c136476b7a8c6f2f1bde5d1247b08cf/webapp/package.json#L7

and in the follow up PR this will be also imported in the backend

transcendence-app/tsconfig.json
```diff
@@ -15,6 +15,6 @@
     "forceConsistentCasingInFileNames": false,
     "noFallthroughCasesInSwitch": false,
     "strict": true,
-    "typeRoots": ["./node_modules/@types", "./types"]
+    "typeRoots": ["./node_modules/@types", "../node_modules/@types", "./types"]
```
Damrod, line 18: do we still need this one `"./node_modules/@types"` ?

transcendence-app/Dockerfile
```diff
@@ -25,14 +26,15 @@ ENV NODE_ENV=production
 RUN apk add --no-cache dumb-init \
   && mkdir -p ${TRANSCENDENCE_APP_DATA} \
   && chown -R node:node ${TRANSCENDENCE_APP_DATA}
-WORKDIR  /transcendence-app
+WORKDIR  /service
```
Damrod, line 29: why did you call this folder (the monorepo root, I assume) service?

README.md
```diff
@@ -178,7 +202,7 @@ To create a migration, run the following command from the transcendence-app dire
 npx knex migrate:make migration_name
 ```
 
-### Seed the database
+### Seed the database <a name = "start-seeds"></a>
```
Damrod, line 205: 🌱

README.md
```diff
@@ -162,7 +186,7 @@ make gen
 
 `make gen` creates both yaml file and generated webapp files. To generate only webapp generated files run `make gen-webapp`. To generate only swagger spec run `make spec`.
 
-### Migrate the database schema
+### Migrate the database schema <a name = "start-migrations"></a>
```
Damrod, line 189: 🧭

Damrod (approved):
I explored the monorepo concept a few months ago, and eventually gave up, but I'm glad we've found a use case that makes it worth our time. Thanks for your implementation! Looks great!

packages/pong-engine/src/state.ts
```diff
@@ -95,35 +84,23 @@ export const reducer = (
     case 'paddleMoveRight':
       return {
         ...state,
-        paddle: {
-          ...paddle,
-          slide: PADDLE_SLIDE_SPEED,
-        },
+        paddle: movePaddleRight(paddle),
```
Damrod, line 87:
```
  const { ball, paddle, score } = state;
```
instead of
```
  const ball = state.ball;
  const paddle = state.paddle;
  const score = state.score;
```
?

packages/pong-engine/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
Damrod, line 1: maybe we would like to add another `.prettierrc` in this folder with this content
```
{
  "singleQuote": true,
  "trailingComma": "all"
}
```

So our autoformatters don't mess everything up?
Maybe there's a way to place it in the monorepo root?

Damrod: Also, maybe it'd be convenient to have a GitHub job that checks if the packages folder is properly styled/linted/builds as well?

gbudau (approved):

transcendence-app/Dockerfile
vdedios: Because transcendence-app was converted into a subfolder of the main path and, since these docker images are meant to be doployed as services, I think that was a more less accurate denomination.

README.md
vdedios: Do you mean to add the emoji?

packages/pong-engine/src/state.ts
vdedios: Yes, thank you!

packages/pong-engine/package.json
vdedios: Yes good idea 👍 a benefit of monorepos is to share configuration also, I'll add a common prettier and eslint configuration.

Also I'll add an additional GI job for the packages, I also think is a good idea to test them in PRs

README.md
Damrod: Yes, but don't lose your time if you don't feel like it :D

vdedios: Not a problem, I'll add it!

packages/pong-engine/package.json
vdedios: Finally added:

- a common config for prettier. Not for eslint since webapp and transcendence-app have different configurations
- extract style checking to a independent GH action. Done like this because, apparently, there is no way to only install `devDependencies` at root level. This would be desirable so we wouldn't have to do the full `npm i` globally in each service action just to install prettier and run the style check. Maybe there is a better way to do it but don't want to waste more time on this :(
- create a new GH action for the packages

transcendence-app/tsconfig.json
vdedios: done 👍 

Just a warning, you may still find `node_modules` inside each workspace. Not 100% sure why this happens but it seems related with version mismatch of a same package between workspaces and root node modules. Despite having the same module with the same version defined in `package.json` there might be some transitive dependencies which don't match the package version -> https://stackoverflow.com/questions/71048697/npm-workspaces-some-modules-installed-to-local-node-modules-dir

.prettierignore
```diff
@@ -0,0 +1 @@
+./webapp/generated
```
Damrod, line 1: shouldn't this be `/webapp/src/shared/generated`?
I don't think it will find any folder name generated inside webapp folder, which is the intent, like this.

PR #460, opened 2023-01-26
#454/pretty confirm alert

### LINKED ISSUE #454: ⚛️ persistent notifications
In order to accept or decline game challenges, the challengee needs to have some kind of prompt that will pop when challenged, with some options. A possibility to achieve this is to upgrade the notification/snackbar component, in order to make it persist, and have some buttons to trigger actions/navigate to games.

Damrod:
As described in #454 
closes #454


webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.tsx
```diff
@@ -0,0 +1,54 @@
...
+import './CustomConfirmAlert.css';
+import { Button, ButtonProps, Text, TextVariant } from '../index';
+
+type CustomConfirmAlertProps = {
+  title: string;
+  message: string;
+  buttons: ButtonProps[];
+};
+
+export default function CustomConfirmAlert({
```
Damrod, line 11: This is technically not a component, since it's invoked like a function is called, doesn't return a ReactNode, but it's close enough, I think, and has UI and style, which we would expect from a regular component. Let me know if this bothers you!

webapp/src/pages/EditChatroomDetailsPage/EditChatroomDetailsPage.tsx
```diff
@@ -164,18 +166,36 @@ export default function EditChatroomDetailsPage() {
...
-        chatApi.chatControllerDeleteChatroom({
-          chatroomId: chatroomId!,
-        });
-        notify('Chatroom successfully deleted');
-        navigate(`${overridePermissions ? ADMIN_URL : ''}${CHATS_URL}`);
-      } catch (error) {
-        handleRequestError(error, "Couldn't delete chatroom", warn);
-      }
-    }
+    CustomConfirmAlert({
```
Damrod, line 171: I changed this in order to have an example of how does the component work.

webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.css
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.react-confirm-alert-overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  z-index: 99;
+  background: var(--color-background);
```
fizcamposancos, line 12: I think I like it more without the background, overlapping the current content instead of showing the alert over a black background

Damrod: What amount of background would we like to see? It's as simple as changing the transparency channel of the background attribute. I.e.:
```diff
Index: webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.css
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.css b/webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.css
--- a/webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.css	(revision 81007dfe565f77de2f21f7e4121d8cacba2718ab)
+++ b/webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.css	(date 1674736988733)
@@ -9,7 +9,7 @@
   right: 0;
   bottom: 0;
   z-index: 99;
-  background: var(--color-background);
+  background: rgb(0, 0, 0, 0.8);
   display: flex;
   justify-content: center;
   align-items: center;
```

gbudau (approved):
Thank you! Looks great :tada:

webapp/src/shared/components/ConfirmAlert/CustomConfirmAlert.tsx
```diff
@@ -0,0 +1,65 @@
...
+  buttons: Partial<ButtonProps>[];
+};
+
+export default function CustomConfirmAlert({
+  title,
+  message,
+  buttons,
+}: CustomConfirmAlertProps): void {
+  confirmAlert({
+    customUI: ({ onClose }) => {
```
fizcamposancos, line 24: I would add `closeOnClickOutside: true,` option to prevent from rejecting the invition to a game accidentally, but I tried and I wasn't able to make it work

Damrod:
```
    closeOnClickOutside: false,
```
works for me. Will push it.

PR #462, opened 2023-01-26
#453/timer component

### LINKED ISSUE #453: ⚛️ timer component
In order to give deadlines to player's game setup, and possibly set timeouts for games, we need a timer component. It should run both forwards and backwards

Damrod:
as detailed in #453 
closes #453

### DIFF
--- a/webapp/src/shared/components/Timer/Timer.tsx
+++ b/webapp/src/shared/components/Timer/Timer.tsx
@@ -25,7 +25,7 @@ const secondsToClockType = (seconds: number): clockType => {
   };
 };
 
-function parseTime(input: string): clockType {
+const parseTime = (input: string): clockType => {
   let a = input.split(':'); // split it at the colons
   if (a[0] && a[1] && a[2]) {
     return {
@@ -37,7 +37,7 @@ function parseTime(input: string): clockType {
     };
   }
   throw new Error('Bad time format');
-}
+};
 
 export default function Timer({
   textVariant = TextVariant.TITLE,
--- a/webapp/src/shared/components/Timer/Timer.tsx
+++ b/webapp/src/shared/components/Timer/Timer.tsx
@@ -3,6 +3,7 @@ import { Text, TextVariant } from '../index';
 import './Timer.css';
 
 type TimerProps = {
+  shouldRun?: boolean;
   textVariant?: TextVariant;
   timeString?: string;
   isBackwardsCount?: boolean;
@@ -40,6 +41,7 @@ const parseTime = (input: string): clockType => {
 };
 
 export default function Timer({
+  shouldRun = true,
   textVariant = TextVariant.TITLE,
   timeString = '00:01:00',
   isBackwardsCount = true,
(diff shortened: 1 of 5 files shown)


fizcamposancos (approved):
Well done, thanks for this component. Just a comment, the precision of the timer is rounded to seconds, I think that it would be enough for our needs, and we have the option set it to hours, that I don't think we will use it.

Damrod:
I'm considering that, when we do need this, it will be to coordinate times actions between users, which means that maybe the clock will need to run in the backend instead, as a single source of truth. I will not merge this for now. Maybe I'll take a look at this concept.

vdedios:
> I'm considering that, when we do need this, it will be to coordinate timed actions between users, which means that maybe the clock will need to run in the backend instead, as a single source of truth. I will not merge this for now. Maybe I'll take a look at this concept. [Example src](https://dev.to/rukshanjs/part-13-how-to-create-a-server-side-timer-using-websockets-with-socketio-nestjs-and-flutter-3821)

Yes, good point, Maybe you can merge this just as a dummy visual, without any logic and we'll handle time in other place

webapp/src/shared/components/Timer/Timer.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export default function Timer({
+  shouldRun = true,
+  textVariant = TextVariant.TITLE,
+  timeString = '00:01:00',
+  isBackwardsCount = true,
+  onTimeOut,
+}: TimerProps) {
+  const [startDate, _] = useState<clockType>(parseTime(timeString));
```
vdedios, line 53: Why are you using a React state if not changing the state? maybe a constant is better?

webapp/src/shared/components/Timer/Timer.tsx
```diff
@@ -0,0 +1,106 @@
...
+  void _;
```
vdedios, line 54: This is weird :/

webapp/src/shared/components/Timer/Timer.tsx
```diff
@@ -0,0 +1,106 @@
...
+        : {
+            timeInSeconds: 0,
+            hours: 0,
+            minutes: 0,
+            seconds: 0,
+          },
+    );
+  useEffect(() => {
+    if (shouldRun) {
+      setTimeout(() => {
```
gbudau, line 68: suggestion: We could add a `clearTimeout` here on the `useEffect` cleanup.

webapp/src/shared/components/Timer/Timer.tsx
```diff
@@ -0,0 +1,106 @@
...
+      seconds: parseInt(a[2]),
+    };
+  }
+  throw new Error('Bad time format');
+};
+
+export default function Timer({
+  shouldRun = true,
+  textVariant = TextVariant.TITLE,
+  timeString = '00:01:00',
```
vdedios, line 49: What do you think about better exposing a `Number` property , for example `timeInSeconds`? This way we could easily reuse this component as mentioned in other comment and avoid parsing

gbudau (approved):

vdedios (approved):

webapp/src/shared/components/Timer/Timer.tsx
Damrod: But why would we want to avoid parsing? The point is that a user can write down a time interval of its liking, and have a timer set up in place for the appropriate amount of time. I don't see a way around parsing. Maybe we can push the parsing up into some of the components/pages above this one, and only render this after the input has been validated and converted to seconds as a number, so that we don't have to manage a possible throw inside the render method of a JSX.Element. Since we don't have a page that uses this yet, I'll still leave it like this for the time being.

webapp/src/shared/components/Timer/Timer.tsx
Damrod: Added (only in the paths where we actually want to clear it, although I'm not sure we do need it there. If we add it in the other paths, then the timer doesn't run).

webapp/src/shared/components/Timer/Timer.tsx
Damrod: Yes, but then it will calculate this over and over every time it renders. I can settle for a useMemo, though. Done.

webapp/src/shared/components/Timer/Timer.tsx
Damrod: As commented above.

PR #470, opened 2023-01-31
[1] #431 / run game training in backend

### LINKED ISSUE #431: Run game training in backend
This subtask is about running all the game logic in the backend, reconciliating the information sent by the server with the client simulation and sending all that information through ws

vdedios:
closes #431


Damrod:
webapp is not compiling for me:
```
Compiled with problems:X

ERROR in src/shared/components/Game/context.tsx:3:21

TS2305: Module '"pong-engine"' has no exported member 'newGame'.
    1 | import * as React from 'react';
    2 |
  > 3 | import { GameState, newGame } from 'pong-engine';
      |                     ^^^^^^^
    4 |
    5 | type GameStateContextType = {
    6 |   gameStateRef: React.MutableRefObject<GameState>;


ERROR in src/shared/components/Game/hooks/useGameControls.tsx:7:3

TS2305: Module '"pong-engine"' has no exported member 'paddleMoveRight'.
     5 | import {
     6 |   GameCommand,
  >  7 |   paddleMoveRight,
       |   ^^^^^^^^^^^^^^^
     8 |   paddleMoveLeft,
     9 |   paddleStop,
    10 |   paddleDrag,
```

packages/pong-engine/src/state.ts
```diff
@@ -22,14 +22,16 @@ import {
...
+
 type Act<Type extends string, Payload extends object> = {
   type: Type;
   payload: Payload;
 };
 
 export type Action =
-  | Act<'move', { deltaTime: number }>
-  | Act<'lose' | 'win' | GamePaddleMoveCommand, object>
+  | Act<'move' | 'addPoint', { deltaTime: number }>
```
Damrod, line 33: I have to say that this `Act` type is veeery confusing. Why are losePoint and GamePaddleMoveCommand (as ) together? Why is  'move' and 'addPoint' together?
I assume it's basically grouping all types that have the same payload, but it doesn't make it readable.

packages/pong-engine/src/game.ts
```diff
@@ -0,0 +1,60 @@
+import { GameBall, GameState } from './models';
+import {
+  CANVAS_HEIGHT,
+  CANVAS_WIDTH,
+  BALL_RADIUS,
+  BRICK_WIDTH,
+} from './constants';
+import { reducer, initialBallState, initialPaddleState } from './state';
+
+const losePoint = (ball: GameBall) => ball.y > 1.5 * CANVAS_HEIGHT;
```
Damrod, line 10: I'm not reviewing the math, but I have to say that I don't get how the ball being at a y coordinate larger than 1.5 times the canvas makes it a lose point. It's kind of works so I'm inclined to believe it makes sense, but maybe it's dead code, and I have to voice my question

Damrod:
I'm seeing that the game doesn't flush anymore when navigating out of the game page. Is this intended behaviour?

Also, I'd say sometimes I'm hitting bullseye, and not scoring the point. But maybe it's just my faulty human senses.

packages/pong-engine/src/state.ts
```diff
@@ -62,21 +64,22 @@ export const reducer = (
 
   switch (type) {
     case 'move':
-      const { deltaTime } = payload;
       return {
         ...state,
-        ball: getBallPos(ball, paddle, deltaTime),
-        paddle: getPaddlePos(paddle, deltaTime),
+        ball: getBallPos(ball, paddle, payload.deltaTime),
```
Damrod, line 69: why don't we like destructuring the object here but we like it in L101?

transcendence-app/src/game/validators/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const GAME_COMMANDS = [
+  'paddleMoveRight',
+  'paddleMoveLeft',
+  'paddleStop',
+  'paddleDrag',
+];
+
+// https://stackoverflow.com/questions/51528780/typescript-check-typeof-against-custom-type
+type GameCommandType = (typeof GAME_COMMANDS)[number];
```
Damrod, line 11: interesting read!

webapp/src/shared/components/Game/hooks/useOnlineGame.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from 'react';
+import socket from '../../../socket';
+import { GameCommand, GameState } from 'pong-engine';
+import { WsException } from '../../../types';
+import { useNotificationContext } from '../../../context/NotificationContext';
+
+// Tipar mensajes ws -> investigar si se puede hacer con OpenApi
```
Damrod, line 7: I don't think OpenApi does this.
But we can always have some more types added to a shared package (pong-engine in this case).

transcendence-app/src/game/validators/index.ts
```diff
@@ -0,0 +1,32 @@
+import { registerDecorator } from 'class-validator';
+
+const GAME_COMMANDS = [
+  'paddleMoveRight',
```
Damrod, line 4: It appears to me as if this `GAME_COMMANDS` should be a shared object across fe and be.

vdedios:
> PS. The second time I tried make clean and make it worked. ???

This is expected. Since we don’t have hot reload in the library and we’ve introduced new changes, there’s no other way than rebuilding it.

Damrod (approved):
Besides the comments, kind of works so I won't stop you!

packages/pong-engine/src/state.ts
vdedios: Yes, they are together because they share the same type of payload. We can take a look to split them a bit

packages/pong-engine/src/game.ts
vdedios: This is a visual “trick” so we have a goal effect. This is, game does not reset immediately after scoring a point and we give a few milliseconds to the player to realize the ball went out of the court

transcendence-app/src/game/validators/index.ts
vdedios: Thanks!

webapp/src/shared/components/Game/hooks/useOnlineGame.tsx
vdedios: Yes, @gbudau told me about a lib we could use but we’re not 100% sure if it worth

transcendence-app/src/game/validators/index.ts
vdedios: As commented above, we’ll think about it 👍

packages/pong-engine/src/state.ts
vdedios: Because we are also using deltaTime inside other case condition and, since the scope of the switch statement is the same for all case conditions, we cannot declare twice the variable

vdedios:
> I'm seeing that the game doesn't flush anymore when navigating out of the game page. Is this intended behaviour?

Do you mean, the result does not reset to zero?

> Also, I'd say sometimes I'm hitting bullseye, and not scoring the point. But maybe it's just my faulty human senses.

since the game is running in backend and we don’t have any reconciliation mechanism yet, that’a totally possible 🤪

Damrod:
>  Do you mean, the result does not reset to zero?

Yes

vdedios:
> Do you mean, the result does not reset to zero?

About this, I've been reviewing it and couldn't get to reproduce it. That being said, since this "online" version is going to be replace in next PRs with the multiplayer one I think it does not worth to spend time investigating the issue

PR #471, opened 2023-01-31
[3] #469/run multiplayer game in backend

### LINKED ISSUE #469: Run multiplayer game in backend
Subtask of #425

gbudau:
close #469 

Working:

- Users can play with other users or spectate ongoing games (alpha version with authoritative server)

- When a player disconnects, the game pauses (max 30 seconds)

- Update pairing context with the playing status

TODO (in other PRs):

- When the user reconnects, if they have an ongoing game, either redirect them to the game page or show a confirmation dialogue to join or quit the game.

- The game ends when a player reaches an arbitrary number of points

- Enable paddle drag

- When a player disconnects or pauses the game, show a notification

- Refactor the `GameService`, and create in-memory repositories for the data.


Damrod:
As far as I can see, the changes from [this](https://github.com/42AGV/ft_transcendence/pull/484) PR are still relevant. When we receive 'socket.userDisconnect' from the queue, we neet to update the waiting status in the server, but we also still need to notify the user that has not left so they can also stop waiting.

transcendence-app/src/game/game.queue.service.ts
```diff
@@ -29,13 +28,13 @@ export class GameQueueService {
   private gameQueue: QueuedUser | null = null;
 
   constructor(
-    private gamesOngoing: IGamesOngoingRepository,
```
Damrod, line 32: didn't we say that this could be readonly?

transcendence-app/src/game/game.service.ts
```diff
@@ -0,0 +1,343 @@
...
+    const isPlayerTwo = userId === game.playerTwoId;
+
+    if (!(isPlayerOne || isPlayerTwo)) {
+      return;
+    }
+    const playerClients = this.playerClients.get(userId);
+    if (!playerClients) {
+      this.playerClients.set(userId, new Set<ClientId>([client.id]));
+    } else {
+      playerClients.add(client.id);
```
Damrod, line 121: should this be
```
      this.playerClients.get(userId)?.add(client.id);
```
Or is `playerClients` a reference to `this.playerClients.get(userId)` and any change on the first affects the seconds? I get confused with these things in javascript...

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -1,78 +1,157 @@
...
+    (command: GameCommand) => {
+      if (isPlayerRef.current && gameJoined) {
+        socket.emit(GAME_COMMAND, {
+          command,
+          gameRoomId: gameId,
+        });
+      }
+    },
+    [gameId, gameJoined],
+  );
```
vdedios, line 59: Why did you delete the online hook and move the socket handling here? I think we are mixing things up but I'm sure you have a good reason :D

gbudau:
> As far as I can see, the changes from [this](https://github.com/42AGV/ft_transcendence/pull/484) PR are still relevant. When we receive 'socket.userDisconnect' from the queue, we neet to update the waiting status in the server, but we also still need to notify the user that has not left so they can also stop waiting.

Updated. Thanks for the suggestions!

transcendence-app/src/game/game.queue.service.ts
gbudau: We moved the ongoing games to the game service, and now the queue service uses that one to check if a player is in a game.

The ongoing games are read-only for the queue service meaning that it only reads the game status (get) and doesn't write (create, update, delete)

transcendence-app/src/game/game.service.ts
Damrod: Checked, this seems to work. Sorry for my ignorance.

gbudau: Sorry, this was a bit confusing. I renamed `this.playerClients` to `this.playerToClients`.

And yes, [`Map.prototype.get()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get) returns a reference, in this case, a reference to the player clients set or `undefined`.

transcendence-app/src/game/game.queue.service.ts
```diff
@@ -29,13 +30,13 @@ export class GameQueueService {
...
+    private gameService: GameService,
```
Damrod, line 33:
```
    private readonly gameService: GameService,
```

This is what i mean

gbudau: Oh! Done :+1:

webapp/src/shared/context/GamePairingContext.tsx
```diff
@@ -71,11 +69,7 @@ export const GamePairingProvider = ({ children }: { children: ReactNode }) => {
...
           });
-          navigate(PLAY_GAME_URL);
-          // `${PLAY_GAME_URL}/${gameRoomId}` should probably be the final
-          // navigation, but PLAY_GAME_URL is where our game prototype waits,
-          // for now
-          // navigate(`${PLAY_GAME_URL}/${gameRoomId}`, { replace: true });
+          navigate(`${PLAY_GAME_URL}/${gameRoomId}`, { replace: true });
           break;
         }
         // The handling of these two is similar and should be handled
```
Damrod, line 75: We don't need this comment anymore.

webapp/src/shared/types.ts
```diff
@@ -28,3 +30,12 @@ export type Avatar = {
   avatarX: number;
   avatarY: number;
 };
+
+type PlayState = 'playing' | 'paused';
```
Damrod, line 34: I think this should be already in pong engine, and imported from there, since we also have it and use it in game.service.ts

webapp/src/shared/types.ts
```diff
@@ -28,3 +30,12 @@ export type Avatar = {
...
+
+export type GameInfo = {
```
Damrod, line 36: Actually this one is missing some of the fields that that one has.
What about creating a common one, and have an intersection type with the common one + the extra ones in game.service.ts, and just the common one here?

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
```diff
@@ -27,7 +29,11 @@ export default function ChatBubble({
   return (
     <div className={`chat-bubble chat-bubble-${variant}`}>
       <div className="chat-bubble-avatar">
-        {avatar && !isConsecutive && <SmallAvatar {...avatar} />}
+        {avatar && !isConsecutive && (
```
Damrod, line 32: Since we're already here changing this what do you think about changing this as well:

```diff
diff --git a/webapp/src/shared/components/ChatBubble/ChatBubble.tsx b/webapp/src/shared/components/ChatBubble/ChatBubble.tsx
index 0ca0a38..c8c9e41 100644
--- a/webapp/src/shared/components/ChatBubble/ChatBubble.tsx
+++ b/webapp/src/shared/components/ChatBubble/ChatBubble.tsx
@@ -41,7 +41,7 @@ export default function ChatBubble({
         <div className={'chat-bubble-text-author'}>
           {name && isFirst && (
             <Text
-              variant={TextVariant.CAPTION}
+              variant={TextVariant.PARAGRAPH}
               color={TextColor.LIGHT}
               weight={TextWeight.MEDIUM}
               children={name}
@@ -50,7 +50,7 @@ export default function ChatBubble({
         </div>
         <div className={`chat-bubble-text chat-bubble-text-${variant}`}>
           <Text
-            variant={TextVariant.CAPTION}
+            variant={TextVariant.PARAGRAPH}
             color={TextColor.LIGHT}
             weight={TextWeight.REGULAR}
             children={text}
```

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -88,59 +167,21 @@ const Wait = () => {
...
-      {/* TODO , this is a temporal fix, replace with a menu */}
+    <GameStateContextProvider>
       <Header
         icon={IconVariant.ARROW_BACK}
         onClick={() => {
-          leaveGame();
           goBack();
         }}
       >
         Hit the brick!
```
Damrod, line 180: I'm pretty sure there's no brick here to hit.

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -1,78 +1,157 @@
...
-    canvasContext && renderFrame(canvasContext, gameState);
-    window.requestAnimationFrame(() => gameLoop());
-  }, [runGameFrame, renderFrame]);
+  React.useEffect(() => {
+    return () => {
+      if (gameJoined) {
+        socket.emit(LEAVE_GAME, { gameRoomId: gameId });
+      }
+    };
+  }, [gameId, gameJoined]);
```
Damrod, line 74: I see that this works, because the game kind of works, but I don't understand how.

We enter this `Play` function, and we initialize gameJoin to false with useState. Eventually we click, and the gameJoin becomes true. Then this useEffect should be triggered with gameJoin: true, ... which should have the effect of emitting a message to leave the game!

I see this is working, but I don't understand how.

webapp/src/shared/components/Game/Game.tsx
```diff
@@ -1,78 +1,157 @@
...
+    function handleGameEnded() {
+      notify('Game ended');
+      navigate(PLAY_URL, { replace: true });
+    }
+
+    socket.on(GAME_JOINED, handleGameJoined);
+    socket.on(UPDATE_GAME, handleUpdateGame);
+    socket.on(GAME_NOT_FOUND, handleGameNotFound);
+    socket.on(GAME_FINISHED, handleGameEnded);
+    socket.on('exception', (wsError: WsException) => {
```
Damrod, line 114: We're subscribed to the
`exception` ws message in at least three places, two of them overlapping, in two places.
We are subscribed in the GamePairingContext, and here, and in the GamePairingContext and somewhere else. I've seen that this doesn't appear to break anything, but maybe it will bring problems? Maybe we'd like to opena  ticket?

transcendence-app/src/game/game.gateway.ts
```diff
@@ -15,99 +16,47 @@ import { Socket, Server } from 'socket.io';
...
-      this.server.to(userId).emit('joinGame', { res: 'ok' });
-    }, 2000);
+  handleGameJoin(
+    @ConnectedSocket() client: Socket,
+    @MessageBody('gameRoomId', ParseUUIDPipe) gameRoomId: string,
+  ) {
+    this.gameService.handleGameJoin(client, gameRoomId);
   }
 
   @SubscribeMessage('leaveGame')
```
Damrod, line 42: I have to say that I would like to see a websocket sequence diagram for all of these, but maybe not for now. It is confusiong to just see them in the code. It's hard to catch bugs.

Damrod (approved):
Feel free to merge after you take a look at my comments and decide what to do about it.

webapp/src/shared/components/Game/Game.tsx
gbudau: Updated. I've added it back and applied the changes to the current implementation.

webapp/src/shared/context/GamePairingContext.tsx
gbudau: Done :+1:

Damrod:
I have seen this, as well:

Screenshot from 2023-02-09 14-42-09


Maybe it was me who introduced the bug? I'm not sure.

webapp/src/shared/components/Game/Game.tsx
gbudau: Updated. I've removed this condition and moved the leave game emit in the cleanup function of the other `useEffect`.

webapp/src/shared/components/Game/Game.tsx
gbudau: I've removed the listener from this component. Maybe we could remove all and leave only the one from the `GamePairingContext` since it's the outermost ws socket context?

Damrod: Yes, good idea. Don't necessarily do it here and now, though.  Should I open an issue?

gbudau: Done, #489

gbudau:
> I have seen this, as well:
> 
> Screenshot from 2023-02-09 14-42-09
> 
> Maybe it was me who introduced the bug? I'm not sure.

Thanks, its fixed now :sweat_smile:

webapp/src/shared/components/ChatBubble/ChatBubble.tsx
gbudau: Lgtm :+1:, done

webapp/src/shared/types.ts
gbudau: Updated. I've moved the types to the pong engine and split them into `GameInfoClient` and `GameInfoServer`. Naming suggestions are appreciated :smile:

transcendence-app/src/game/game.gateway.ts
gbudau: I'll try to make one :+1:

gbudau: Done: https://excalidraw.com/#room=49a2157c70806baa3387,r-Mkp4AvdX_K__tHj4_4Hw

Damrod: As far as I can see, 'updateGame' is sent to gameId, not to the separate userIds. What do you think about the server withholding sending the 'gameJoined' response until you have both clients (the first one joins, and waits to see a 'gameJoined', just like it does now), and when the second player sends 'joinGame' and you can have its client join the room, then send 'gameJoined' to the gameroom, at the same time?

Damrod: And what happens with the disconnects and with our game map if a user disconnects after its opponent has sent 'gameJoined' and before the disconnecting player sends 'gameJoined'?

Damrod: Otherwise you have a total of three layers of cascading wait, the third one silent and disconcerting.

Damrod: But otherwise, the wait will be with the button, and will make him/her think that the button didn't work.... mmmh

Damrod: It's okay the way it is, I guess.

Damrod (changes requested):
> > I have seen this, as well:
> > Maybe it was me who introduced the bug? I'm not sure.
> 
> Thanks, it's fixed now. It was because of one of my changes sweat_smile

The queue page is now not working. We're being expelled from /play/queue as soon as we join it.

gbudau:
> The queue page is now not working. We're being expelled from /play/queue as soon as we join it.

I think this happens because the user is redirected from the client while the context still has 'none' as their status, and they get the "waiting" confirmation from the server, which can take a while.

I see two possible ways to fix this:

Remove the queue page. The users join a queue on the backend, and they can still quit waiting from the play page, and we won't redirect them to a queue page.
Set the context to waiting before the backend confirmation and redirect only the client/tab that joined the queue to the queue page.
Which one do we prefer?

transcendence-app/src/game/game.gateway.ts
gbudau: I agree that as it is now, it's a bit confusing since the users don't have a way to know the game state.

I think this will be better when we implement the pause notification.

Damrod:
> Set the context to waiting before the backend confirmation and redirect only the client/tab that joined the queue to the queue page.

That's how it was. I don't see a reason to remove the queue page.

gbudau:
> > Set the context to waiting before the backend confirmation and redirect only the client/tab that joined the queue to the queue page.
> 
> That's how it was. I don't see a reason to remove the queue page.

Done :+1:. I prefer the first option because showing the queue page prevents the user from navigating the app while not adding many benefits (the users can still use a mouse or keyboard shortcut to go back or open a new tab and close the first one with the queue page, and they will still be in the queue)

gbudau:
> The queue page is now not working. We're being expelled from /play/queue as soon as we join it.

After discussing this on Slack, we decided to remove the queue page.

Damrod (approved):
Now, hopefully, LGTM!

vdedios (approved):

PR #472, opened 2023-01-31
[2] #473 / multiplayer pong client

### LINKED ISSUE #473: Multiplayer game in client
subtask of #400

gbudau:
closes #473


webapp/src/shared/components/Game/GameMultiplayer.tsx
```diff
@@ -0,0 +1,59 @@
...
+        if (requestFrameRef.current) {
+          cancelAnimationFrame(requestFrameRef.current);
+        }
+      };
+    }
+  }, [gameLoop]);
+
+  return (
+    <div className="game-multiplayer">
+      <h1 className="game-multiplayer-score heading-bold">{opponentScore}</h1>
```
Damrod, line 46: why don't we use the `Text` component we created?

packages/pong-engine/src/game.ts
```diff
@@ -19,14 +28,26 @@ export const paddleMoveRight = (state: GameState): GameState => {
   return reducer(state, { type: 'paddleMoveRight', payload: {} });
 };
 
+export const paddleOpponentMoveRight = (state: GameState): GameState => {
```
Damrod, line 31: I think this concept of naming these commands `xxxOpponentXxx` and `xxxXxx` is not a really good idea. When we have two users interacting, which is the main player and which is the opponent? Why not just player A and player B?

packages/pong-engine/src/game.ts
```diff
@@ -58,3 +81,21 @@ export const runGameFrame = (
...
   return reducer(state, { type: 'move', payload: { deltaTime } });
 };
+
+export const runGameMultiplayerFrame = (
+  deltaTime: number,
+  state: GameState,
+): GameState => {
+  const { ball } = state;
+
+  if (losePointMultiplayer(ball)) {
```
Damrod, line 91: IMHO multiplayer should be the non suffixed default. We shouldn't have to say `losePointMultiplayer` vs `losePoint`. It should rather be `losePoint` vs `losePointSinglePlayer`, the first one not defaulted.

packages/pong-engine/src/physics.ts
```diff
@@ -53,6 +51,32 @@ const bounceArkanoid = (
...
+): GameBall => {
+  const paddleBallDistance = ball.x - paddle.x;
+  const bounceAngle =
+    MAX_PADDLE_BOUNCE_ANGLE -
+    (paddleBallDistance / PADDLE_WIDTH) *
+      (MAX_PADDLE_BOUNCE_ANGLE - MIN_PADDLE_BOUNCE_ANGLE);
+  return getArkanoidBounce(ball, bounceAngle, deltaTime);
+};
+
+const bounceArkanoidOpponent = (
```
Damrod, line 67: `bounceArkanoid` and this are the same except for `MXX_PADDLE_OPPONENT_BOUNCE_ANGLE` vs `MXX_PADDLE_BOUNCE_ANGLE`. Why not packing these player dependant constants into a config object, and passing this object to different functions to return the player specific `playerA` and `playerB` behaviour?

Damrod:
As it is right now in production, the touch events are not working. Will we stop supporting them? Or is there something we can do about it?
Thank you

Damrod:
Btw now I see I wasn't requested for review, despite the fact that the PR is not marked as a draft anymore. So I'm gonna withold my approval and see that you mean to ask for review.

Damrod:
I think we should put the `npm run build-packages` command in `make all` rule, and by extension the rest of the rules that depend on `all`. I spend an average of ten minutes every time trying to find why webapp cannot find the exported symbols from the pong-engine package.

Damrod (approved):
besides the comments, LGTM!

Makefile
```diff
@@ -44,6 +45,7 @@ clean:
 	make down
 	rm -rf $(PROJECT_ROOT)/webapp/src/shared/generated || true
 	rm $(PROJECT_ROOT)/transcendence-app/swagger-spec/swagger-spec.yaml || true
+	rm -rf $(PROJECT_ROOT)/packages/pong-engine/lib || true
```
vdedios, line 48: I think this is not needed because when we run npm run build-packages we are already deleting the generated code

packages/pong-engine/src/game.ts
vdedios: The opponent is just a point of view, when you are the player the opponent is the other and viceversa. If we name them as playerA and playerB that would be harder to understand since it does not have any semantic meaning

Makefile
gbudau: Done :+1:

packages/pong-engine/src/game.ts
vdedios: This is done like this to have retro compatibility with the game training game.

We could change it but making one be the multiplayer or the other being the training wouldn’t be much different, the thing is to have clear which one is which

packages/pong-engine/src/physics.ts
vdedios: This has, kind of, already been done, we’ve move the common part of  both functions to a common one. These angle calculations are specific to each bounce and this code has be somewhere. Also does not make sense to move the angles outside it because is only responsibility of the bounce function to know about them.

vdedios:
> As it is right now in production, the touch events are not working. Will we stop supporting them? Or is there something we can do about it?

As commented by chat, this is not currently supported but it will in next PRs. We decided to go for the minimal functionality

gbudau:
Thank you!

PR #478, opened 2023-02-05
Fix library package compilation in development mode

### LINKED ISSUE #477: Fix library package compilation in development mode

vdedios:
closes #477 

When the project is up and running in development mode, we are facing some issues with library builds:

- If rebuilding the lib, the webapp and transcendence app warn about not being able to find the library causing we have to re launch the app with `make re` or `make`
- The first time we launch the app with a simple `make` we get the same error.

This is happening because we are deleting the generated ts `lib` every time we rebuild it. Since each of webapp and transcendence-app link to the build package, due to it being deleted, they display that error.


gbudau (approved):
Thanks! LGTM :rocket:

Issue #483, opened 2023-02-06
refactor game queue gateway/service to remove code duplication


Damrod:
as described

Damrod:
As discussed with gbudau, game takes ownership of games ongoing model. This ticket and its PR are not necessary or at the very least delayed since they may not be needed at all.

PR #484, opened 2023-02-06
#483/refactor queue

Damrod:
and better handle user disconnect (bugfix)


Damrod:
As discussed with gbudau, game takes ownership of games ongoing model. This PR and its ticket are not necessary or at the very least delayed since they may not be needed at all.

Issue #378, opened 2022-11-30
Allow users logged in with OAuth have a password or hide the edit password button


fizcamposancos:
thought: The users who log in with OAuth have a `null` password.

question: Should we allow them to assign a password so they can also log in with a username + password?
Or do we want to hide this button for OAuth users?

suggestion: To check if a user is local (username + password) or from an OAuth provider, we could add a new field to the user entity, `isLocal`?, which we serialize and assign to true if the user has a password or false if they don't, similar to what we have for the chatroom entity (`isPublic`).

Maybe we could leave a TODO until we can test this.

_Originally posted by @gbudau in https://github.com/42AGV/ft_transcendence/pull/374#discussion_r1035119149_

PR #486, opened 2023-02-07
#378/hide edit password for non local users

Damrod:
as explained in #378


Damrod:
This works with 42 login.

It's not the cleanest solution (to add isLocal to the GET authorization endpoint), but it's way cleaner and simpler than adding it to user.entity (we have many many things that would break if we add it there)

gbudau (approved):
Nice solution! Also, thanks for updating the readme. LGTM :tada:

Issue #489, opened 2023-02-09
Multiple ws 'exception' listeners


gbudau:
We could remove the listeners for this event from the components and only have one in the `GamePairingContext` since it wraps up our entire app.

PR #490, opened 2023-02-09
#489/webapp websocket exception handler now only in one place

Damrod:
As explained in the ticket #489


gbudau (approved):

Issue #476, opened 2023-02-05
Share websockets DTOs between transcendence-app and webapp


Damrod:
This ticket tracks the resolution of giving a consistent and reliable solution for all of our websockets data model sharing problems. E.g.:

Just as we share data models with the open-api generator between transcendence-app and webapp for http requests/responses, we need to share data models for websocket events. These are not generated by the openapi-generator in it's current form.

The patchwork of ways we have to do this so far is:

- we have some unused http endpoints the backend, just for the sake of having openapi-generator generate the code in webapp.
- we have some DTOs in the file `packages/pong-engine/src/queue.dto.ts` that don't really belong in pong-engine, since they don't belong to the game engine (but do belong to the game problem space), as interfaces.
- we are creating duplicated models by hand in frontend and backend.
- We are not typing the DTOs, and sending them as raw JS objects.

Maybe we would like to create classes for all such dtos in order to have classes validators and serializers for the incoming data.
As proposed tasks, not in a particular order, and not particularly prioritized or necessary:

**Investigation of possible solutions:**
- [ ] maybe an expanded openapi generator. 
- [ ] maybe a different package or program.
- [ ] Maybe a package we can add by hand to the private packages of our monorepo (`./packages` folder), together with some coding style or readme on how to use it consistently.

**Action:**
- [ ] refactor the old code to stop using the different patchwork of ways to do what they were doing.
- [ ] teach others how to do this the new way.

PR #491, opened 2023-02-09
#476/package to share data models between transcendence-app and webapp

Damrod:
As explained in #476

This PR prepares an extra private package (transcendence-shared) with a rough draft of how can it work, in order to enable others to use it to export their data types


Damrod:
I have investigated asyncApi (the async solution analogous to openApi), and I've deemed that doing what we need with it has a work overhead that is too large, for too little value in our use case. Besides this, I think we may benefit from using namespaces. But I think the number of changes in this PR is large enough as it is, and further work should happen in further PRs.

fizcamposancos (approved):
I think I got the concept but I would need futther explanation. And, AFAIK, everything looks good.

docker-compose.override.yml
```diff
@@ -12,7 +12,7 @@ services:
       target: development
     volumes:
       - ./webapp/src:/service/webapp/src:ro
-      - ./packages/pong-engine:/service/packages/pong-engine:ro
+      - ./packages:/service/packages:ro
```
gbudau, line 15:
```suggestion
```

suggestion: We could remove this since we need to rebuild the images if we change a package.

Same suggestion for the transcendence-app volume.

PR #500, opened 2023-02-13
#495/Finish the game when a player reaches an arbitrary number of points

### LINKED ISSUE #495: Finish the game when a player reaches an arbitrary number of points
The game should end when a player reaches an arbitrary number of points.

gbudau:
close #495

### DIFF
--- a/transcendence-app/src/game/game.service.ts
+++ b/transcendence-app/src/game/game.service.ts
@@ -25,6 +25,7 @@ type ClientId = string;
 const FPS = 30;
 const DELTA_TIME = 1 / FPS;
 const MAX_PAUSED_TIME_MS = 30 * 1000; // 30 seconds
+const MAX_SCORE = 10;
 
 @Injectable()
 export class GameService {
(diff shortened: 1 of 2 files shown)


Damrod:
It works. If anything, I'd either send an event to the clients with the final score so it can be reported, or run the animation until AFTER the scoreboard has been updated with the required score. But these things can happen in a different PR.

transcendence-app/src/game/game.service.ts
```diff
@@ -25,6 +25,7 @@ type ClientId = string;
 const FPS = 30;
 const DELTA_TIME = 1 / FPS;
 const MAX_PAUSED_TIME_MS = 30 * 1000; // 30 seconds
+const MAX_SCORE = 10;
```
Damrod, line 28: I think eventually we should try to pass all of these constant as a configuration object, and pull from the configuration, so that we can, you know, have powerups and such, as required by the subject. The whole "constants" concept is highly inflexible for our use case.

Out of scope, I'll admit.

Damrod (approved):
Besides the comments, LGTM! Thanks for this feature!

fizcamposancos (approved):
Good job!!

gbudau:
> It works. If anything, I'd either send an event to the clients with the final score so it can be reported, or run the animation until AFTER the scoreboard has been updated with the required score. But these things can happen in a different PR.

Updated. Thanks for the suggestions!

transcendence-app/src/game/game.service.ts
gbudau: :+1: Let's track this in #506

PR #502, opened 2023-02-13
#488/List ongoing games and update them in real-time

### LINKED ISSUE #488: List ongoing games and update them in real-time

gbudau:
close #488

### DIFF
--- a/webapp/src/pages/PlayPage/PlayPage.tsx
+++ b/webapp/src/pages/PlayPage/PlayPage.tsx
@@ -3,38 +3,55 @@ import { ButtonVariant, IconVariant } from '../../shared/components';
 import { useNavigate } from 'react-router-dom';
 import { MainTabTemplate } from '../../shared/components/index';
 import { SearchContextProvider } from '../../shared/context/SearchContext';
-import { PLAY_GAME_TRAIN_URL } from '../../shared/urls';
+import { PLAY_GAME_TRAIN_URL, PLAY_GAME_URL } from '../../shared/urls';
 import { ENTRIES_LIMIT } from '../../shared/constants';
 import { Query } from '../../shared/types';
 import socket from '../../shared/socket';
 import { useGamePairing } from '../../shared/hooks/UseGamePairing';
 import { gameQueueClientToServerWsEvents } from 'transcendence-shared';
-import { GamePairingStatusDtoGameQueueStatusEnum } from '../../shared/generated';
+import {
+  GamePairingStatusDtoGameQueueStatusEnum,
+  User,
+} from '../../shared/generated';
 import { useCallback, useMemo } from 'react';
+
+type GameWithPlayers = {
+  gameId: string;
+  playerOne: User;
+  playerTwo: User;
+};
 
 export default function PlayPage() {
   const { gameQueueStatus } = useGamePairing();
   const navigate = useNavigate();
+  const [ongoingGames, setOngoingGames] = React.useState<GameWithPlayers[]>([]);
 
-  // To be replaced with real games request
-  const mockFetchFn = React.useCallback(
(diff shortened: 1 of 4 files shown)


webapp/src/pages/PlayPage/PlayPage.tsx
```diff
@@ -47,6 +68,33 @@ export default function PlayPage() {
     },
   ];
 
+  React.useEffect(() => {
+    function ongoingGamesListener(games: GameWithPlayers[]) {
```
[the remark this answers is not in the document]
gbudau: To check if a user is playing, we could store a set of player user IDs for an O(1) lookup.

If we store a map of game IDs and the players as the value of that map, like on this page, the lookup is less efficient, O(n), where n is the number of ongoing games.

webapp/src/pages/PlayPage/PlayPage.tsx
```diff
@@ -58,9 +108,46 @@ export default function PlayPage() {
...
 
+  if (!authUser) {
+    return <LandingPage />;
+  }
+
   return (
-    <SearchContextProvider fetchFn={mockFetchFn} maxEntries={ENTRIES_LIMIT}>
-      <MainTabTemplate dataMapper={mockedMapper} buttons={buttons} />
-    </SearchContextProvider>
+    <div className="play-page">
```
gbudau, line 116: There might be some duplicated code, but I couldn't make it work with the templates because they use HTTP calls.

webapp/src/pages/PlayPage/PlayPage.tsx
```diff
@@ -58,9 +108,46 @@ export default function PlayPage() {
...
+            iconVariant={IconVariant.SEARCH}
+            placeholder="search"
+            value={search}
+            name="search"
+            onChange={(e) => setSearch(e.target.value)}
+          />
+        </div>
+      </div>
+      <div className="play-page-games">
+        <RowsList rows={gameMapper()} />
```
gbudau, line 139: todo (in other PRs): We could replace the row list with a list of items similar to the score component.

webapp/src/pages/PlayPage/PlayPage.tsx
```diff
@@ -1,42 +1,65 @@
...
-      },
-      url: '#',
-      title: '',
-      subtitle: 'level x',
-      key: '',
-    }),
-    [],
-  );
+  function gameMapper(): RowItem[] {
+    return ongoingGames.reduce<RowItem[]>((rowItems, currentGame) => {
```
gbudau, line 47: This is the same as filter and map but a bit more efficient. If it's hard to read, we could change it.

webapp/src/pages/PlayPage/PlayPage.tsx
```diff
@@ -1,42 +1,65 @@
...
-import { GamePairingStatusDtoGameQueueStatusEnum } from '../../shared/generated';
+import {
+  GamePairingStatusDtoGameQueueStatusEnum,
+  User,
+} from '../../shared/generated';
+import { Link } from 'react-router-dom';
+import { useAuth } from '../../shared/hooks/UseAuth';
+import { LandingPage } from '..';
+
+type GameWithPlayers = {
```
Damrod, line 33: We could share this data model somewhere in transcendence-shared/pong engine, couldn't we?

Damrod: I see. That's complicated. I imagine the packages cannot forward reference a type, or import it from transcendence-app. Let's forget about this for now. I'll investigate some time in the future, anyway.

Damrod (approved):
LGTM!
Thanks for this feature!

gbudau:
Thanks!

Issue #503, opened 2023-02-13
persist game queue action/display in UI


Damrod:
When we join the game queue, or are  challenged or challenge someone, the waiting state is persisted in the backend, and in the play page. The persisted state is not in any way represented in the front end otherwise. This ticket explores different ways to bring about such persistence.

PR #504, opened 2023-02-13
#503/persist queued/playing UI status and action

Damrod:
As described in #503

### DIFF
--- a/webapp/src/shared/components/Button/Button.css
+++ b/webapp/src/shared/components/Button/Button.css
@@ -65,3 +65,10 @@
   position: absolute;
   left: 0.675rem;
 }
+
+.button-bubble {
+  z-index: 99;
+  position: fixed;
+  right: -0.5rem;
+  top: -0.5rem;
+}
--- a/webapp/src/shared/components/Button/Button.css
+++ b/webapp/src/shared/components/Button/Button.css
@@ -65,10 +65,3 @@
   position: absolute;
   left: 0.675rem;
 }
-
-.button-bubble {
-  z-index: 99;
-  position: fixed;
-  right: -0.5rem;
-  top: -0.5rem;
-}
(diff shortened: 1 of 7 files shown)


webapp/src/shared/components/Button/Button.css
```diff
@@ -65,3 +65,10 @@
   position: absolute;
   left: 0.675rem;
 }
+
+.button-bubble {
+  z-index: 99;
+  position: fixed;
+  right: -0.5rem;
+  top: -0.5rem;
+}
```
fizcamposancos, line 74: Don´t you think the cropped button looks a bit weird ? What about something like this?

```suggestion
}.button-bubble {
  z-index: 99;
  position: fixed;
  right: -0.2rem;
  top: -0.2rem;
}

.button-bubble > .button-chip {
  border-radius: 0.5rem;
}
```
```

**Review by fizcamposancos — APPROVED** [review]
I like the idea of showing in the front end the waiting state, and making it interactive

**Comment by Damrod** [comment]
I'm going to be thinking a bit more about the button. Maybe create a new component. I'm not satisfied.

**Comment by Damrod** [comment]
I think now it's a bit prettier. Please, comment.

File: webapp/src/shared/components/Button/Button.css (hunk 1 above)
Comment (Damrod) [reply webapp/src/shared/components/Button/Button.css]: take a look at the latest proposition, please

File: webapp/src/shared/components/Button/CollapsibleButton.css [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+  height: 3rem;
+  width: auto;
+  position: relative;
+}
+
+.collapsible-button-uncollapsed .button-chip {
+  z-index: 98;
+  width: 100%;
+  position: absolute;
+  top: 0.5rem;
```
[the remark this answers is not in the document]
Damrod: done

Issue #505, opened 2023-02-14
create Game entry in game table upon finishGame


Damrod:
When a user finishes a game, transcendence app (game.service.ts) should create a relevant entry in the game table in order to persist user stats as required.

PR #511, opened 2023-02-14
#505/create game entry in database after game finished

Damrod:
as explained in #505

### DIFF
--- a/transcendence-app/src/game/game.service.ts
+++ b/transcendence-app/src/game/game.service.ts
@@ -17,17 +17,21 @@ import { WsException } from '@nestjs/websockets';
 import { OnEvent } from '@nestjs/event-emitter';
 import { GameInputDto } from './dto/game-input.dto';
 import { GamePairing } from './infrastructure/db/game-pairing.entity';
+import { User } from '../user/infrastructure/db/user.entity';
+import { IUserRepository } from '../user/infrastructure/db/user.repository';
 import { IGameRepository } from './infrastructure/db/game.repository';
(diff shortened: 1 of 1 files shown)


Damrod:
Max SMALLINT can store is about 32000 seconds, or about 9 hours. We should kick the users if they stay longer than 9 hours, otherwise the database will die, complaining that we're trying to pass it too large number, and the server will be down.
I'd have chosen INT rather than SMALLINT, I think...

Damrod:
Throwing exceptions from the game service does not appear to bubble up all the way to the front end, for some reason. Would gladly accept help in finding out about this.

gbudau (approved):

gbudau:
> Throwing exceptions from the game service does not appear to bubble up all the way to the front end, for some reason. Would gladly accept help in finding out about this.

Not sure why it doesn't work in this case, and it also crashes the server :(

Maybe we could log the error without throwing a `WsException`?

I don't think it happens because it's in the service. For example, the game not found `WsException` works.

transcendence-app/src/game/game.service.ts
```diff
@@ -54,6 +96,9 @@ export class GameService {
       .emit('gameStatusUpdate', {
         status: GameStatus.FINISHED,
       } as GameStatusUpdateDto);
+    this.addGameWhenFinished(gameInfo, gameId).catch((error: Error) => {
+      throw new WsException(error.message);
```
gbudau, line 100:
```suggestion
      this.logger.error(error.message);
```

Issue #518, opened 2023-02-15
add level column to user table


Damrod:
Part of #508

Damrod:
This is outdated

PR #519, opened 2023-02-15
#518/add level column to user table

Damrod:
as described in the ticket #518


Damrod:
We don't need this. I'll also be removing the branch later.

PR #520, opened 2023-02-15
#496/Enable paddle drag in online games

### LINKED ISSUE #496: Enable paddle drag in online games

gbudau:
close #496

### DIFF
--- a/webapp/src/shared/components/Game/hooks/useOnlineGame.tsx
+++ b/webapp/src/shared/components/Game/hooks/useOnlineGame.tsx
@@ -26,11 +26,15 @@ export function useOnlineGame(gameId: string) {
   const [gameJoined, setGameJoined] = React.useState(false);
 
   const sendGameCommand = React.useCallback(
-    (command: GameCommand) => {
+    (
+      command: GameCommand,
+      payload?: { dragCurrPos: number; dragPrevPos: number },
+    ) => {
       if (isPlayer && gameJoined) {
         socket.emit(GAME_COMMAND, {
           command,
           gameRoomId: gameId,
+          payload,
         });
       }
     },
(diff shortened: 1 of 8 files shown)


webapp/src/shared/components/Game/hooks/useOnlineGame.tsx
```diff
@@ -26,11 +26,15 @@ export function useOnlineGame(gameId: string) {
   const [gameJoined, setGameJoined] = React.useState(false);
 
   const sendGameCommand = React.useCallback(
-    (command: GameCommand) => {
+    (
+      command: GameCommand,
+      payload?: { dragCurrPos: number; dragPrevPos: number },
```
Damrod, line 31: I'm going to say what I'm always saying: why don't we type this explicitly? 
```
class DragPayload {
  @IsNumber()
  dragCurrPos!: number;

  @IsNumber()
  dragPrevPos!: number;
}
```
seems to be just two numbers...
Although I'm not sure about the GameInputPayload...

You tell me.

Damrod (approved):
LGTM!
Thanks for this feature!

webapp/src/shared/components/Game/hooks/useOnlineGame.tsx
gbudau: > I'm going to say what I'm always saying: why don't we type this explicitly?

Done. Please let me know what you think.

> seems to be just two numbers...

I'm not sure what you mean.

Damrod: Just two numbers means that we don't have the problem we had yesterday with the User type.

Damrod: About the most recent changes: you can also put the class validator in the pong engine as well, and validate your object.

gbudau: Thanks :+1:
I prefer not to add the class validator library to the pong engine because it doesn't use it, and the validation happens on the server side.

gbudau:
Thanks!

Issue #549, opened 2023-02-20
Investigate internal server error when changing username


Damrod:
The snackbar warning is popping up with a string saying "Internal server error" whenever the authenticated user changes its username, and keeps popping up when navigating, as long as the app is not reloaded. Investigate where is this problem coming from and fix it. Investigate as well if this is happening elsewhere where we have a pattern (whatever it is) similar to the one that produces this effect.

UPDATE:

It is also happening when attempting to join a chatroom where the user is not a member of. It's not an internal server error, but still an unhandled ws exception that we can clean up.

PR #557, opened 2023-02-23
#549/ fix internal server error ws exception

Damrod:
As explained in #549

UPDATE:
This PR fixes the internal server errors, and also fixes some style issues we had across the whole website, where buttons that should be in matching places were not.


gbudau (approved):
Thank you for these fixes and improvements! LGTM :tada:

PR #560, opened 2023-02-24
#509/pagination offset problems when adding entries or removing already read entries

### LINKED ISSUE #509: pagination offset problems when adding entries or removing already read entries
Right now we are using offset and limit in our queries to retrieve and paginate some rows for users and chats in the front end.
Users should be persistent enough, since we don't have a mechanism to delete them. But chats may be deleted, which means that if someone deletes a chat we may have already read with a previous pagination query, the offset will be messed up.

Possibly the same goes for entries added to either users or chatroom tables.

We should research an consider action points about this possible problem.

gbudau:
These changes solve the duplicate entries bug when a new entry gets added before the offset.
When an entry before the offset gets deleted, it will still skip some entries :sweat_smile:.

It's not the most optimal solution, but it works for now.

The better option would be to implement the [seek method](https://use-the-index-luke.com/sql/partial-results/fetch-next-page) ([keyset pagination](https://use-the-index-luke.com/no-offset)), which requires changing all the pagination queries, the row list templates, and the search context.

close #509


Damrod:
If you try the following, you will still see a problem:

1. with a normally seeded instance of our app:
2. log with user `a` in a browser, and user `b` in a different browser.
3. create with user `b` some chat `(b)`
4. join chat `(b)` with both users,
5. go to `/chats` page and reload, with user `a`
6. with user `b`, go to the edit chat page `(b)`, and delete chat `(b)`.
7. scroll down with user `a`. The old 15th entry of the chat listing will be missing.

Damrod:
In the opposite case (where there is a duplicated key in the case a row was added), I see there were duplicated rows, and now it's fixed.

Damrod (approved):
I guess the case I'm reporting is not a bug anymore, just a quircky behaviour :D

transcendence-app/seeds/002_seed_chatrooms.ts
```diff
@@ -20,7 +20,9 @@ let startTime = new Date(Date.now());
 
 const createRandomChatroomMessage = (chatroomId: string, userId: string) => {
   const oldDate = startTime;
-  startTime = new Date(startTime.setSeconds(1 + startTime.getSeconds()));
+  startTime = new Date(
+    startTime.setMilliseconds(1 + startTime.getMilliseconds()),
```
Damrod, line 24: why just one millisecond of difference?

gbudau: While seeding the chat messages, some got created far in the future, and if we write new ones in a chatroom right after and then refresh the page, the new ones would get displayed on top of the seeded ones.
I thought it was because of some pagination bug and spent some time to figure out why it happened :sweat_smile:.

gbudau:
Thank you!

If we have time, we will implement the seek method, which should fix the potential missing entries :+1:

PR #561, opened 2023-02-24
fix lag time reference assignation

vdedios:
I pushed but forgot to save the file, so I missed this change suggested by @gbudau in last PR 😅


gbudau (approved):
Thanks! LGTM :tada:

PR #574, opened 2023-03-02
#573/Remove the updateGameInfo method from GameService

### LINKED ISSUE #573: Remove the updateGameInfo method from GameService
We might not need to check if the games has the gameInfo before updating it because node.js is a single-threaded event-driven system.
We could remove this check to improve the game performance.

gbudau:
close #573


vdedios (approved):
Not sure if we are going to gain much performance with this because, with the few games we are going to hold, iterating over the map would be a fraction of a ms. But agree we don’t need this anyway, thanks! 🚀

gbudau:
You are right. Thanks!

Issue #577, opened 2023-03-02
create additional game modes for stats


Damrod:
as explained in the title

PR #578, opened 2023-03-03
#577/add additional gamemodes to stats

Damrod:
as described in the ticket

### DIFF
--- a/webapp/src/pages/UserStatsPage/UserStatsPage.tsx
+++ b/webapp/src/pages/UserStatsPage/UserStatsPage.tsx
@@ -6,8 +6,8 @@ import { useParams } from 'react-router-dom';
 import {
   GameControllerGetUserLevelHistoryModeEnum,
   GameControllerGetUserStatsModeEnum,
-  GameStats,
   UserLevelWithTimestamp,
+  GameStats,
 } from '../../shared/generated';
 import { useData } from '../../shared/hooks/UseData';
 import NotFoundPage from '../NotFoundPage/NotFoundPage';
--- a/webapp/src/pages/UserStatsPage/UserStatsPage.tsx
+++ b/webapp/src/pages/UserStatsPage/UserStatsPage.tsx
@@ -114,6 +114,9 @@ export default function UserStatsPage() {
   if (areLevelsLoading || isUserLoading || areStatsLoading)
     return <LoadingPage />;
   if (!levels || !user || !stats) return <NotFoundPage />;
+  const toArray = (stats: GameStats) => {
+    return [stats.wins, stats.losses, stats.draws];
+  };
   return (
     <div className="stats-page">
       <Header icon={IconVariant.ARROW_BACK} onClick={goBack}>
@@ -165,11 +168,15 @@ export default function UserStatsPage() {
               labels: ['win', 'loss', 'draw'],
               datasets: [
                 {
-                  data: stats.map((stat) => [
-                    stat.wins,
-                    stat.losses,
-                    stat.draws,
-                  ]),
+                  data: toArray(
+                    stats.reduce((prev, curr) => {
+                      return {
+                        wins: prev.wins + curr.wins,
+                        losses: prev.losses + curr.losses,
+                        draws: prev.draws + curr.draws,
+                      };
+                    }),
+                  ),
                 },
               ],
             }}
--- a/webapp/src/pages/UserStatsPage/UserStatsPage.tsx
+++ b/webapp/src/pages/UserStatsPage/UserStatsPage.tsx
@@ -1,5 +1,5 @@
-import 'chartjs-adapter-date-fns';
 import './UserStatsPage.css';
+import 'chartjs-adapter-date-fns';
 import { Doughnut, Line } from 'react-chartjs-2';
 import { gameApi, usersApi } from '../../shared/services/ApiService';
 import { useCallback } from 'react';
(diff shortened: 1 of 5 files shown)


Damrod:
The workflows are not running, don't know why

webapp/src/pages/UserStatsPage/UserStatsPage.tsx
```diff
@@ -17,44 +16,107 @@ import 'chartjs-adapter-date-fns';
...
-    useData<UserLevelWithTimestamp[]>(getLevelHistory);
+  const { data: user, isLoading: isUserLoading } = useData(getUserByUserName);
+  const getLevelHistory = useCallback(async (): Promise<
+    FiveArray<UserLevelWithTimestamp[]>
+  > => {
+    try {
+      if (!user) {
+        return Promise.reject('Unresolved');
+      }
+      let pclassic: proArray = [
```
Damrod, line 39: There's a better way to do it: populate all arrays with the fake entry first, and THEN use the callbacks.
But I was too exhausted last night to be able to see it.

webapp/src/pages/UserStatsPage/UserStatsPage.tsx
```diff
@@ -17,44 +16,107 @@ import 'chartjs-adapter-date-fns';
...
+        [
+          GameControllerGetUserStatsModeEnum.Classic,
+          GameControllerGetUserStatsModeEnum.ShortPaddle,
+          GameControllerGetUserStatsModeEnum.MysteryZone,
+          GameControllerGetUserStatsModeEnum.Training,
+          GameControllerGetUserStatsModeEnum.Unknown,
+        ].map((mode) => {
+          return gameApi.gameControllerGetUserStats({
+            username: username!,
+            mode,
```
Damrod, line 106: mode is optional. If I don't pass it, I get all of them, and then I don't need the reducer.

Damrod: will do.

webapp/src/pages/UserStatsPage/UserStatsPage.tsx
Damrod: will do.

gbudau: suggestion: We could fetch all the user levels and convert them to an object where the key is the game mode, and the value is the levels for that game mode. This way, we make one fetch and process it on the client.

<details>

```tsx
type GameModeLevels = {
  [key in GameControllerGetUserLevelHistoryModeEnum]: UserLevelWithTimestamp[];
};

...

  const getGameModeLevels = (): GameModeLevels => {
    let initialGameModeLevels: GameModeLevels = {
      training: [],
      classic: [],
      shortPaddle: [],
      mysteryZone: [],
      unknown: [],
    };
    if (!levels) {
      return initialGameModeLevels;
    }
    const gameModesWithLevels = levels.reduce<GameModeLevels>(
      (gameModeLevels, level) => {
        switch (level.gameMode) {
          case 'training':
            return {
              ...gameModeLevels,
              training: [...gameModeLevels.training, level],
            };
          case 'classic':
            return {
              ...gameModeLevels,
              classic: [...gameModeLevels.classic, level],
            };
          case 'mysteryZone':
            return {
              ...gameModeLevels,
              mysteryZone: [...gameModeLevels.mysteryZone, level],
            };
          case 'shortPaddle':
            return {
              ...gameModeLevels,
              shortPaddle: [...gameModeLevels.shortPaddle, level],
            };
          case 'unknown':
            return {
              ...gameModeLevels,
              unknown: [...gameModeLevels.unknown, level],
            };
          default:
            return gameModeLevels;
        }
      },
      initialGameModeLevels,
    );
    return gameModesWithLevels;
  };

...

          <Line
            data={{
              datasets: Object.entries(getGameModeLevels()).map((entry) => {
                const [gameMode, gameModeLevels] = entry;
                return {
                  label: gameMode,
                  data: gameModeLevels.map((item) => ({
                    x: item.timestamp,
                    y: item.level,
                  })),
                  borderWidth: 1,
                  stepped: true,
                };
              }),
            }}
            ... />

```

</details>

gbudau (approved):

webapp/src/pages/UserStatsPage/UserStatsPage.tsx
Damrod: We'll have to change the endpoint as well. As of now, without a mode, it returns only the ones for classic mode. Will do.

Damrod: Also, I'll still need to add a fake entry in the beginning, when the user registers. Any other solution to that?

gbudau: > We'll have to change the endpoint as well. As of now, without a mode, it returns only the ones for classic mode. Will do.

You are right :+1:

> Also, I'll still need to add a fake level 1 entry in the beginning of all game modes, when the user registers. Any other solution to that?

I think we don't need to add the fake entry, or am I missing something?

Damrod: Well, unlike a XP oriented system, in an elo system every user, when coming to existence, is assigned a level. I do think we need it, if only to provide a real progression of the level, and to make it pretty. I cannot insert it in the database because the user levels are referenced to games, also I wouldn't want to. It's a matter of representation, which is webapp responsibility.

Bottom line: I absolutely intend to add it.

vdedios (approved):
LGTM! 🚀

webapp/src/pages/UserStatsPage/UserStatsPage.tsx
Damrod: done

webapp/src/pages/UserStatsPage/UserStatsPage.tsx
Damrod: done

PR #584, opened 2023-03-09
#582/fix score row appearance

### LINKED ISSUE #582: ⚛️ Fix ScoreRow list appearance
Something is not right with the ScoreRow on the mobile version, An horizontal scroll makes the rows move in a strange way.

Also, 

- Alignment is not matching
- Usernames font size to be adjusted
- Gap between elements

fizcamposancos:
Close #582 
Sorry, forgot to fix horizontal scroll bar


gbudau (approved):
Thanks for the fix :+1:

PR #589, opened 2023-03-14
#585/get users with levels in single query

### LINKED ISSUE #585: refactor how do we get users with levels
right now, to get a list of users with their levels, we get those users from the users table with a single query, but then, for each of them, we do an additional query to retrieve their level. This is a larger workload load than needed in the server side. Consider if we can refactor this to get the list of users with levels in a single query.

Damrod:
as explained in #585
closes #585

### DIFF
--- a/transcendence-app/src/user/infrastructure/db/postgres/friend.postgres.repository.ts
+++ b/transcendence-app/src/user/infrastructure/db/postgres/friend.postgres.repository.ts
@@ -83,7 +83,7 @@ export class FriendPostgresRepository
           WITH ulevelwithgame AS (SELECT ul.*, g.${gameKeys.CREATED_AT}, g.${gameKeys.GAMEDURATIONINSECONDS}
                                   FROM ${table.USER_LEVEL} ul
                                            INNER JOIN ${table.GAME} g ON ul.${userLevelKeys.GAMEID} = g.${gameKeys.ID}
-                                  WHERE g.${gameKeys.GAMEMODE} = ${GameMode.classic}),
+                                  WHERE g.${gameKeys.GAMEMODE} = '${GameMode.classic}'),
                ulevelwithtime
                    AS (SELECT (ulg.${gameKeys.CREATED_AT} + INTERVAL '1 second' * ulg.${gameKeys.GAMEDURATIONINSECONDS}) AS "timestamp",
                               ulg.${userLevelKeys.USERNAME},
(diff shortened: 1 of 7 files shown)


transcendence-app/src/user/infrastructure/db/postgres/friend.postgres.repository.ts
```diff
@@ -67,20 +74,46 @@ export class FriendPostgresRepository
...
-      WHERE f.${FriendKeys.FOLLOWER_ID} = $1 AND u.${userKeys.USERNAME} ILIKE $2
-      ORDER BY ${orderBy}
-      LIMIT $3
-      OFFSET $4;`,
+    const usersData = await makeQuery<UserWithLevelData>(this.pool, {
+      text: `
+          WITH ulevelwithgame AS (SELECT ul.*, g.${gameKeys.CREATED_AT}, g.${gameKeys.GAMEDURATIONINSECONDS}
+                                  FROM ${table.USER_LEVEL} ul
+                                           INNER JOIN ${table.GAME} g ON ul.${userLevelKeys.GAMEID} = g.${gameKeys.ID}
+                                  WHERE g.${gameKeys.GAMEMODE} = '${GameMode.classic}'),
```
gbudau, line 86: praise: This query works perfectly :rocket:

question: Any reason why we select the `classic` game mode here?

thought: In hindsight, I wish we have added a `"lastLevel"` column on the Users table to simplify the queries for a user with a level.

gbudau (approved):
Thank you for these improvements! LGTM :rocket:

transcendence-app/src/user/infrastructure/db/postgres/friend.postgres.repository.ts
Damrod: I agree. Both here and have a last message for a chatroom. On the other hand, I've learned a lot from writing these queries.

Damrod: About classic:
Well, a user has a different level in each of the modes. It's possible that they'll have level 1 in short paddle and level 15 in classic, so if the last game they played at one point was short paddle, without this filter, the last level would be 1, and as soon as they played a classic game for an end level of 15, it would suddenly jump to 15.
We could specify what the last level is for, it just take one of them, and stay consistent.
I picked classic because I think it's our baseline.

PR #593, opened 2023-03-16
#592/Run migrations from docker compose file

### LINKED ISSUE #592: Run migrations from docker compose file
Subtask of #291

gbudau:
close #592

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,7 +57,7 @@ services:
     environment:
       - POSTGRES_DB=ft_transcendence
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER}"]
       interval: 5s
       timeout: 10s
       retries: 5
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -36,12 +38,29 @@ services:
...
+
   db:
     image: postgres:14.7-alpine3.17
     restart: always
     user: postgres
     environment:
       - POSTGRES_DB=ft_transcendence
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
```
Damrod, line 61: how did you come up with these values for interval, timeout, retries?

Damrod (approved):
LGTM!
Thank you!

docker-compose.yml
gbudau: I first used the values from the [docs](https://docs.docker.com/compose/compose-file/#healthcheck), then decreased the interval and increased the retries because compose runs the first health check command after the interval has passed, increasing the time it takes to deploy.

gbudau:
Thank you!

vdedios (approved):
Thank you! 🚀

Issue #594, opened 2023-03-18
error log line in db container


Damrod:
As of now, the health check in the container running the database is raising this error:
```
ft_transcendence-db-1  | 2023-03-18 12:58:06.232 UTC [10204] FATAL:  database "your-postgres" does not exist
```

PR #595, opened 2023-03-18
#594/fix db health check trying to connect to non existing database

Damrod:
as explained in #594


gbudau (approved):
Thank you for the fix! LGTM :rocket:

PR #599, opened 2023-03-18
#291 restore make spec dependencies

### LINKED ISSUE #291: Everything should run with a single docker compose up
thought: In the subject, there is this line:
```
Everything should run with a single call to docker-compose up –build
```
Currently, we generate the code for the webapp, then do `docker-compose up --build`. I'm concerned that this might affect our evaluation.

I added this here because I didn't have time to review the previous PR in time.

_Originally posted by @gbudau in https://github.com/42AGV/ft_transcendence/pull/289#discussion_r1008844591_

Damrod:
when hardcoding the generated webapp files, I removed these dependencies (altogether removing the generating code), but then we decided we would preserve it. This is a leftover from the removal (I am readding it).
close #291


gbudau (approved):
Thanks! LGTM :tada:

PR #602, opened 2023-03-19
#601/all server config files in folders srcs

### LINKED ISSUE #601: all configuration files of the server should be in folder srcs
as explained in slack

Damrod:
As explained in #601 and [here](https://42born2code.slack.com/archives/C033YQ28QA0/p1679239187832209)
close #601


gbudau (approved):
Thank you for these improvements! LGTM :tada:

PR #604, opened 2023-03-19
#603/Refactor LocalService saveFileDataFromBuffer

### LINKED ISSUE #603: Refactor LocalService saveFileDataFromBuffer
The [docs](https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback) recommend encapsulating the logic into a `Readable` using `pipe` instead of the `WriteStream.write` to handle backpressure.

Thanks to [@Damrod](https://github.com/Damrod) for pointing this out.

gbudau:
close #603


transcendence-app/src/shared/local-file/local-file.service.ts
```diff
@@ -81,9 +82,13 @@ export class LocalFileService {
       path,
       mimetype,
     );
+    const dataStream = Readable.from(data);
 
-    outStream.write(data);
-    outStream.end();
+    dataStream.pipe(outStream);
+
+    outStream.on('error', (err) => {
```
Damrod, line 89: shouldn't we subscribe the listener before we call the pipe on the write stream? Not that I am sure about it...

Damrod (approved):
If this is it, feel free to merge

transcendence-app/src/shared/local-file/local-file.service.ts
gbudau: I think you are right. I've updated it. Thanks for the suggestion.

gbudau: I've changed this part to use [`stream.pipeline()`](https://nodejs.org/api/stream.html#streampipelinestreams-callback) which close the streams automatically in case of an error.

From docs:
> stream.pipeline() closes all the streams when an error is raised.

gbudau:
Thanks!

PR #605, opened 2023-03-21
#386 fix express adapter premature close error

### LINKED ISSUE #386: Update frameworks and packages
Use the last stable version of every framework or library and fix vulnerabilities

- [x] Update the NestJS version and the libraries from the transcendence-app package.json
- [x] Update the React version and the libraries from the webapp package.json
- [x] Update Dockerfiles and docker-compose.yml files to use the latest LTS node version and docker images
- [x] Update GitHub actions node version

Damrod:
As explained in #386, ever since we updated node we have continuosly an error log line when we attempt to stream avatar files. This has no functional impact, but the error log has to be handled. This is the handling of those log lines.
close #386


gbudau (approved):
It works perfectly. Thanks for the fix.

PR #608, opened 2023-03-24
#606/migrations on campus macs

### LINKED ISSUE #606: migrations on campus macs
docker-compose version there is 1.28.5. This version does not yet handle `condition: service_completed_successfully`, so we have to have a workaround to have migrations there.

gbudau:
close #606

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ DOCKER_COMPOSE := $(shell $(PROJECT_ROOT)/scripts/get-docker-compose.sh)
 .PHONY: all
 all:
 	$(DOCKER_COMPOSE) up --build -d
-	make seed
 	npm run build-packages
 
 .PHONY: down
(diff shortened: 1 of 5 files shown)


transcendence-app/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN npm clean-install && npm run build-packages
 RUN npm run build -w transcendence-app
 VOLUME ${TRANSCENDENCE_APP_DATA}
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
-CMD ["npm", "run", "start-app-dev"]
+CMD ["sh", "-c", "npm run knex:migrate:dev -w transcendence-app && npm run knex:seed:run -w transcendence-app && npm run start-app-dev"]
```
Damrod, line 20: are we sure we want to seed the database for the correction?

gbudau: Good point. I've updated it. Thanks for the suggestion.

Makefile
```diff
@@ -7,7 +7,6 @@ DOCKER_COMPOSE := $(shell $(PROJECT_ROOT)/scripts/get-docker-compose.sh)
 .PHONY: all
 all:
 	$(DOCKER_COMPOSE) up --build -d
-	make seed
```
gbudau, line 10: After these changes, we can't run make seed here because the migrations take a while to execute.

vdedios (approved):
Thank you!

gbudau:
Thank you!

